PR #5, opened 2020-03-23
refactor: move all core components to widgets

varl:
This introduces the new separation of what is a core component (style +
structure) and what is a widget (behavior). The idea is to keep core
components very stable and generic over time, and instead using the
concept of "widges" to mean things that are fully formed and usable
directly in an application.

This stability means that it becomes easier and safer to build new
components by composing core components. It also provides a clear
distinction between what is a core component and what is a widget.

BREAKING CHANGE: All @dhis2/ui-core exports have been migrated to
@dhis2/ui-widgets.


packages/core/src/ButtonBase/ButtonBase.js
```diff
@@ -0,0 +1,69 @@
...
+            disabled={disabled}
+            ref={ref}
+        >
+            {children}
+            <style jsx>{styles}</style>
+        </button>
+    )
+}
+
+export default React.forwardRef(ButtonBase)
```
HendrikThePendric: I like the default exports, because then users can use a name that's most appropriate for them. Also great to have refs here 👍 

Why don't we just export from the bottom of the file though? That seems more conventional...

packages/widgets/src/Button/Button.styles.js
```diff
@@ -0,0 +1,31 @@
...
+
+    .button-icon {
+        margin-right: 6px;
+        color: inherit;
+        fill: inherit;
+        font-size: 26px;
+        vertical-align: middle;
+        pointer-events: none;
+    }
+`
```
HendrikThePendric: What's the reasoning behind this split? I'd expect that the icon-related styles are also just part of the `ButtonBase`. I'm sure you've got a good reason for it, but I don't immediately see it....

packages/core/src/ButtonBase/ButtonBase.js
varl: It turned out that way for a few arbitrary reasons, so I can move it:

- The component needs to be a named function otherwise the name doesn't show up in devtools.
- We could wrap the function definition with the `React.forwardRef(function ButtonBase) {})` but I found it unclear.
- The `forwardRef` is important, so I wanted it as close to the function definition as possible.

varl: 9711ab34

varl:
@HendrikThePendric If you would like to review this, go ahead. Keep this in mind:

- This moves all our current components to widgets as they are.
- One component has then been extracted to core (`ButtonBase`) from widgets (`Button`).
- The `ButtonBase` is only style + structure and is intentionally kept very generic in terms of props.
- Base components need to be exported with a `forwardRef` so they can be kept very generic in terms of behavior and allow widgets to build on them.
- This PR does not intend to extract all of the base components to core, that will be done component by component so more people can contribute.

The main thing to think about:

- **core**: base components that consist of style + structure
- **widgets**: components that provide behavior and are ready to be used in a UI

HendrikThePendric:
@varl thanks for explaining, luckily that's how interpreted the changes I was seeing already 😄 

> The `ButtonBase` is only style + structure and is intentionally kept very generic in terms of props.

Is the comment above related to my https://github.com/dhis2/ui/pull/5#discussion_r396727267? I still don't really see why the `icon` related logic cannot be classified as "style and structure". Could you elaborate?

varl:
> @varl thanks for explaining, luckily that's how interpreted the changes I was seeing already smile
> 
> > The `ButtonBase` is only style + structure and is intentionally kept very generic in terms of props.
> 
> Is the comment above related to my [#5 (comment)](https://github.com/dhis2/ui/pull/5#discussion_r396727267)? I still don't really see why the `icon` related logic cannot be classified as "style and structure". Could you elaborate?

The icon is a component itself, so it has its own "style and structure" that is not part of the ButtonBase atom.

Having an "icon button" is a DHIS2 specific behavior that composes a `ButtonBase` and an `Icon`, so that's why it belongs in Button.

HendrikThePendric:
> The icon is a component itself, so it has its own "style and structure" that is not part of the ButtonBase atom.
> 
> Having an "icon button" is a DHIS2 specific behavior that composes a `ButtonBase` and an `Icon`, so that's why it belongs in Button.

Aha. Yes, the penny has dropped. Thanks for that.

PR #12, opened 2020-03-30
refactor(select): make selection string based

HendrikThePendric:
I've addressed the following:
- Refactored the components so that they work with string based values
- Updated docs and propTypes
- Updated demo stories
- Updated e2e stories and tests

All tests are green and when I do a full test run, I don't seem to be collecting any prop-type errors, so I think I've caught everything.

However, this PR touched a lot of files, so I won't be surprised if we end up finding some little issues.

### DIFF
--- a/cypress/integration/MultiSelect/duplicate_option_values.feature
+++ b/cypress/integration/MultiSelect/duplicate_option_values.feature
@@ -1,0 +1,7 @@
+Feature: Duplicate option values result in a conflict between the input and dropdown UI
+
+    Scenario: The MultiSelect has options with a duplicate value and this value is selected
+        Given a MultiSelect with options with a duplicate value and this value is selected
+        And the MultiSelect is open
+        Then the first option with the selected value is displayed in the input
+        But both options are highlighted in the dropdown
--- a/cypress/integration/SingleSelect/duplicate_option_values.feature
+++ b/cypress/integration/SingleSelect/duplicate_option_values.feature
@@ -1,0 +1,7 @@
+Feature: Duplicate option values result in a conflict between the input and dropdown UI
+
+    Scenario: The SingleSelect has options with a duplicate value and this value is selected
+        Given a SingleSelect with options with a duplicate value and this value is selected
+        And the SingleSelect is open
+        Then the first option with the selected value is displayed in the input
+        But both options are highlighted in the dropdown
--- a/cypress/integration/MultiSelect/allows_selecting/index.js
+++ b/cypress/integration/MultiSelect/allows_selecting/index.js
@@ -41,7 +41,7 @@ Then('the clicked option is selected', () => {
     cy.window().then(win => {
         expect(win.onChange).to.be.calledOnce
         expect(win.onChange).to.be.calledWith({
-            selected: [{ label: 'option one', value: '1' }],
+            selected: ['1'],
         })
     })
 })
@@ -50,10 +50,7 @@ Then('the clicked option is selected as well', () => {
     cy.window().then(win => {
         expect(win.onChange).to.be.calledOnce
         expect(win.onChange).to.be.calledWith({
-            selected: [
-                { label: 'option one', value: '1' },
-                { label: 'option two', value: '2' },
-            ],
+            selected: ['1', '2'],
         })
     })
 })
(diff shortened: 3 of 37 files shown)


HendrikThePendric:
@ismay Thanks for spotting those three things, I've addressed them in https://github.com/dhis2/ui/pull/12/commits/1ccc2238e35014957433359f9e0d2065db85cc9c. Could you re-review?

varl:
Now that we do not rely on the label + value, we could change the Option implementations to use the pattern we use elsewhere where the label is the children:

```
<Option value="smth">
  This is the label
</Option>
```

May be a different PR if we do that though.

HendrikThePendric:
> None of the commits that introduce breaking changes has a proper commit message.

Good catch. I'll squash my commits and add a correct footer

> > One thing we could do is also add some tests for what will happen with duplicate values (and other data the Select will process, but not function correctly with). So that we know how it fails, and can make sure that the failure won't change as well.
> 
> Do you think that'd be better suited as a follow up PR, or in this one? (Or not at all is also an option of course)

I'll look into this as well. It should end up being a passing test on "failing" behaviour. It'd be good to add, but I might require some assistance....

HendrikThePendric:
> Now that we do not rely on the label + value, we could change the Option implementations to use the pattern we use elsewhere where the label is the children:
> 
> ```
> <Option value="smth">
>   This is the label
> </Option>
> ```
> 
> May be a different PR if we do that though.

Done  in 7db3e9f

ismay:
> > Now that we do not rely on the label + value, we could change the Option implementations to use the pattern we use elsewhere where the label is the children:
> > ```
> > <Option value="smth">
> >   This is the label
> > </Option>
> > ```
> > 
> > 
> > May be a different PR if we do that though.
> 
> Done in [7db3e9f](https://github.com/dhis2/ui/commit/7db3e9f890b7cae36171f46781adc75d3ff791b1)

I think that that'll break the assumption the Select wrapper makes an the options having a label prop. I believe it uses that for filtering for example.

HendrikThePendric:
> I think that that'll break the assumption the Select wrapper makes an the options having a label prop. I believe it uses that for filtering for example.

Yeah, I can confirm that this is in fact happening. Somehow the test was green for me locally, but the test run here just failed due to filtering. I think filtering based on children is not really a valid approach. So I'll revert 7db3e9f.

I am still working on a test for the duplicate value behaviour though.

HendrikThePendric:
@varl @ismay I think this PR is now ready for a final re-review:
- I've squashed my commits and included a `BREAKING CHANGE` section in the commit body
- I've added tests that document what happens if the selects do encounter duplicate values
- I've kept the `label` prop on the option components instead of switching to `children`, because the children approach doesn't go well with how options are being filtered.

cypress/integration/SingleSelect/duplicate_option_values.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Dealing with duplicate option values inconsistently
```
ismay, line 1: Maybe you could call this: `Highlights all duplicate option values`, because even though it could be considered wrong behaviour, it's still consistent.

cypress/integration/MultiSelect/duplicate_option_values.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Dealing with duplicate option values inconsistently
```
ismay, line 1: Same here as https://github.com/dhis2/ui/pull/12/files#r401496277

cypress/integration/SingleSelect/duplicate_option_values.feature
HendrikThePendric: The inconsistency I am referring to is the fact that only a single option is "selected" in the input, but both are "selected" in the dropdown. I suppose you are right too, in the sense that it is consistently displaying this inconsistent behaviour 🙄 ...

The problem with `Highlights all duplicate option values` is that it only focusses on the dropdown, not the input.

How about something like this:
`Duplicate option values result in a conflict in between the input and dropdown UI`

ismay: Ah I see what you mean now, in that case your suggestion seems like a good description 👍

Issue #15, opened 2020-03-31
Issues with context when using different instances of our libs


HendrikThePendric:
@martinkrulltott encountered an issue where the Select rendered below the Modal. We have encountered this issue before and thought we had solved it by using context (`React.createContext()`). However, there seems to be a case where this falls apart:
- The `data-visualizer-app` is dependent on the `analytics` package
- Both data-visualizer and analytics have their own instance of ui-core.
- As such the different layers and layer-contexts seem to be fully unaware of each other
- And as a result the Select (from analytics) uses its default zIndex value because it doesn't know it's stacked on top of another layer

Martin [provided a screen recording on Slack](https://dhis2.slack.com/files/UPN89F9NU/F01161CQKUP/quick_walkthrough.mov) that demoes the issue very well. You can see in the screen-recording that the Select which is coming from the same ui-core instance as the Modal (both data-viz) is working fine and is on top of the Modal (Select gets z-index 3001), but the one coming from analytics is misbehaving and gets rendered below the Modal (it just used a z-index of 2000)....

I didn't get a chance to debug this properly, but going by what we are seeing in the screen-recording, I can't really think of another explanation. This isn't necessarily an urgent issue, but it does seem to be quite a fundamental one... I don't really know how to tackle it.

And this issue is relevant to the components I am working on at the moment, see https://github.com/dhis2/ui/pull/10

HendrikThePendric:
Some directions we could take this:
- Just file this as a known limitation and make sure `@dhis2/ui-core` / `@dhis2/ui` is always a peer-dependency, so every app just ends up with a single instance of it...
- Don't use context, but implement a layer system that purely relies on DOM traversal.
- Introduce some sort of wrapper component that creates the base Layer context so the context is not created in the Layer module
- Forget about the whole layering system and expose a `zIndex` prop on every component that "pops up", so that the app can control this...

These are just some pretty random ideas... Nothing is tested and everything is fully hypothetical...

HendrikThePendric:
FYI: I see we also have a `TableContext` file. It seems less likely that this is going to cause similar problems, but it does also create the context in the module too.

HendrikThePendric:
See this PR for an illustration of the problem https://github.com/dhis2/ui/pull/16, specifically the http://localhost:5000/?path=/story/component-widget-layer--it-falls-apart story...

varl:
[...]

PR #16, opened 2020-03-31
chore(layer): demo issue with context from different modules

HendrikThePendric:
This is not a real PR, it's just a minimalistic demo of the issue @martinkrulltott found....

We should look into a way of creating the context differently....

### DIFF
--- a/packages/widgets/src/LayerContext/LayerContext.stories.js
+++ b/packages/widgets/src/LayerContext/LayerContext.stories.js
@@ -1,0 +1,33 @@
+import React from 'react'
+import { Layer } from './LayerContext.js'
+import { LayerTwo } from './LayerContextTwo.js'
+
+export default { title: 'Component/Widget/Layer', component: Layer }
+
+const FIXED_INDEX = 1000
+
+export const ItWorksOK = () => (
+    <Layer zIndex={FIXED_INDEX}>
+        {zIndexComputed => (
+            <>
+                <h1>{zIndexComputed}</h1>
+                <Layer zIndex={FIXED_INDEX}>
+                    {anotherZIndexComputed => <h1>{anotherZIndexComputed}</h1>}
+                </Layer>
+            </>
+        )}
+    </Layer>
+)
+
+export const ItFallsApart = () => (
+    <Layer zIndex={FIXED_INDEX}>
+        {zIndexComputed => (
+            <>
+                <h1>{zIndexComputed}</h1>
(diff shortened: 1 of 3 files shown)


packages/widgets/src/LayerContext/LayerContextTwo.js
```diff
@@ -0,0 +1,62 @@
...
+ * @static
+ * @prop {function} children
+ * @prop {number} zIndex
+ */
+LayerTwo.propTypes = {
+    children: propTypes.func.isRequired,
+    zIndex: propTypes.number,
+}
+
+export { LayerTwo, useLayer }
```
HendrikThePendric: This is just a duplicate of the `Layer` from `LayerContext.js` but with a different name.

packages/widgets/src/LayerContext/LayerContext.stories.js
```diff
@@ -0,0 +1,33 @@
...
+        {zIndexComputed => (
+            <>
+                <h1>{zIndexComputed}</h1>
+                <LayerTwo zIndex={FIXED_INDEX}>
+                    {anotherZIndexComputed => <h1>{anotherZIndexComputed}</h1>}
+                </LayerTwo>
+            </>
+        )}
+    </Layer>
+)
```
HendrikThePendric: It goes wrong here, because each layer is referring to a different base context.

HendrikThePendric:
In https://github.com/dhis2/ui/pull/16/commits/1e5c9036bb85f49ed633f4cfa4599c9d87e734c5 I've found a "fix" for this problem. The only way I could think of was to attach the context to the window to avoid duplicates from being created. It's definitely not the most elegant solution available, but I couldn't think of another way to solve it. Any other suggestions would be most appreciated....

Personally I can't think of any other way to solve this, apart from not allowing several instances of ui-core from being used in one app.... Which, to be fair, probably is something worth discussing further anyway....

ismay:
Are we looking for an intermediate solution here, while we figure out a long term solution to the whole zindex issue? Just for some context on the solution we're looking for, because as an intermediary solution something dirty seems less problematic.

HendrikThePendric:
This pull request was just there to illustrate https://github.com/dhis2/ui/issues/15 which is now closed, so we can close this PR too.

For the record: in the end we decided not to fix this in our library, but to let the app prevent this problem from occurring, see https://github.com/dhis2/ui/pull/17.

PR #24, opened 2020-04-16
feat: add DHIS2 Sharing Dialog (TECH-274)

amcgee:
This is the implementation of the [DHIS2 Sharing Dialog](https://jira.dhis2.org/browse/TECH-274) (previously [d2-ui-sharing-dialog](https://github.com/dhis2/d2-ui/tree/master/packages/sharing-dialog))

HUGE credit to @cooper-joe who [implemented the first React version of this](https://codesandbox.io/s/sharing-81x52?file=/src/index.js) :-D (now outdated)

Here is the [design spec](https://www.sketch.com/s/7474b62e-0dd7-4de5-a07b-b79104f998cd/a/4lw5Yq#Inspector).

Screenshot 2020-11-23 at 10 08 42

## Cascading sharing for dashboard

[Design spec](https://www.sketch.com/s/7def0e34-0f0e-471e-a98f-be246cf5ed68/a/25bpYz1)

Screenshots:
Screenshot 2021-07-16 at 16 51 06
Screenshot 2021-08-09 at 10 56 02

TODO (help wanted):
- [x] Basic UI layout
- [x] Add / edit / delete permissions
- [x] ~Live state and diffing of current state vs initial state for Save button enablement~
- [ ] Fixed-width select input
- [x] Support adding users and groups instead of just users
- [ ] Validate and finalize the `sharingSettings` object structure
- [x] Dynamically load users and groups for auto-complete in new permission
- [ ] Public icon and "My organization" text (should this be dynamic @cooper-joe?)
- [x] Proper ordering of users and groups (should the groups always come first?)
- [x] Group icon
- [x] User avatars (name initials as per design)
- [ ] Maybe use the noun `Permission` for a sharing grant to a particular user or group?
- [x] Clean up code, add PropTypes
- [ ] Add support for `data` sharing settings (only applies to some objects)
- [ ] Optimise for small screens 
- [x] Complete dashboard specific part for cascading sharing once the backend for it is ready


amcgee:
This was supposed to be a draft... apologies

edoardo:
I added the search for users/groups while typing in the search field.
A list is populated with the matching results and each item can be clicked to be added to the list of permissions.
The removal of a permission also works now for both users and groups.
debounce doesn't work properly yet.
propTypes have been added where missing.

The `Autocomplete` component should be generic enough so that when we have a new use case for it, it can be moved as a standalone widget and imported wherever needed.
I'm not convinced about the name, I'm open for suggestions.
Also we can discuss and decide how flexible we want it, right now the list with the search results is built-in.

packages/widgets/src/SharingDialog/sharingConstants.js
```diff
@@ -0,0 +1,55 @@
...
+        publicDescription: i18n.t('Anyone logged in can find and view'),
+        description: i18n.t('Can find and view'),
+        option: i18n.t('View only'),
+    },
+    [ACCESS_VIEW_AND_EDIT]: {
+        publicDescription: i18n.t('Anyone logged in can find, edit, and view'),
+        description: i18n.t('Can find, edit, and view'),
+        option: i18n.t('Edit and view'),
+    },
+}
```
HendrikThePendric, line 24: This needs to be changed to a function that is called from within a component, otherwise these strings will always be translated to `en`

HendrikThePendric:
My apologies @edoardo: in my last comment https://github.com/dhis2/ui/pull/24#discussion_r421443470 I pointed out a problem with the `i18n` strings, but I failed to point out we already have a small `translate` helper for this (could be that wasn't the case when I last reviewed). See a usage example here:
https://github.com/dhis2/ui/blob/c1fa323e9a0daef8f2a1953826988d414f973c73/packages/widgets/src/FileInputField/FileInputField.js#L85-L103

Also, the code is looking very good already and I noticed only one `TODO` in  a comment. This PR is still in draft, so I haven't done a full review. What is still needed to take this out of draft status?

packages/widgets/src/SharingDialog/sharingConstants.js
edoardo: Fixed.

cooper-joe:
Note that the original sketch posted here doesn't include the interaction if a sharing dialog needs to handle both data and metadata. I'm not sure that is included in this PR, but adding this mockup here for information.
image

edoardo:
@cooper-joe the data part is not implemented in this PR.

packages/widgets/src/SharingDialog/sharingConstants.js
```diff
@@ -0,0 +1,72 @@
...
+export const convertConstantToAccess = (constant, useBoolean) => {
+    switch (constant) {
+        case ACCESS_NONE:
+            return useBoolean ? false : '--------'
+        case ACCESS_VIEW_ONLY:
+            return useBoolean ? true : 'r-------'
+        case ACCESS_VIEW_AND_EDIT:
+            return useBoolean ? true : 'rw------'
+        default:
+            return useBoolean ? true : '--------'
```
Mohammer5, line 70: Why is the default boolean `true` while the string indicates no access whatsoever?
The string is identical to the `ACCESS_NONE` string, so I'd expect there to be false

Mohammer5 (commented):
Component appears to be quite generic, we could use it in all apps that allow setting sharing options, right?

Left some minor comments. When I figured out that this is still a draft, I stopped looking at details. Variable names could be improved a bit.

packages/widgets/src/SharingDialog/sharingConstants.js
edoardo: Good catch.
Will fix.

edoardo:
The component is supposed to be generic and reusable by different apps.
However, this PR does not include the implementation for the Data part of the sharing, which is used in Maintenance app.
The plan was to adopt this widget in the new FileMenu analytics apps are now using, there we only need the Metadata part of the sharing.

Mohammer5:
I'm not sure what we'd have to change in terms of props (e. g. `initialSharingSettings`) to make it work with data as well.
Is this component already built in a way that we just have to add information to that structure?

amcgee:
@edoardo do you think we could release this with metadata support only and then add data later?  I think it would be really great to get this into use.  If it looks like this might be close to ready I can take a look and "review" as well (though I can't actually review since I authored the original PR)

@Mohammer5 were you asking about `initialSharingSettings` etc. to determine if a breaking change would be needed to add data sharing later?

Mohammer5:
> @Mohammer5 were you asking about initialSharingSettings etc. to determine if a breaking change would be needed to add data sharing later?

@amcgee yes

edoardo:
The plan was to release this with metadata support only.
Since it didn't make it for 2.36 I'm wondering if we should complete it before we publish it.
There is another relatively important issue, currently it's not responsive and doesn't work in small screens.
Adopting it as it is would be a problem in dashboards app.

As for the breaking change, I don't think it would necessary be the case, I think the changes can be all internal, keeping the same props.

Mohammer5:
Do we have a design or specs for what it should look like on smaller screens?

edoardo:
> Do we have a design or specs for what it should look like on smaller screens?

Not that I know. @cooper-joe ?

edoardo:
> I'm not sure what we'd have to change in terms of props (e. g. `initialSharingSettings`) to make it work with data as well.
> Is this component already built in a way that we just have to add information to that structure?

No, I think we need to refactor that part.

cooper-joe:
> > Do we have a design or specs for what it should look like on smaller screens?
> 
> Not that I know. @cooper-joe ?

The first draft of the mobile design is done, but it needs some more work. It will also require that the `modal` `ui` component be set up to handle mobile too.

packages/widgets/src/SharingDialog/SharingDialog.js
```diff
@@ -0,0 +1,287 @@
...
+        setIsDirty(true)
+
+        removeUserOrGroupAccess(data)
+    }
+
+    return (
+        <Modal large position="top" onClose={onClose}>
+            <ModalTitle>
+                {i18n.t('Sharing & Access') +
+                    (sharingSettings.name ? `: ${sharingSettings.name}` : '')}
```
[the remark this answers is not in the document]
edoardo: Fixed.

packages/widgets/src/SharingDialog/SharingDialog.js
```diff
@@ -0,0 +1,287 @@
...
+}
+
+SharingDialog.defaultProps = {
+    onError: Function.prototype,
+    onSave: Function.prototype,
+}
+
+SharingDialog.propTypes = {
+    id: PropTypes.string,
+    initialSharingSettings: PropTypes.object,
```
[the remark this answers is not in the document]
edoardo: Fixed.

packages/widgets/src/SharingDialog/sharingConstants.js
edoardo: Fixed.

Mohammer5:
@edoardo are you still working on this? Would you like to discuss the current state? Or could this already be moved out of the draft state?

PR #27, opened 2020-04-18
feat: implement basic layout building blocks

amcgee:
**WORK IN PROGRESS**

This add composable `TopbarLayout` and `SidebarLayout` components.  The Sidebar is responsive by default, becoming a drawer on small screens.  It can also be fully controlled or static.

There's a lot of cleanup to do, probably quite a few edge cases, but this seems like a decent foundation to build on and should be sufficient for probably 70% of DHIS2 app use-cases.

### DIFF
--- a/packages/widgets/src/layouts/Layout.js
+++ b/packages/widgets/src/layouts/Layout.js
@@ -1,0 +1,9 @@
+import { layoutStyles } from './Layout.styles'
+import classNames from 'classnames'
+
+export const Layout = ({ className, style, children }) => (
+    <div className={classNames(className, 'layoutContainer')} style={style}>
+        <style jsx>{layoutStyles}</style>
+        {children}
+    </div>
+)
--- a/packages/widgets/src/layouts/Layout.stories.js
+++ b/packages/widgets/src/layouts/Layout.stories.js
@@ -1,0 +1,392 @@
+import React, { useState } from 'react'
+import { storiesOf } from '@storybook/react'
+import { TopbarLayout } from './TopbarLayout/TopbarLayout.js'
+import { Topbar } from './TopbarLayout/Topbar.js'
+import { SidebarLayout } from './SidebarLayout/SidebarLayout.js'
+import { Sidebar } from './SidebarLayout/Sidebar.js'
+
+import { LogoIcon } from '../'
+import Button from '../Button/Button.js'
+
+const AppsIcon = () => (
+    <svg
+        xmlns="http://www.w3.org/2000/svg"
+        width="16px"
+        height="16px"
+        viewBox="0 0 48 48"
+    >
+        <path d="M8 16h8V8H8v8zm12 24h8v-8h-8v8zM8 40h8v-8H8v8zm0-12h8v-8H8v8zm12 0h8v-8h-8v8zM32 8v8h8V8h-8zm-12 8h8V8h-8v8zm12 12h8v-8h-8v8zm0 12h8v-8h-8v8z" />
+        <path d="M0 0h48v48H0z" fill="none" />
+    </svg>
+)
+
+// TODO: height: 100% is too big when sibling layout panels have overflow.
+// Using a <Layout /> component for absolute positioning would work for "fill" needs, otherwise height: 100% shouldn't be necessary.
+
+const DemoContentBox = ({ filled, children, outlined = true }) => (
+    <div
+        style={
+            outlined
+                ? {
+                      border: '1px dashed green',
+                      width: '100%',
+                      minHeight: '100%',
+                  }
+                : undefined
+        }
+    >
+        {children}
+        {filled && (
+            <>
+                <h1>Hello World!</h1>
+                {[0, 1, 2, 3, 4, 5, 6, 7, 8, 9].map(idx => (
+                    <React.Fragment key={idx}>
+                        <h3>Hello there!</h3>
+                        <p>
+                            This is a paragraph of lots and lots of text, in
+                            order to demonstrate overflow behavior and other
+                            cool things.
+                        </p>
+                    </React.Fragment>
+                ))}
+            </>
+        )}
+    </div>
+)
+
+storiesOf('Component/Layout', module)
+    .add('Topbar', () => (
+        <TopbarLayout>
+            <Topbar>
+                <DemoContentBox />
+            </Topbar>
+            <DemoContentBox />
+        </TopbarLayout>
+    ))
+    .add('Sidebar', () => (
+        <SidebarLayout>
+            <Sidebar collapseBreakpoint={640}>
+                <DemoContentBox />
+            </Sidebar>
+            <DemoContentBox>
+                <h3>Try me!</h3>
+                <p>
+                    Resize your browser so that this viewport is narrower than
+                    640px.
+                </p>
+                <p>The sidebar will "collapse" out of the way of the content</p>
+            </DemoContentBox>
+        </SidebarLayout>
+    ))
+    .add('Sidebar (uncollapsible)', () => (
+        <SidebarLayout>
+            <Sidebar collapsible={false}>
+                <DemoContentBox />
+            </Sidebar>
+            <DemoContentBox>
+                <h3>Try me!</h3>
+                <p>
+                    Resize your browser so that this viewport is narrower than
+                    640px.
+                </p>
+                <p>
+                    Unlike the previouse example, this sidebar will not collapse
+                </p>
+            </DemoContentBox>
+        </SidebarLayout>
+    ))
+    .add('Sidebar (collapsed)', () => (
+        <div style={{ position: 'relative', width: 760, height: '100%' }}>
+            <SidebarLayout>
+                <Sidebar collapseBreakpoint={800}>
+                    <DemoContentBox />
+                </Sidebar>
+                <DemoContentBox>
+                    <h3>Sidebar Breakpoint</h3>
+                    <p>
+                        Let's artificially constrain the width of the container
+                        to 760px.
+                    </p>
+                    <p>
+                        We've also configured a custom{' '}
+                        <strong>collapseBreakpoint</strong> at 800px instead of
+                        the default 640px
+                    </p>
+                    <h4>BONUS FEATURE</h4>
+                    <p>
+                        A Layout can be embedded within any element with{' '}
+                        <strong>position: relative</strong> (notably including
+                        the Content section of other layouts!) and will adopt
+                        that element's size.
+                    </p>
+                </DemoContentBox>
+            </SidebarLayout>
+        </div>
+    ))
+    .add('Sidebar (force-collapsed)', () => (
+        <SidebarLayout>
+            <Sidebar collapsed>
+                <DemoContentBox />
+            </Sidebar>
+            <DemoContentBox>
+                <h3>Try me!</h3>
+                <p>Resize your browser so that this viewport is very large.</p>
+                <p>This sidebar will be collapsed at any size</p>
+            </DemoContentBox>
+        </SidebarLayout>
+    ))
+    .add('Sidebar (manual)', () => {
+        const [open, setOpen] = useState(false)
+        return (
+            <SidebarLayout>
+                {open && (
+                    <Sidebar collapsible={false}>
+                        <DemoContentBox />
+                    </Sidebar>
+                )}
+                <DemoContentBox>
+                    <h3>Try me!</h3>
+                    <p>Click the button below to manually toggle the sidebar</p>
+                    <p>
+                        Unlike a standard collapsible sidebar, this one will
+                        "push" the main content out of the way
+                    </p>
+                    <button onClick={() => setOpen(x => !x)}>
+                        Toggle sidebar
+                    </button>
+                </DemoContentBox>
+            </SidebarLayout>
+        )
+    })
+    .add('Sidebar (manual collapsed)', () => {
+        const [open, setOpen] = useState(false)
+        return (
+            <SidebarLayout>
+                {open && (
+                    <Sidebar collapsed={true} toggled={true}>
+                        <DemoContentBox>
+                            <button onClick={() => setOpen(false)}>
+                                Hide sidebar
+                            </button>
+                        </DemoContentBox>
+                    </Sidebar>
+                )}
+                <DemoContentBox>
+                    <h3>Try me!</h3>
+                    <p>Click the button below to manually toggle the sidebar</p>
+                    <p>
+                        In this example, the sidebar will open above the content
+                        and obscure it
+                    </p>
+                    <button disabled={open} onClick={() => setOpen(true)}>
+                        Show sidebar
+                    </button>
+                </DemoContentBox>
+            </SidebarLayout>
+        )
+    })
+    .add('Sidebar (overflow)', () => (
+        <SidebarLayout>
+            <Sidebar>
+                <DemoContentBox filled />
+            </Sidebar>
+            <DemoContentBox>
+                <h3>Try me!</h3>
+                <p>
+                    Resize your browser so that this viewport is{' '}
+                    <strong>short</strong> vertically.
+                </p>
+                <p>
+                    The sidebar and the content div are independently scrollable
+                </p>
+            </DemoContentBox>
+        </SidebarLayout>
+    ))
+    .add('Sidebar (right)', () => (
+        <SidebarLayout side="right">
+            <Sidebar>
+                <DemoContentBox>
+                    <h3>Hello, I'm a Right Sidebar!</h3>
+                </DemoContentBox>
+            </Sidebar>
+            <DemoContentBox>
+                <h3>Try me!</h3>
+                <p>
+                    Just like before, here's a Sidebar - but now it's on the
+                    right!
+                </p>
+                <p>
+                    It's collapsible too - Resize your browser so that this
+                    viewport is narrower than 640px.
+                </p>
+            </DemoContentBox>
+        </SidebarLayout>
+    ))
+    .add('Sidebar (right, collapsed)', () => (
+        <SidebarLayout side="right">
+            <Sidebar collapsed>
+                <DemoContentBox>
+                    <h3>Hello, I'm a Right Sidebar!</h3>
+                </DemoContentBox>
+            </Sidebar>
+            <DemoContentBox>
+                <h3>Try me!</h3>
+                <p>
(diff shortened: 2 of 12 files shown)


amcgee:
I manually pushed the Storybook to Netlify, so you can play around with it [here](https://goofy-bassi-239747.netlify.app/?path=/story/component-layout--composite-clean)

amcgee:
(from [Slack thread](https://dhis2.slack.com/archives/CBM8LNEQM/p1587371545177600?thread_ts=1587232827.156700&cid=CBM8LNEQM))

I think we want these to be opinionated but open to customization, which is what I’ve gone for here.  The goal should be that any DHIS2 app with a sidebar should look basically the same- if every sidebar is a different size or different border or has different collapse controls we will be back in the Wild Wild West of app design.

Some questions for @cooper-joe  -
1. how wide should that sidebar be?  I went for 320px here with automatic collapse at 640px (which can be configured or disabled)
2. I don’t love the look of the toggle button, any suggestions or want to design something quickly?
I’ve set the default height for the top bar to 46px (expandable), is that right?
3. All panels are independently scrollable, so the layout by design covers the entire page and disables the page scroll bar, do you see any possible challenges with that?  When would we want the sidebar to scroll with the content?

@Mohammer5 pointed out that the top bar could easily be flipped to the bottom with a prop (like the side prop in sidebar), anyone have suggestions for a better name if it's not necessarily on top?

I was thinking the sidebar could also be a bottom panel, which would make it more of an “
aside”, but thats maybe a future enhancement...

The three additional layout building blocks I’ve found I would like in building different apps are
1. A centered overlay panel
2. A simple grid container for a group of DOM elements with an 8px gutter between them, possibly with
3. A set of buttons which becomes a hamburger menu on small screens
4. Maybe a simple utility for hiding/showing something at different breakpoints
5. Maybe also a content container with the ability to “break out” into full-screen mode and hide all the other layout elements

amcgee:
One option - maybe just `ColumnLayout` and `RowLayout` with a `reversed` prop to flip them?  I don't like the names as much, since they're a bit more opaque, and also it's not as clear that `Sidebar` only really works in `RowLayout`...

amcgee:
Another question for @cooper-joe 
4. Should we show a ScreenCover behind the Sidebar (drawer) when it's collapsed but toggled?  That would obscure the page content but provide another way to "hide" the drawer than just by clicking the toggle

cooper-joe:
Great initiative, I'm happy to see this discussion moving so thank you @amcgee for putting in the work to move this along.

> this seems like a decent foundation to build on and should be sufficient for probably 70% of DHIS2 app use-cases.

I agree strongly with this. I do not think a layouts library can provide _all_ layouts DHIS2 apps might use, but we can reasonably provide the basics that _most_ will use. I think this is important to keep in mind throughout the discussion.


> I think we want these to be opinionated but open to customization, which is what I’ve gone for here

I agree that the default design can be opinionated. I want to be clear that I do not think that the layout elements should be opinionated about the content they should/can contain. That may come further down the line, but for now, I think these layout elements should just be containers oblivious to their content.

### Sidebar / SidebarLayout

I think renaming `sidebar` to `panel` would serve the unopinionated-but-useful cause. I think `sidebar` carries too much weight as to the content it should contain. Sidebar often means 'links, navigation, related content'. This is often true, but not always. There are cases that we do not know about that may use a left-aligned panel in a non-standard way. I would not want to discourage this by presenting an opinion about that component's content via its name. All of the functionality you've included is still valid on a `panel`.

I like the functionality that's included: fixed panels, floating panels, collapsable panels. This is often repeated stuff that is great to standardize and help devs move faster.

> @Mohammer5 pointed out that the top bar could easily be flipped to the bottom with a prop (like the side prop in sidebar), anyone have suggestions for a better name if it's not necessarily on top?

We could just refer to `bar` and allow `top` and `bottom`, props?


### Responses to the questions raised:

> how wide should that sidebar be? I went for 320px here with automatic collapse at 640px 

I think we can reduce this to 280px for the width. As long as it is configurable we can figure out a good default in testing.
	
> I don’t love the look of the toggle button, any suggestions or want to design something quickly?
	
I'll take a look and see if some minor adjustments can help.

> I’ve set the default height for the top bar to 46px (expandable), is that right?

Specs for headerbar listed it as 48px

> All panels are independently scrollable, so the layout by design covers the entire page and disables the page scroll bar, do you see any possible challenges with that? When would we want the sidebar to scroll with the content?

I am vaguely uncomfortable with this, although I am unable to articulate why. I feel like this is rebuilding functionality (page scrolling) that already exists natively. But I am not a css wizard so this may be a necessary trade-off to get the functionality we want. I'd like to hear more opinions about this as I do not feel informed enough to trust my instinct.
	
	
> Should we show a ScreenCover behind the Sidebar (drawer) when it's collapsed but toggled? That would obscure the page content but provide another way to "hide" the drawer than just by clicking the toggle

In most cases, yes. I think this would be the default behaviour. Disabling this could be an option.

varl:
I think that layouts deserves its own package: `ui-layouts` as they will be very well isolated, though I guess they may be added to widgets here for simpler prototyping?

amcgee:
> I think that layouts deserves its own package: ui-layouts as they will be very well isolated, though I guess they may be added to widgets here for simpler prototyping?

@varl yeah absolutely!  I just added them here since all the scaffolding was already in place, will move them when I get back to this

amcgee:
@varl moved to a separate `@dhis2/ui-layouts` package

amcgee:
> I am vaguely uncomfortable with this, although I am unable to articulate why. I feel like this is rebuilding functionality (page scrolling) that already exists natively. But I am not a css wizard so this may be a necessary trade-off to get the functionality we want. I'd like to hear more opinions about this as I do not feel informed enough to trust my instinct.

@cooper-joe I get that, for sure.  It's not actually rebuilding much functionality, since we're still using the browser-native overflow and scroll behaviors, just on a `div` rather than on `body` (we do this many places already)

I think the question for me is if, say, you have a `<Panel left>` and some content, should the entire layout (both the `<Panel>` and the content) scroll together?  I could see this as another option, since I see that as a valid use-case as well (though only for a static `Panel`, not a collapsible one) but I worry a bit about maybe having too many options?

amcgee:
Question: what should we call the containers of `Panel` and `Bar` layouts?  Previously these were `SidebarLayout` and `TopbarLayout`, `PanelLayout` and `BarLayout` don't seem quite right though... maybe `HorizontalLayout` and `VerticalLayout`?

Also, should `Panels` and `Bars` be direction-agnostic?  I.E. can you have a vertical left-aligned Bar (like a collapsed sidebar with icons) or a horizontal bottom-aligned Panel (like the DataTable in Maps)?  I think the answer might be **Yes** but this does make the job of building easy-to-use general utilities a bit more challenging.

amcgee:
Here's the API I'm currently envisioning, with all props optional:

```jsx
<RowLayout [reversed] />
<ColumnLayout [reversed] />

// Simple 48px (expandable) bar
<Bar />

// TENTATIVE Something to be able to split the `Bar` into sections, optionally, i.e. the section of the topbar above the sidebar in DV
<BarSection />

// Rendered inline, can be conditionally rendered (should we have a `hidden` prop?)
<StaticPanel />

// When collapsed is `true` or `false`, hide the toggle button
<CollapsiblePanel [collapsed] /> 

// <Panel /> = <StaticPanel /> @ width/height >= collapseBreakpoint, else <CollapsiblePanel [collapsed] />
<Panel [collapsed] [collapseBreakpoint=840px] />

// And maybe a few helper components
<InlineGrid [row | column] [small | medium | large] /> = 
                <div
                    style={{
                        display: 'inline-grid',
                        gridAutoFlow: 'row | 'column',
                        gridGap: 8 | 16 | 24, // small, medium, large?
                        alignItems: 'center',
                    }}
                />
```

So the above example would be :

```jsx
        <ColumnLayout>
            <Bar>
                <InlineRow large>
                    <div style={{ width: 32 }}>
                        <LogoIcon />
                    </div>
                    <strong>Greetings from DHIS2!</strong>
                </InlineRow>
                <Button primary>Login</Button>
            </Bar>
            <RowLayout>
                <Panel>
                    <p>
                        Now let's take off the training wheels and get rid of
                        those content block outlines
                    </p>
                </Panel>
                <div>
                    <h1>So long, and thanks for all the fish!</h1>
                    <p>
                        That's all for now, go build DHIS2 apps with these
                        stunningly simple Layouts!
                    </p>
                </div>
            </RowLayout>
        </ColumnLayout>
```

cooper-joe:
> 
> >I am vaguely uncomfortable with this, although I am unable to articulate why. I feel like this is rebuilding functionality (page scrolling) that already exists natively. But I am not a css wizard so this may be a necessary trade-off to get the functionality we want. I'd like to hear more opinions about this as I do not feel informed enough to trust my instinct.
> 
> @cooper-joe I get that, for sure. It's not actually rebuilding much functionality, since we're still using the browser-native overflow and scroll behaviors, just on a div rather than on body (we do this many places already)

True! If it does not bother you then I can safely say it doesn't bother me either ;) 

> I think the question for me is if, say, you have a <Panel left> and some content, should the entire layout (both the <Panel> and the content) scroll together? I could see this as another option, since I see that as a valid use-case as well (though only for a static Panel, not a collapsible one) but I worry a bit about maybe having too many options?

I don't think the scenario you describe here needs a pre-built option in the layouts library. It theoretically could exist, but I think the vast majority would want the areas to be independent of one another.

>Also, should Panels and Bars be direction-agnostic? I.E. can you have a vertical left-aligned Bar (like a collapsed sidebar with icons) or a horizontal bottom-aligned Panel (like the DataTable in Maps)? I think the answer might be Yes but this does make the job of building easy-to-use general utilities a bit more challenging.

To keep it simple, maybe we should just provide a default direction, or just `LTR` and `RTL` versions (maybe this is handled higher up?). I think if a user needs a bottom aligned horizontal layout inside a `Panel`, for example, it's simple enough to compose that yourself? I guess it's simple to compose the content _inside_, but the actual _placement_ of the `Panel` might be more difficult.

>Here's the API I'm currently envisioning, with all props optional

I'm liking this. There is a great deal of flexibility available if a user can compose an app with `Panel`s and `Bar`s inside `Column`s and `Row`s.

amcgee:
@cooper-joe do you think `RowLayout` and `ColumnLayout` might be too generic as names?  A `ColumnLayout` which can only have a `Bar` and some content, and a `RowLayout` which can only have a `Panel` and some content seems a bit strange... and I think `RowLayout` and `ColumnLayout` kind of imply that you can have an arbitrary number of (equal-size?) columns within them...  Maybe `BarLayout` and `PanelLayout`?  

We **might** be able to get away with `<FlexLayout direction="row|column|row-reverse|column-reverse">` which aligns a little better with Flex layouts, but maybe that exposes too much of the internals...

cooper-joe:
Yeah, agreed that `RowLayout` and `ColumnLayout` are too generic.

I think `BarLayout` and `PanelLayout` is a nice middle-ground that avoids exposing the internals, the danger you mentioned with the `Flex..` names.

amcgee:
Ref an older PR with a very similar concept by @Mohammer5 https://github.com/dhis2/ui-core/pull/318

amcgee:
@cooper-joe I think you had updated the collapsible sidebar icon but it doesn't look like you pushed, do you want to do that before we go about renaming things?

HendrikThePendric:
@amcgee I like the idea of having layout building blocks. In their current form I think they are doing a little bit too much, IMO they should only focus on layout and be suitable for any content. So things like padding, borders, background-color should probably be omitted.

If we do want to enforce styling at this level, then I would still argue that we should at least get rid of the padding on the side-bar and the top-bar. The sidebar would typically hold a nav Menu and the topbar would typically hold Tabs. Both would look bad in a padded container.

Also, I wonder about the `useParentSize` hook. We should be able to implement this sidebar incl. collapsing by just using CSS (fixed positioning and some css transform), and get rif of this hook.

amcgee:
> Also, I wonder about the useParentSize hook. We should be able to implement this sidebar incl. collapsing by just using CSS (fixed positioning and some css transform), and get rid of this hook.

That's not possible when the layout can be nested, AFAIK

> If we do want to enforce styling at this level, then I would still argue that we should at least get rid of the padding on the side-bar and the top-bar. The sidebar would typically hold a nav Menu and the topbar would typically hold Tabs. Both would look bad in a padded container.

I'm open to removing the padding, but I worry a bit that it will be misused.  The idea is to make it as simple as possible to build a good-looking, standard app layout with these building blocks.  I'm not entirely sure that the nave menu & tabs is an assumption we want to make.

> @amcgee I like the idea of having layout building blocks. In their current form I think they are doing a little bit too much, IMO they should only focus on layout and be suitable for any content. So things like padding, borders, background-color should probably be omitted.

I think the borders and background-colors are important quite useful, actually, as otherwise the CSS needs to be duplicated everywhere we use a sidebar or topbar, but I get your point.  @cooper-joe what do you think?

HendrikThePendric:
> I'm not entirely sure that the nave menu & tabs is an assumption we want to make as many apps don't follow that pattern (DV, Maps, Event Reports / Visualizer, dashboard, Headerbar in any app, .

🤔  I think there are many apps that will have a sidebar with menu-items, i.e. user-app / user-profile-app / settings-app / maintenance-app / import-export-app / etc. But I guess the Tabs part is probably not a common pattern. It makes a lot of sense to have it in the maintenance app though.

amcgee:
> I think there are many apps that will have a sidebar with menu-items, i.e. user-app / user-profile-app / settings-app / maintenance-app / import-export-app / etc. But I guess the Tabs part is probably not a common pattern. It makes a lot of sense to have it in the maintenance app though.

Very true.  Any reason these shouldn’t be in a collapsible sidebar with padding though?  I personally find it cleaner but @cooper-joe is the authority here.  We could have styled and « pure layout » options, technically

HendrikThePendric:
> Any reason these shouldn’t be in a collapsible sidebar with padding though?

I have no objections to the collapsable part, but the menu should not have any padding around it

cooper-joe:
A sidebar-menu should not have padding. Hendrik's screenshots above illustrate the issue with padding.

I think a lot of sidebars will be employed as a sidebar-menu, so it makes sense to make this easy. 

One possible way is to distinguish between sidebar-as-a-container-for-whatever and sidebar-as-a-menu. Sidebar-as-a-menu would probably have other helpful features built in: item styling, flyout menus etc. I'm not sure whether sidebar-as-a-menu would live in this layout library though, it seems like something to be included in `ui`. 

So, even though it's tempting to add a consistent padding, I think the sidebar that may house a sidebar-menu needs to be without padding.

amcgee:
Yeah, I'm on board with omitting the padding.  @cooper-joe what about colors?  (see @HendrikThePendric 's screenshot)

cooper-joe:
> Yeah, I'm on board with omitting the padding. @cooper-joe what about colors?

Including basic color styling by default is definitely helpful. I think being able to customise these defaults with a `class` override or something will also be needed.

amcgee:
Yeah, class overrides are supported 👍

Mohammer5:
Up to now, we've tried to communicate that class overrides should be the last resort and are not encouraged as it might break even with minor version updates. We shouldn't make an exception for this one, as that'd just cause confusion about where to use css classes and where not to.
Props accepting colors or theme-names is better imo.

Mohammer5:
Still relevant, especially with [this slack conversation](https://dhis2.slack.com/archives/CBM8LNEQM/p1608108621136500) in mind

Mohammer5:
Notes for later:

* I thought about this the last couple of days a bit more. The more I think about it, the more I think we shouldn't have layout components that try to be flexible enough to work in all kinds of situations, as we'd have to cover a lot of situations (e. g. the Transfer component) that are quite different from page layouts
* Maybe it makes more sense to have components, like the `CollapsibleSidebar`, a `TopBar` and whatnot that apps can use or simply provide css variables that we can use in apps to define the dimensions of grid areas
* Especially as more or less all components and some apps deviate from the layouts of apps, I think that:
  * The abstraction of layout components does not save any work
  * Could make layouting more complicated when custom grids need to be defined

So maybe it makes more sense to not have layout components but rather let apps and components do it themselves (even when the component is a organism) and instead provide components and css variables that apps can use like the `CollapsibleContainer` or `--sidebar-width-full`/`--sidebar-width-collapsed`

packages/layouts/src/useParentSize.js
```diff
@@ -0,0 +1,42 @@
...
+                width: el.clientWidth,
+                height: el.clientHeight,
+            })
+        }
+
+        onResize(el)
+
+        if (renderCounter) {
+            setSize(initialState)
+        }
```
HendrikThePendric, line 33: Why do we need this part?

HendrikThePendric (commented):
In general, I think we should aim for a simpler implementation with less JavaScript:
- I would prefer just having a few specific building blocks that a dev manually sets props for
- I think all components should be purely controlled, or if we do want/need to have a component with internal state, we should also expose the simpler building block(s) it is rendering.
- I don't like the components cloning their children
- I am unsure we really need the `useParentSize` hook... I believe it's only required by the sidebar, and we should be able to implement that without such a hook, provided we restrict ourselves from [not nesting sidebars](https://github.com/dhis2/ui/pull/27#issuecomment-642545745).

I think some/most of these comments I have voiced before in the past but I just had a fresh look and this is what my impression was.

Mohammer5 (commented):
> In general, I think we should aim for a simpler implementation with less JavaScript:
I think I agree with that sentiment.

Maybe a good first step would be to just have some html + css components that don't do anything smart (the rest could be added later on if needed). The only thing that could be controllable is `collapsed` so it's consistent across apps; I guess react's context is really nice here, but I know that we want to avoid that because it creates a singleton issue, I guess. But apps can create context themselves if the sidebar just passes the prop to a render fn).

Regarding the position: absolute thing that @HendrikThePendric mentioned, we could also just set `display: flex` & `align-items: stretch` properties on the layout content divs (and the app root; might make sense to have sth like `LayoutRoot` that's just `height: 100%` to have the same height as `#root`):

image

varl:
> Maybe a good first step would be to just have some html + css components that don't do anything smart (the rest could be added later on if needed).

I third the "less javascript, more css+html" due to lessons learned in the past. Let's start as simple as we can and see what we need to add once we start to uncover use cases. :detective:

Mohammer5:
Kind of still relevant, but has been forgotten

PR #35, opened 2020-04-22
fix: re-export final-form exports

### LINKED ISSUE #32: Reexport final-form from ui-forms as well
We reexport the `react-final-form` exports, but not the `final-form` exports. The `final-form` package contains exports that are also potentially useful in an app, like: `import { FORM_ERROR } from 'final-form'` (which I need atm.).

To unify the approach it'd be good to reexport `final-form`'s exports as well. I'm not a fan of reexporting, but since we're doing it I think it'd be good to have a uniform approach. Addressing the above would allow apps to omit `final-form` from their dependencies (which technically they could already do, but it's common practice to list all packages you're importing from explicitly in package.json).

ismay:
This re-exports the named exports from final-form (see here: https://final-form.org/docs/final-form/api). There are no default exports from the package, so this covers all exports.

See linked issue for details.

Closes #32


Mohammer5:
I've "complained" about this before. I'd prefer to export `*` instead of a chosen subset, which would be future-proof if they add more stuff. I don't see a reason to limit the libraries exports.

ismay:
> I've "complained" about this before. I'd prefer to export `*` instead of a chosen subset, which would be future-proof if they add more stuff. I don't see a reason to limit the libraries exports.

That's fine with me as well. @HendrikThePendric, objections to using `*`?

ismay:
Tentatively updated the style to `*`, let me know if you agree/disagree Hendrik.

varl:
One practical argument is that if we export `*` from our re-exported libraries, then any breaking change they do upstream forces us to release a breaking change for our library.

**Example:**

Someone does `import { FinalFormFeature } from '@dhis2/ui-forms`, and then upstream drops `FinalFormFeature` from their package and publishes a breaking change.

In `@dhis2/ui-forms` we update _our_ code to not rely on `FinalFormFeature`, but figure it is not a breaking change for us, so we publish `@dhis2/ui-forms` with a `fix(deps): update final-form to new major`.

Somewhere, elsewhere, a developer has imported `FinalFormFeature` from our `@dhis2/ui-forms` package, and bumps the patch version of our library and _poof_. Now his code has broken without realizing why. All he did was bump the patch version and we silently shipped a breaking change to him.

**Problem**:

Yes, it's easier for us to just re-export `*` and it is more future-proof in the sense that we can transparently update our dependencies and have those changes reflected through our public API.

The point of being explicit in re-exporting is that we get a safety mechanism in that our build will fail if we attempt to re-export an export that no longer exists in the library we are re-exporting. Our build would fail **before** our users' builds would fail. I think that this is an important responsibility that we have.

**Alternative**:

We can gather up all the named exports for FinalForm and perhaps ReactFinalForm in a namespace that we can provide jsdocs for explaining the risks:

```
/**
 * @module
 * @desc Allows direct access to the FinalForm library. Please note that this is considered advanced
* usage and that you need to stay up to date with breaking changes in the FinalForm library.
 * @return FinalForm 
 */
export { * as FinalForm } from 'final-form'

/**
 * @module
 * @desc Allows direct access to the ReactFinalForm library. Please note that this is considered
* advanced usage and that you need to stay up to date with breaking changes in the FinalForm
* library.
 * @return ReactFinalForm 
 */
export { * as ReactFinalForm } from 'react-final-form'
```

In addition to that, we can provide "safe" named re-exports for react-/final-form features that are absolutely essential as we did before if we want to.

varl:
We (@HendrikThePendric, @ismay, @Mohammer5, @varl) had a meeting about this and decided:

1. We need tests to check if the re-exported libraries change their API and automatically catch those errors, and we can do that using Jest Snapshots.

2. We are going to namespace wrapped libraries using e.g. `export { * as FinalForm } from 'final-form'`.

ismay:
I've added a test for the exports. The advantage of our new convention is that the test can be very simple:

```js
describe('final-form named export', () => {
    it('should re-export the expected exports', () => {
        const exports = Object.keys(FinalForm)

        expect(exports).toMatchSnapshot()
    })

    it.each(Object.keys(FinalForm))(
        'should re-export the expected type for %s',
        exportName => {
            const type = typeof FinalForm[exportName]

            expect(type).toMatchSnapshot()
        }
    )
})
```

A lot easier than the test for our prop-types, where we had to check if anything was being overwritten as well.

One thing that's happening, where I'm not sure what the cause is is:

Screenshot 2020-04-23 at 12 07 25

Any idea what's going on there @HendrikThePendric? I could suppress it if we know that it's not an indication that something's going wrong.

varl:
@ismay I talked to @HendrikThePendric a while back about this, and he came up with a solution (and he reminded me of it on Slack now).

The problem here is the ButtonBase component, not your tests. This fixes it:

```
diff --git a/packages/core/src/ButtonBase/ButtonBase.js b/packages/core/src/ButtonBase/ButtonBase.js
index fe29fc3b..fa32a70d 100644
--- a/packages/core/src/ButtonBase/ButtonBase.js
+++ b/packages/core/src/ButtonBase/ButtonBase.js
@@ -10,21 +10,21 @@ import styles from './ButtonBase.styles.js'
  * @param {ButtonBase.PropTypes} props
  * @returns {React.Component}
  */
-function ButtonBase(
-    {
-        children,
-        className,
-        primary,
-        secondary,
-        destructive,
-        small,
-        large,
-        disabled,
-        ...props
-    },
-    ref
-) {
-    return (
+const ButtonBase = React.forwardRef(
+    (
+        {
+            children,
+            className,
+            primary,
+            secondary,
+            destructive,
+            small,
+            large,
+            disabled,
+            ...props
+        },
+        ref
+    ) => (
         <button
             {...props}
             className={cx(className, {
@@ -41,7 +41,9 @@ function ButtonBase(
             <style jsx>{styles}</style>
         </button>
     )
-}
+)
+
+ButtonBase.displayName = 'ButtonBase'
 
 /**
  * @typedef {Object} PropTypes
@@ -66,4 +68,4 @@ ButtonBase.propTypes = {
     small: propTypes.bool,
 }
 
-export default React.forwardRef(ButtonBase)
+export default ButtonBase
```

Do you want me to push that to this PR or do it in my branch for prepping v5?

varl:
Either way this looks very nice!

ismay:
@varl Ah thanks for the info.

> Do you want me to push that to this PR or do it in my branch for prepping v5?

Whichever you prefer is fine with me 👍

varl:
We can keep your branch clean and I'll fix it on my prep-branch.

ismay:
> We can keep your branch clean and I'll fix it on my prep-branch.

Cool, thx 👍

PR #37, opened 2020-04-22
test: change root-dir config to match all packages

ismay:
A follow up to the discussion here: https://dhis2.slack.com/archives/CBM8LNEQM/p1587567537209200

Todo:

- [x] We're mixing `__test__` and `__tests__` dirs
- [x] On alpha we're only running tests in core and widgets due to our jest.config.js


jest.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
     rootDir: '.',
     setupFilesAfterEnv: ['<rootDir>config/jest/enzymeSetup.js'],
-    roots: ['<rootDir>/packages/core/src', '<rootDir>/packages/widgets/src'],
+    roots: ['<rootDir>/packages'],
```
ismay, line 4: This config worked for me locally. Another approach would be to list each package explicitly, and @amcgee proposed:

```
testMatch: [ "packages/*/src/**/__tests__/**/*.[jt]s?(x)", "packages/*/src/**/?(*.)+(spec|test).[jt]s?(x)" ],
roots: ['<rootDir>'],
```

Which didn't work for me, but it does seem like a nice mix of explicit and terse. Maybe it just needs to be tweaked a little.

amcgee: I hadn't tested that... maybe it needs to be `packages/**`?  I think keeping the default match pattern is nice though..

amcgee:
@ismay is the failing test unrelated?

ismay:
> @ismay is the failing test unrelated?

Those are tests that were broken somewhere, but that we didn't catch because we weren't running them. So related in a way, but not really.

jest.config.js
ismay: Ok, packages needed to be prepended with `**/`, so `**/packages/...`. That works for me locally.

I'm wondering though if 

```
    testMatch: [
        '**/packages/**/__tests__/**/*.[jt]s?(x)',
        '**/packages/**/?(*.)+(spec|test).[jt]s?(x)',
    ],
    roots: ['<rootDir>'],
```

is effectively not the same as

```
    // standard testMatch regex
    roots: ['<rootDir>/packages'],
```

I think the results are the same. I assume that @varl used the rootDir config to set the root per package, to stop the lookups for mocks etc. at the package boundaries. Which the above wouldn't do I think.

ismay: Anyway, I'll give @varl a chance to take a look, and look at this myself again tomorrow morning.

varl: I tried so many different patterns here and eventually settled on being explicit. I'm fine with `roots: ['<rootDir>/packages']`.

Related from when I battled with this: https://github.com/dhis2/app-platform/issues/269

ismay: > I'm fine with roots: ['<rootDir>/packages'].

All right, I'll stick with that one 👍

PR #40, opened 2020-04-23
style: remove stray eslint-disables

ismay:
Doesn't necessarily have to go into alpha, just based it on that to be working the latest version.

* `packages/widgets/src/OrganisationUnitTree/OrganisationUnitNode.js` I noticed a stray eslint-disable here during another PR. I left one linting error that I wanted to ask JG whether we should ignore it, or remove the prop-type.
* `packages/widgets/src/OrganisationUnitTree/propTypes.js` I did a grep for any other eslint-disables and found some eslint-disables here that I made more specific. A linting error for the top regex popped up, that I fixed as suggested. Let me know if that's all right.


Mohammer5:
Nice, can you also remove the `hasChildren` prop that's not used and causes the checks to fail?

ismay:
> Nice, can you also remove the `hasChildren` prop that's not used and causes the checks to fail?

N.p. I just wanted to double check with you that I wasn't there for a reason I was overlooking. Removed it.

Issue #30, opened 2020-04-22
Menu refactor


HendrikThePendric:
# Intro
The current version of the Menu and related components have a few issues that could be improved. Some changes were made in the past with a big focus on maintaining backwards compatibility (https://github.com/dhis2/ui-core/pull/297 and https://github.com/dhis2/ui-core/pull/354). We discussed [on Slack](https://dhis2.slack.com/archives/CBM8LNEQM/p1584539404115300) that some more fundamental improvements couldn't be done without introducing breaking changes, and since we are now building up to a next major version we can now break 👹 and improve stuff. I'll go through the aspects of these components where there's room for improvement and conclude with a list of components I think we should implement for v5. 

I had previously began to discuss various options for this with @Mohammer5 [on Slack](https://dhis2.slack.com/archives/CBM8LNEQM/p1585037257085800).

## Making the menu click based
The current implementation responds to hover, and [we discussed](https://dhis2.slack.com/archives/CBM8LNEQM/p1584435230062800) this needs to be changed to a click based implementation. For a click based menu we'd expect the following functionality:
1. Given that a menu has multiple sub-menus (at one level) and one is open, we'd expect that one to close when another sub-menu is opened (toggling). To achieve this we need to control which sub-menu is open from the parent menu.
2. When the user clicks outside of the menu, we'd expect all of the open sub-menus to close. Looking at [the design specs](https://github.com/dhis2/design-system/blob/master/molecules/menu.md), this behaviour doesn't have to be provided by the menu itself. Instead, the app would render a menu in a Popover and when the Popover's backdrop-layer is clicked would hide the Popover and the Menu as well.

## Distinguishing between a menu-item and a sub-menu
Currently we only have a `MenuItem` component which has various props that determine the content of the menu-item, such as `label` and `icon`. And if the `MenuItem` has `children` then we assume it's a sub-menu. A dedicated `SubMenu` component would make things a lot more explicit, provide a clear separation of concerns, and would allow for the MenuItem to just render its children, which would be more in line with the rest of our components.

## Separating behaviour from style
[...]

PR #44, opened 2020-04-25
Feature/click menu

HendrikThePendric:
This PR implements the Menu and related components according to [the updated design specs](https://www.sketch.com/s/07b991f1-a188-4977-91af-40c51f24ad0a/a/DQAppk) and #30 can be closed when this PR gets merged into alpha. 

**We've ended up with the following components:**
- `MenuList`: an `ul` with appropriate styles that just renders children
- `MenuItem`: a `li` with appropriate styles depending on props
- `MenuSectionHeader`: a `li` with a `Divider` and a `h6` in which the children are rendered
- `MenuDivider`: a `li` with a `Divider`
- `Menu`: renders children into a `Card` with a `MenuList` and controls the toggling of the `SubMenu` open state
- `SubMenu`: renders a `MenuItem` and a fly-out menu, which is toggled on click. It has a `menuItemContent` prop to control the content of the menu-item-label and an `icon` prop to set the icon for the menu-item.

**Some aspects of the design-specs have not been implemented:**
- **Clicking outside the menu closes everything**: I've kept this out-of-scope for the actual `Menu` component. However the "Dropdown Menu" story illustrates how an app could easily implement this behaviour.
- **Custom icon colour**: I don't think we need to add any custom functionality for this. If an SVG with a specific `fill` colour is passed to the `icon` prop the icon will get that colour anyway. FYI: the icon colour for default / destructive / disabled is inherited from the li element. This means that any icon with a fill colour applied to it directly will not get our colour-scheme applied to it. I have included a story to illustrate how to do apply a custom icon colour.
- **Toggle menu item**: I haven't introduced any custom functionality for this either, because I think it's quite straightforward to implement. I've illustrated that in a story too.
- **Keyboard shortcuts section in `MenuItem`**: I reckon we should add support for this once we have a system in place for actual keyboard shortcuts and scopes.

**Some things that I'd like to get your opinion about:**
- I'm using `cloneElement` and some "private props" which are prefixed with an underscore (`_open` and `_toggleSubMenu`) on the `SubMenu`. Not overly excited about this, but the `cloneElement` strategy does seem to be in line with what we do elsewhere. I could also implement this via a menu-context, but that comes with its own set of potential problems. The underscore prefix aspect would be new, but I think in this case it's quite helpful to make the private/internal nature of these props explicit.
- As hinted at above, the custom approach to setting a fill on SVG icons relies on the assumption that the SVGs being used won't have a fill property if they want to follow the Menu's colour-scheme, and that they do have a fill specified if they need a custom colour. I think this is a pretty safe assumption.
- I think that keeping the "clicking outside closes everything" part of the specs out-of-scope for the Menu makes sense. This does mean that currently the menu doesn't support having a "root-menu" in the normal document flow with nested submenus that all close when clicking outside. Adding support for that would be possible without breaking changes. Currently this doesn't seem necessary because the only two valid use-cases @cooper-joe and I identified are these:
    - A fly-out menu that is triggered from some sort of anchor
    - A simple menu-list, for example in a sidebar
- I don't think we need to expose a dedicated `PopupMenu` or `ToggleMenuItem` component because they are so easy to build.

**TODO:**
- [x] Rebase on alpha and fix merge issues
- [x] Add story for custom icon colour
- [x] Add story for toggle menu item
- [x] Add story sidebar menu list
- [x] Add/tweak e2e tests
- [x] Add/tweak stories (organise per component)
- [x] Add docs
- [x] Remove old files
- [x] Organise new files
- [x] Export everything
- [x] Squash commits

**TODO - after PR review meeting**
- [x] Restructure and remove SubMenu
- [x] Squash commits and adjust BREAKING CHANGE commit-message-section to reflect new structure


packages/core/src/Menu/Menu.js
```diff
@@ -12,54 +12,89 @@ import { MenuList } from '../MenuList/MenuList.js'
...
+                              })
+                            : child
+                    )}
+                </MenuList>
+            </Card>
+
+            <style jsx>{`
+                div {
+                    display: inline-block;
+                    min-width: 128px;
```
Mohammer5, line 65: This should be over-writable as well. If I want to use the menu in the sidebar, I'd want it to always take 100% of the space. So either we add a `useFullWidth` prop (just an ad hoc name) or we allow `min-width` to be changed to `100%`.

packages/core/src/MenuItem/MenuItem.js
```diff
@@ -1,62 +1,97 @@
...
+
+                <style jsx>{styles}</style>
+            </li>
+            {showSubMenu &&
+                createPortal(
+                    <Popper placement="right-start" reference={menuItemRef}>
+                        <Menu dense={dense}>{children}</Menu>
+                    </Popper>,
+                    node
+                )}
```
Mohammer5, line 93: I think we could move this into the `li`. Technically it wouldn't make any difference, but then in the code, the `li` would be the outer component instead of a fragment (easier to parse)

packages/core/src/MenuItem/MenuItem.js
```diff
@@ -68,32 +103,35 @@ MenuItem.defaultProps = {
...
  * @prop {boolean} [dense]
- * @prop {boolean} [active]
  * @prop {boolean} [destructive]
  * @prop {boolean} [disabled]
- * @prop {string} [dataTest]
+ * @prop {string} [href]
+ * @prop {Element} [icon]
+ * @prop {string} [target]
+ * @prop {string} [value]
+ * @prop {function} [onClick] - Click handler called with `value` in the payload
```
Mohammer5, line 118: Some props are missing

packages/core/src/MenuSectionHeader/MenuSectionHeader.js
```diff
@@ -0,0 +1,70 @@
...
+ * @returns {React.Component}
+ *
+ * @example import { MenuSectionHeader } from '@dhis2/ui
+ *
+ * @see Specification: {@link https://github.com/dhis2/design-system/blob/master/molecules/menu.md|Design system}
+ * @see Live demo: {@link /demo/?path=/story/components-core-menusectionheader--default|Storybook}
+ */
+const MenuSectionHeader = ({ className, dataTest, dense, label }) => (
+    <li data-test={dataTest} className={cx(className, { dense })}>
+        <Divider dense={dense} />
```
Mohammer5, line 20: Should this one also have the `Divider` when it's the first section? I think it'd make sense to wrap this one with a div that's `display: none` when the `li` is `:first-child`

packages/core/src/index.js
```diff
@@ -24,12 +24,15 @@ export { FileListPlaceholder } from './FileListPlaceholder/FileListPlaceholder.j
...
 export { Input } from './Input/Input.js'
 export { Label } from './Label/Label.js'
+export { Layer, useLayerContext } from './Layer/Layer.js'
 export { Legend } from './Legend/Legend.js'
 export { LinearLoader } from './LinearLoader/LinearLoader.js'
 export { Logo, LogoIcon, LogoIconWhite, LogoWhite } from './Logo/Logo.js'
 export { Menu } from './Menu/Menu.js'
+export { MenuDivider } from './MenuDivider/MenuDivider.js'
 export { MenuItem } from './MenuItem/MenuItem.js'
 export { MenuList } from './MenuList/MenuList.js'
```
Mohammer5, line 34: Do we need to export this one?

packages/core/src/package.json
```diff
@@ -0,0 +1,72 @@
+{
```
Mohammer5, line 1: Something wrong happened here. This is the `package.json` that should be in the root folder

ismay:
>I'm not sure the following is an actual issue, but it could become one:
Currently when opening a sub-menu while another one is opened, the previously opened one closes because it's not being rendered anymore. These menus lose their entire state, which could interfere with custom component's logic that store some local state (e. g. whether a specific request was made, it's data, etc). So this could be a bit fragile.
>
>I'm not sure if there's an easy way around this though, maybe display: none would be a better solution than not rendering the sub-menus. I don't think that will trickle down the actual dom tree, so this might require a different approach than the current one.

I think using display none would confuse me more. In my opinion state should not live in a component if it needs to persist beyond that component being rendered, so I wouldn't cater to that usecase. Seems like normal react behaviour to me. If the state should persist, the dev should move it up.

packages/core/src/Menu/Menu.js
ismay: With the select we had a similar issue. We decided to always render full width there. For consistency I think it'd be nice to do the same here. Then we'd only set the width for the popups (as those are independent of the layout).

cypress/integration/Layer/click_behavior.feature
```diff
@@ -1,10 +1,5 @@
 Feature: The Layer has configurable click behaviour
 
-    Scenario: A non-blocking layer
-        Given a Layer with pointerEvents none and a button below it is rendered
-        When the user clicks the button
-        Then the onClick handler of the button is called
-
```
ismay, line 7: Was wondering why this test was removed.

cypress/integration/Menu/position/index.js
```diff
@@ -3,33 +3,36 @@ import { Given, Then, When } from 'cypress-cucumber-preprocessor/steps'
...
+When('the user clicks on the MenuItem', () => {
+    cy.get('[data-test="dhis2-uicore-menuitem"]').click()
 })
 
 Then(
     'the right of the MenuItem is aligned with the left of the SubMenu',
     () => {
         getMenuItemAndSubMenuRects().then(([menuItemRect, subMenuRect]) => {
-            expect(menuItemRect.right).to.equal(subMenuRect.left)
+            expect(menuItemRect.right).to.be.approximately(
```
ismay, line 32: Was wondering why this can't be asserted exactly. With the Select tests I'm working on in another PR the assertions with popper placement seem to work with exact comparisons.

ismay: Also, if my findings in the other PR are correct, these tests will probably have to be updated to at least use `should` so that cypress keeps trying for a while instead of asserting once.

cypress/integration/Menu/toggles_submenus.feature
```diff
@@ -0,0 +1,9 @@
+Feature: The Menu toggles SubMenus
+
+    Scenario: a Menu with two SubMenus
+        Given a Menu with two SubMenus is rendered
+        When the user clicks the first SubMenu anchor
+        Then first SubMenu is visible
+        When the user clicks the second SubMenu anchor
+        Then second SubMenu is visible
+        And the first SubMenu is not visisble
```
ismay, line 9: visisble

packages/core/src/Menu/Menu.js
```diff
@@ -12,54 +12,89 @@ import { MenuList } from '../MenuList/MenuList.js'
...
+    const toggleSubMenu = index => {
+        const toggleValue = index === openedSubMenu ? null : index
+        setOpenedSubMenu(toggleValue)
+    }
+
+    const cardCSS = resolve`
+        padding: ${spacers.dp4} 0;
+        max-height: ${maxHeight};
+        overflow-y: auto;
+    `
```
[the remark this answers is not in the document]
ismay: I think that this is in the render function because it uses maxHeight. You mean that the static part should be outside of render()?

packages/core/src/MenuSectionHeader/MenuSectionHeader.js
ismay: Yeah it seems like the first shouldn't have that. I would prefer not rendering over display none if possible. It seems like the poor mans version of not rendering.

ismay:
The new api looks good to me. Would it be possible to add an extra story to `Menu` that integrates all newly introduced components? They're now quite isolated. I'd like to see how they'd all be used together in a realistic scenario, both as a user and as a reviewer.

That'd help me take a more in-depth look at the code, with the intended use-case in mind.

packages/core/src/Menu/Menu.js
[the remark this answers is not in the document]
Mohammer5: Ah, I didn't see that.. yeah, it's fine to stay inside

HendrikThePendric:
> When clicking an item that does not have a sub menu, there's still an empty menu card opened.

Resolved in 8c40e03

----

> I'm not sure the following is an actual issue, but it could become one:
Currently when opening a sub-menu while another one is opened, the previously opened one closes because it's not being rendered anymore. These menus lose their entire state, which could interfere with custom component's logic that store some local state (e. g. whether a specific request was made, it's data, etc). So this could be a bit fragile.
> 
> I'm not sure if there's an easy way around this though, maybe display: none would be a better solution than not rendering the sub-menus. I don't think that will trickle down the actual dom tree, so this might require a different approach than the current one.

I did not address this. I agree it _could_ be or become a problem, but the current approach is in line what we do in other places too, i.e. `Modal`, `Popover`, etc.  I'd argue that if we'd decide to change this, we'd have to address it for all these other components as well.

For now, I think this is the most clean approach. And I'd suggest we keep advising app-devs to create stateful wrapper components to solve these type of issues. IMO we should only address this once we start seeing problems that can't be solved using some wrapper.

Oh, I just noticed @ismay already commented on this too:
> I think using display none would confuse me more. In my opinion state should not live in a component if it needs to persist beyond that component being rendered, so I wouldn't cater to that usecase. Seems like normal react behaviour to me. If the state should persist, the dev should move it up.

I completely agree with that 👍 

Let me know if you are convinced @Mohammer5 😄 

----

These were my responses to the general comments that needed a reply/fix. The rest of the comments are about specific sections of code, so I can reply in thread. I'll start doing that now.

packages/core/src/Menu/Menu.js
HendrikThePendric: In the `MenuList` demo stories I've included an example for creating a sidebar menu. The point is, you'd use the `MenuList` for that, and not the `Menu`. The `MenuList` just occupies the full width of its parent.

To me this makes sense, and I even think that is why @Mohammer5 extracted the `MenuList` from the `Menu` a long time ago. See https://github.com/dhis2/ui-core/pull/297

HendrikThePendric: > `min-width: 128px;`
> This should be over-writable as well. 

I was actually going to point you to the [design specs](https://www.sketch.com/s/07b991f1-a188-4977-91af-40c51f24ad0a/a/DQAppk) to show you that these mention a fixed `min-width`. Not sure if that has been changed or if I just got confused. I'll check with @cooper-joe

packages/core/src/MenuItem/MenuItem.js
HendrikThePendric: Personally, I think using a fragment here signals very clearly that these components/elements don't have a parent-child relationship.

This is also in line with what we do pretty much everywhere we use a `createPortal` and `<Layer />` (which uses createPortal internally).

packages/core/src/Menu/Menu.js
HendrikThePendric: Aha, yes there are the fixed sizes, they were slightly hidden in a different place they used to be, because dense and normal are now slightly different:
image (2)
image (1)

It used to be a single fixed min-width, but now dense and normal have their own. FIxed that in 7ae718a.

packages/core/src/MenuItem/MenuItem.js
HendrikThePendric: Fixed in 5865a86

packages/core/src/index.js
HendrikThePendric: Just checked with @cooper-joe, yes we do.

packages/core/src/MenuSectionHeader/MenuSectionHeader.js
HendrikThePendric: I've made this conditional via a `hideDivider` prop in b358c47. Lemme know if that's OK.

packages/core/src/package.json
HendrikThePendric: 👍 removed in cf7f466

cypress/integration/Layer/click_behavior.feature
HendrikThePendric: Good question. This was done intentionally:
- When I was working on the `Layer` component I kept thinking that sometimes it would be useful to have a concept of a "non-blocking layer", meaning that elements would be stacked correctly along the z-axis, but that the user could still interact with the underlying layers.
- However, when I worked on the Menu, this was the first time I actually encountered a scenario like that in real-life. I initially tried it with a `<Layer pointerEvents="none" />`. This worked fine, but I felt that this prop on the Layer was very quirky, and since the actual Layer element was still being created, I also thought it was just the wrong solution to the problem I was trying to solve.
- It occurred to me that what I actually wanted, was to just be able to append something to its parent layer, so if I just had access to the parent layer node, that would be enough.

So what I've done is:
- Remove everything related to "non-blocking layers" from the Layer component
- Export a `useLayerContext` hook that is also exposed in `index.js`
- Refactor the MenuItem (as sub-menu) to just use `createPortal` and append the portal to the layer node which is returned from `useLayerContext`

Let me know if that makes sense...

cypress/integration/Menu/position/index.js
HendrikThePendric: I happened to be getting floats here were ever so slightly different. Not quite sure why.

I agree that we _should_ be using `should` 🤣 but shall we address that more holistically when we tackle issues #87 & #86? 
_(bearing in mind that we want to release v5 ASAP so we can freeze the ui-core and ui-widgets repos)_

Or should I just fix this now?

cypress/integration/Menu/toggles_submenus.feature
HendrikThePendric: fixed in b066403

Mohammer5:
> Let me know if you are convinced @Mohammer5

Yes, I wasn't trying to force a change, just a discussion / create awareness

packages/core/src/MenuSectionHeader/MenuSectionHeader.js
Mohammer5: I think I'm fine with that. Just one thought that we could discuss:

As probably all menu sections that are the first child shouldn't show the divider, we'd force the developers to always provide a prop to the first component, which seems like an odd dev practice. Maybe there are better solutions for this. For example the component that will eventually contain the section (I think that's the `Menu` component that already clones the children?) could also pass a `hideDivider: true` to the first child (at `index === 0`) if the component does not have an explicit value for that prop (`typeof child.props.hideDivider === 'undefined'`)

packages/core/src/Menu/Menu.js
Mohammer5: > and I even think that is why @Mohammer5 extracted the MenuList from the Menu a long time ago

I don't remember that :rofl: 

> In the MenuList demo stories I've included an example for creating a sidebar menu. The point is, you'd use the MenuList for that, and not the Menu. The MenuList just occupies the full width of its parent.

I think we have to work on the names then as they don't convey their intended usecases. Terminology-wise, I'd expect to use a `Menu`, when I need a `Menu`. If I had to use a `MenuList` as a menu, I'd expect it to be a list of menus. It does make sense to have a `MenuList` as an internal component so we wrap everything with an `ul`, but once it becomes public or even has different use-cases, then this is a rather confusing name.

So the `Menu` is actually a `PoppableMenu` and the `MenuList` is a `MenuItemContainer`. These are just ad hoc names, not meant as serious names, but they illustrate what I want to say: They convey a very clear intended usage, while `Menu` and `MenuList` do not.

HendrikThePendric:
> That'd help me take a more in-depth look at the code, with the intended use-case in mind.

I've added one in b157bbe, but beware:
- `MenuList` is not in there, as Menu renders that internally
- I've interpreted your request very literally, and focussed purely on component structure / nesting. I intentionally didn't include any variations for MenuItem based on props, because I think these are better demoed in isolation as they already are.

packages/core/src/MenuSectionHeader/MenuSectionHeader.js
HendrikThePendric: I like that 👍 
Added in df49ce8. I'm only setting `hideDivider` to `true` under the following condition:
```javascript
child.props.hideDivider !== 'boolean' && index === 0
```
So if for whatever reason the developer needs a divider above the first header, he can explicitly set `hideDivider={false}`

HendrikThePendric:
> This lacks a typeof
Well spotted, fixed in 35370e4

HendrikThePendric:
@Mohammer5 and @ismay: as discussed, the I've renamed `Menu` to `FlyoutMenu`. I think this is good to get merged into alpha now.

cypress/integration/Menu/position/index.js
ismay: Addressing it later seems fine to me 👍

cypress/integration/Layer/click_behavior.feature
ismay: Judging from your explanation during our call this would align the behaviour with our other components (open state blocks entire UI). Good change in my opinion 👍

PR #49, opened 2020-04-29
merge: update core

varl:


jest.config.js
```diff
@@ -2,4 +2,8 @@ module.exports = {
     rootDir: '.',
     setupFilesAfterEnv: ['<rootDir>config/jest/enzymeSetup.js'],
     roots: ['<rootDir>/packages'],
+    testPathIgnorePatterns: [
+        '/node_modules/',
+        'Transfer/__tests__/common/createChildren.js',
```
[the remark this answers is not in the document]
varl: I **do not** want to have the `jest.config.js` change (`testPathIgnorePatterns`) in `master`.

I did a "best effort" merge last night since you mentioned that the Transfer will need some refactoring once it's moved to UI. This should be part of that.

varl:
The parts I noticed:

1. We are moving away from `{ value, label }` objects for selected values and instead using string values.

2. The multiselect proptypes is gone, there's a `TODO` where I've copied in the old shared prop types for multiselect, see point 1.

3. Transfer uses InputField, so technically it cannot be a core component. The choice was between moving Transfer to widgets or refactoring to use Field+Input. I did the latter in this PR, but maybe we want it to be a widget (with translations, e.g.)

4. I'm not fond of the `common` pieces. I had a hard time keeping track if I was looking at `Transfer/__e2e__/common.js`, `__tests__/common/*`, or `Transfer/common.js`.

5. Please specify the extension when importing (this is the code style in UI !). Using `import { foo } from 'directory'` or `import { foo } from 'file'` to import from `directory/index.js` or `file.js` is not ES compatible -- it is a Node-ism that is supported by Webpack. Pure ES tools do choke on that unless it is Babelified or handled later. Would be nice to have a ESLint for this, as it makes it a lot easier to understand exactly where the code is imported from.

6. The `__tests__/common` has several options:
  - Add a private package to `packages/test-utils`, then we can import `createChildren` from `ui-test-utils` (I like this one).
  - Create a folder in `packages/core/src` named `utils` and put `createChildren.js` there.
  - Don't automate this.

packages/core/src/Transfer/TransferOption.js
```diff
@@ -0,0 +1,115 @@
...
+                    cursor: not-allowed;
+                }
+            `}</style>
+        </div>
+    )
+}
+
+TransferOption.defaultProps = {
+    dataTest: 'dhis2-uicore-transferoption',
+    addDecorator: {},
```
Mohammer5, line 88: I have no idea how that one got there.. It's also in `ui-core`.
There's no such prop, so this can be removed

packages/core/src/Transfer/Transfer.stories.js
```diff
@@ -0,0 +1,454 @@
+/* eslint-disable react/prop-types */
+import React, { useState } from 'react'
+
+import {
+    SingleSelectField,
```
Mohammer5, line 5: This has to be imported from `@dhis2/ui-widgets`

varl: Good catch! :gloves:

varl: That's another one of those "we cannot import ui-widgets stuff in ui-core" problems, I'll solve it the same way I solved the InputField.

packages/core/src/Transfer/TransferOption.js
varl: 7df9ef76

PR #60, opened 2020-05-07
refactor(transfer): align with select & monorepo structure

### LINKED ISSUE #50: Post-Transfer merge improvements
The parts I noticed:

1. We are moving away from `{ value, label }` objects for selected values and instead using string values.

2. The multiselect proptypes is gone, there's a `TODO` where I've copied in the old shared prop types for multiselect, see point 1.

3. Transfer uses InputField, so technically it cannot be a core component. The choice was between moving Transfer to widgets or refactoring to use Field+Input. I did the latter in this PR, but maybe we want it to be a widget (with translations, e.g.)

4. I'm not fond of the `common` pieces. I had a hard time keeping track if I was looking at `Transfer/__e2e__/common.js`, `__tests__/common/*`, or `Transfer/common.js`.

5. Please specify the extension when importing (this is the code style in UI !). Using `import { foo } from 'directory'` or `import { foo } from 'file'` to import from `directory/index.js` or `file.js` is not ES compatible -- it is a Node-ism that is supported by Webpack. Pure ES tools do choke on that unless it is Babelified or handled later. Would be nice to have a ESLint for this, as it makes it a lot easier to understand exactly where the code is imported from.

6. The `__tests__/common` has several options:
  - Add a private package to `packages/test-utils`, then we can import `createChildren` from `ui-test-utils` (I like this one).
  - Create a folder in `packages/core/src` named `utils` and put `createChildren.js` there.
  - Don't automate this.

_Originally posted by @varl in https://github.com/dhis2/ui/pull/49#issuecomment-621666104_

Mohammer5:
Fixes #50

This commit does several things:
* Use strings as selected values
* Use strings to store which options are highlighted
* Removes jest test abstractions
* Simplifies internal names
* Replaces `common.js` files with files whose names match their content
* Moves the Transfer component from `core` to `widgets`
* Adds the `.js` extension to all import statements where applicable
* Provides the correct `disabled` & `dataTest` props to the re-order up
button's icon

BREAKING CHANGE: The Transfer component now expects options to be passed in as objects, not as children. Custom components can be provided via the `optionComponent` prop for all options or via the `component` property on an individual option.

BREAKING CHANGE: The Transfer component now expects strings as selected
values instead of option objects.

BREAKING CHANGE: The Transfer component is now part of `widgets`


Mohammer5:
image

:+1:

Mohammer5:
@HendrikThePendric The multi select feature failed, it does not fail locally. I assume that's another issue with the `requestAnimationFrame` timing, right?

ismay:
> @HendrikThePendric The multi select feature failed, it does not fail locally. I assume that's another issue with the `requestAnimationFrame` timing, right?

Yeah I think we need to take another at the select. I think we can find another way to deal with this than using RAF, and with the recent refactors we've had of the select I suspect there are bits of it that need a little love.

HendrikThePendric:
> The multi select feature failed, it does not fail locally. I assume that's another issue with the requestAnimationFrame timing, right?

Oh man 🤦 that's very unfortunate. It's not really an issue with `requestAnimationFrame`, but rather with the `ResizeObserver`. I had already implemented a helper called `waitForResizeObserver` to fix it, but apparently that isn't super robust either. I really don't know of another way to address this problem.

ismay:
> > The multi select feature failed, it does not fail locally. I assume that's another issue with the requestAnimationFrame timing, right?
> 
> Oh man 🤦 that's very unfortunate. It's not really an issue with `requestAnimationFrame`, but rather with the `ResizeObserver`. I had already implemented a helper called `waitForResizeObserver` to fix it, but apparently that isn't super robust either. I really don't know of another way to address this problem.

It's weird, I tried rerunning the integration tests, but now they're failing completely:

Screenshot 2020-05-11 at 11 09 52

HendrikThePendric:
I've had a quick attempt at reviewing these PRs but I got the distinct impression that the vast majority was just a matter of moving things around. @Mohammer5 would it be possible to indicate where you have actually changed an existing implementation?

Mohammer5:
yeah, so most work has bee done on functions that use the value of the `selected` prop or a derived value of that prop. It's quite a few files as that was the biggest change because two things have changed at the same time:

1. Use strings instead of objects
2. Use `options` instead of `children` prop

If you want to, we could schedule a meeting where I go over what's changed @HendrikThePendric @ismay

packages/widgets/src/Transfer/Transfer.js
```diff
@@ -75,9 +64,9 @@ export const multiSelectedPropType = propTypes.arrayOf(singleSelectedPropType)
  * rest of the library
  */
 export const Transfer = ({
```
HendrikThePendric, line 66: I've been thinking about rendering custom options and how we could stay a little bit more consistent here with other components that may require customisation. It's quite difficult to come up with something that is exactly the same, since this `options` prop pattern is fundamentally different than the `children` pattern we use elsewhere. After chatting a little bit with @ismay I came up with an idea (not sure whether Ismay will fully agree).

I'll place a few comments in the component code to list the changes I'd like to suggest.

packages/widgets/src/Transfer/Transfer.js
```diff
@@ -96,6 +85,7 @@ export const Transfer = ({
     leftFooter,
     leftHeader,
     maxSelections,
+    optionComponent: TransferOption,
```
HendrikThePendric, line 88: change to an optional prop called `renderOption` of type function

packages/widgets/src/Transfer/Transfer.js
```diff
@@ -202,13 +201,28 @@ export const Transfer = ({
...
+                                disabled={option.disabled}
+                                option={option}
+                                highlighted={highlighted}
+                                {...getOptionClickHandlers(
+                                    option,
+                                    selectSingleOption,
+                                    toggleHighlightedSourceOption
+                                )}
+                            />
+                        )
```
HendrikThePendric, line 224: Change this part to this
```jsx
const renderFn = option.render || renderOption
const optionProps = {
    disabled: option.disabled,
    option: option,
    highlighted: !!highlightedSourceOptions.find(
        highlightedSourceOption =>
            highlightedSourceOption === option.value,
        ...getOptionClickHandlers(
            option,
            selectSingleOption,
            toggleHighlightedSourceOption
        )
    ),
}

return typeof renderFn === 'function' ? (
    renderFn(optionProps)
) : (
    <TransferOption {...optionProps} />
)
```

packages/widgets/src/Transfer/Transfer.js
```diff
@@ -378,16 +399,23 @@ Transfer.defaultProps = {
...
+ * @prop {string[]} selected
  * @prop {string} [selectedWidth]
  * @prop {Function} [onFilterChange]
  */
 Transfer.propTypes = {
+    options: propTypes.arrayOf(
+        propTypes.shape({
+            label: propTypes.string.isRequired,
+            value: propTypes.string.isRequired,
+            component: propTypes.func,
```
HendrikThePendric, line 411: change to `render`

packages/widgets/src/Transfer/Transfer.js
```diff
@@ -401,15 +429,13 @@ Transfer.propTypes = {
     leftFooter: propTypes.node,
     leftHeader: propTypes.node,
     maxSelections: propTypes.oneOf([1, Infinity]),
+    optionComponent: propTypes.func,
```
HendrikThePendric, line 432: change to `renderOption`

ismay:
@HendrikThePendric and I talked a bit about this. In general we're thinking of the same thing, though my preference would be a slightly more conservative approach. In pseudo code I'm thinking of:

```jsx
// Allow the user to override the default option rendering
renderOption = option.renderOption || defaultOptionRenderer

// option rendering just maps over all the options,
// and calls renderOption with the args needed for rendering options
// renderOption is expected to return (react) renderable content
options.map(option => {
  renderOption({ option, etc. })
})
```

So what I'm omitting compared to @HendrikThePendric's approach is allowing the render prop on an option in the options array. the way I see it, you can already implement everything you want in the `renderOption` callback, so I'd omit that for the sake of simplicity. We could always add it later on if it turns out to be necessary.

HendrikThePendric:
We also briefly discussed the option of adhering to the `children` API for v5, but not sure that is feasible at all.... Perhaps we can discuss things again tomorrow?

ismay:
> We also briefly discussed the option of adhering to the `children` API for v5, but not sure that is feasible at all.... Perhaps we can discuss things again tomorrow?

Yeah that'd be nice 👍. If we can align on everything tomorrow then we could maybe finish up the remaining tasks for v5 this week.

Mohammer5:
I've introduced this the way @ismay proposed because there's a performance implication. The prop should either always expect a component or a function.

I've removed the custom option functionality from the individual option items as that can be done through the `renderOption` function. We can easily add it back if ever needed.

PR #61, opened 2020-05-07
feat(widgets): add default translated texts

ismay:
This adds default, translated texts for our widgets, where appropriate. So say, a placeholder for a filter input. Usually devs will leave those types of placeholders as is, so it makes sense for us to set a proper default.

Since we're not relying on the namespace being set correctly when the modules are parsed, we're delaying the call to i18n.t by allowing a function as prop for the translatable texts. A potential follow up: do we want to allow that for all our string/renderable text props, so that our users can do the same?

Currently I'm just changing things in widgets, so core components are not affected for the time being.

See also: https://dhis2.slack.com/archives/CBM8LNEQM/p1588767905388900 (translations)

Todo:

- [x] multiselect
- [x] singleselect
- [x] What do we do with props that accept a node (like the Select's empty prop)?
- [x] organisationunittree (doesn't seem like it needs translations?)
- [x] Add stories demonstrating texts
- [x] Add tests (verify that the translations work correctly)
- [x] Update jsdoc & prop-types
- [x] Verify the translations I've chosen (ask Joe)
- [x] Check if we can interpolate the filter noMatch text (unfortunately not with the current strategy)
- [x] Remove requiredif's everywhere
- [x] FileInput (verify translations, add demos and https://github.com/dhis2/ui/issues/70, technically there are duplicate translations, but they do make the components a bit more independent maybe?)

Follow up:

- [ ] allow func for all our string props?
- [ ] transfer?
- [ ] follow up fileinput (Cancel link (Cancel) & Loading state (Uploading))

### DIFF
--- a/cypress/integration/FileInputField/has_default_button_label.feature
+++ b/cypress/integration/FileInputField/has_default_button_label.feature
@@ -1,0 +1,5 @@
+Feature: Button label for the FileInputField
+
+    Scenario: Rendering a FileInputField
+        Given a default FileInputField is rendered
+        Then the default button label is visible
--- a/cypress/integration/FileInputField/has_default_button_label/index.js
+++ b/cypress/integration/FileInputField/has_default_button_label/index.js
@@ -1,0 +1,9 @@
+import { Given, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given('a default FileInputField is rendered', () => {
+    cy.visitStory('FileInputField', 'Default')
+})
+
+Then('the default button label is visible', () => {
+    cy.contains('Upload a file').should('be.visible')
+})
--- a/cypress/integration/FileInputField/has_default_placeholder.feature
+++ b/cypress/integration/FileInputField/has_default_placeholder.feature
@@ -1,0 +1,5 @@
+Feature: Placeholder for the FileInputField
+
+    Scenario: Rendering a FileInputField
+        Given a default FileInputField is rendered
+        Then the default placeholder is visible
--- a/cypress/integration/FileInputField/has_default_placeholder/index.js
+++ b/cypress/integration/FileInputField/has_default_placeholder/index.js
@@ -1,0 +1,9 @@
+import { Given, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given('a default FileInputField is rendered', () => {
+    cy.visitStory('FileInputField', 'Default')
+})
+
+Then('the default placeholder is visible', () => {
+    cy.contains('No file uploaded yet').should('be.visible')
+})
--- a/cypress/integration/FileInputFieldWithList/has_default_button_label.feature
+++ b/cypress/integration/FileInputFieldWithList/has_default_button_label.feature
@@ -1,0 +1,5 @@
+Feature: Button label for the FileInputFieldWithList
+
+    Scenario: Rendering a FileInputFieldWithList
+        Given a default FileInputFieldWithList is rendered
+        Then the default button label is visible
--- a/cypress/integration/FileInputFieldWithList/has_default_button_label/index.js
+++ b/cypress/integration/FileInputFieldWithList/has_default_button_label/index.js
@@ -1,0 +1,9 @@
+import { Given, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given('a default FileInputFieldWithList is rendered', () => {
+    cy.visitStory('FileInputFieldWithList', 'With default texts')
+})
+
+Then('the default button label is visible', () => {
+    cy.contains('Upload a file').should('be.visible')
+})
--- a/cypress/integration/FileInputFieldWithList/has_default_placeholder.feature
+++ b/cypress/integration/FileInputFieldWithList/has_default_placeholder.feature
@@ -1,0 +1,5 @@
+Feature: Placeholder for the FileInputFieldWithList
+
+    Scenario: Rendering a FileInputFieldWithList
+        Given a default FileInputFieldWithList is rendered
+        Then the default placeholder is visible
--- a/cypress/integration/FileInputFieldWithList/has_default_placeholder/index.js
+++ b/cypress/integration/FileInputFieldWithList/has_default_placeholder/index.js
@@ -1,0 +1,9 @@
+import { Given, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given('a default FileInputFieldWithList is rendered', () => {
+    cy.visitStory('FileInputFieldWithList', 'With default texts')
+})
+
+Then('the default placeholder is visible', () => {
+    cy.contains('No file uploaded yet').should('be.visible')
+})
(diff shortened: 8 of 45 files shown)


Mohammer5:
The PR title is misleading. I was expecting all components in widgets to have been translated. Was that your original plan and you just wanted to confirm the approach before translating the remaining components?

ismay:
> The PR title is misleading. I was expecting all components in widgets to have been translated. Was that your original plan and you just wanted to confirm the approach before translating the remaining components?

It's still a draft, I'm still working on it.

Mohammer5:
I like the approach and wouldn't know a better way right now.
The only thing that bugs me a bit is that the default values' types don't match the prop types.

ismay:
> I like the approach and wouldn't know a better way right now.
> The only thing that bugs me a bit is that the default values' types don't match the prop types.

Yeah I haven't updated the jsdoc and the proptype for `empty` yet, but the rest should be good:

`propTypes.oneOfType([propTypes.string, propTypes.func])`

ismay:
I'm still thinking how I want to tackle the translating for the empty prop, since that expects either a component or a string. Plus it adds styling if it's just a string. So I'm still seeing what would work best there (and for other similar props).

Mohammer5:
Maybe sth like this:

```js
const translation = translateFn =>
  translateFn.isTranslation = true

const translate = value =>
  value.isTranslation ? value() : value

const Comp = ({ empty, label }) => {
  const Empty = translate(empty)

  return (
    <div>
      <Label>{translate(label)}</Label>
      {typeof Empty === string ? Empty : <Empty />}
    </div>
  )
}

Comp.defaultProps = {
  empty: translation(() => i18n.t('Empty!')),
  label: translation(() => i18n.t('Label!')),
}

ismay:
> Maybe sth like this:
> 
> ```js
> const translation = translateFn =>
>   translateFn.isTranslation = true
> 
> const translate = value =>
>   value.isTranslation ? value() : value
> 
> const Comp = ({ empty, label }) => {
>   const Empty = translate(empty)
> 
>   return (
>     <div>
>       <Label>{translate(label)}</Label>
>       {typeof Empty === string ? Empty : <Empty />}
>     </div>
>   )
> }
> 
> Comp.defaultProps = {
>   empty: translation(() => i18n.t('Empty!')),
>   label: translation(() => i18n.t('Label!')),
> }
> ```

I was thinking that the approach used for the string props wouldn't work for the node props, but actually it works just as well there.So I'm just using the same approach there now.

ismay:
Hey @cooper-joe, I was wondering what you think of the translations. These are the ones I've gone with for the Single- and MultiSelectField:

Screenshot 2020-05-12 at 13 46 08

cooper-joe:
> Hey @cooper-joe, I was wondering what you think of the translations.

Thanks for the ping! I'd like to request some minor changes:

`empty` -> "No data found"
`filterPlaceholder` -> "Type to filter options"
`noMatchText` -> "No options found"

`noMatchText`: An aside: I remember when this component was being built I requested a text here that would insert the filter term. For example if I filter by string `dog` and there are no matches the result would be "Nothing found for dog". Is this possible now we're adding translations to these components in ui-widgets? It's a handy way to give the user some feedback, answering any question ("Why is there nothing found...?") before it comes up.

HendrikThePendric:
Great work @ismay, glad to see these are all functions so we won't run into the issues we encountered last week in many repos.

The way we have organised the components has changed quite a bit lately so I'm not completely sure where we'd want to add this.... But according to our design system, the `FileInput` needs quite a few default texts too:
- Button label (Upload a file / Upload files)
- Remove link (Remove)
- Empty list placeholder (No file(s) selected yet)

In the design specs there are more default texts to be found:
- Cancel link (Cancel)
- Loading state (Uploading)

But I think these two are for later on, when we introduce a component that allows us to interact with the file-resources endpoint, i.e. with async functionality.

ismay:
> `noMatchText`: An aside: I remember when this component was being built I requested a text here that would insert the filter term. For example if I filter by string `dog` and there are no matches the result would be "Nothing found for dog". Is this possible now we're adding translations to these components in ui-widgets? It's a handy way to give the user some feedback, answering any question ("Why is there nothing found...?") before it comes up.

@joe Yeah you're right, I remember that we had it and removed it because there were concerns about interpolation with RTL languages. I'm going to test this in the scheduler, see what the issue is exactly.

I've addressed these points btw:

- empty -> "No data found"
- filterPlaceholder -> "Type to filter options"
- noMatchText -> "No options found"

ismay:
> **Could you add a ToDo item for that too?**

@HendrikThePendric N.p. 👍 I've added it to the top comment

ismay:
> `noMatchText`: An aside: I remember when this component was being built I requested a text here that would insert the filter term. For example if I filter by string `dog` and there are no matches the result would be "Nothing found for dog". Is this possible now we're adding translations to these components in ui-widgets? It's a handy way to give the user some feedback, answering any question ("Why is there nothing found...?") before it comes up.

So, this took a while, and I don't have a definitive answer, but it seems safe enough. See here: https://github.com/dhis2/notes/issues/103. So we could use an interpolation and use `Nothing found for {{ search term }}`. Is that the translation you'd prefer @cooper-joe?

cooper-joe:
> So we could use an interpolation and use Nothing found for {{ search term }}. Is that the translation you'd prefer @cooper-joe?

Great! Yes, let's use that for the default. I think it's best to include the search term inside single quotes to highlight it as user-content: `Nothing found for 'malaria'`.

ismay:
> > So we could use an interpolation and use Nothing found for {{ search term }}. Is that the translation you'd prefer @cooper-joe?
> 
> Great! Yes, let's use that for the default. I think it's best to include the search term inside single quotes to highlight it as user-content: `Nothing found for 'malaria'`.

Actually, I was wrong, sorry. It's not possible to include the filter as that's not known at the point where we define the default. Unless we change our translation strategy, but that would be a major change. I'll leave it as is, if you're all right with that as well.

cooper-joe:
> Actually, I was wrong, sorry. It's not possible to include the filter as that's not known at the point where we define the default. Unless we change our translation strategy, but that would be a major change. I'll leave it as is, if you're all right with that as well.

Ah ok. I see the problem. In that case, let's go for: `No options found`.

ismay:
@cooper-joe I've added three more default translations for the FileInputField and the FileInputFieldWithList:

1. `Remove` (for the label to remove a file)
2. `Upload a file` (for the button label)
3. `No file uploaded yet` (as a placeholder)

Let me know what you think.

cooper-joe:
> @cooper-joe I've added three more default translations for the FileInputField and the FileInputFieldWithList:
> 
>     1. `Remove` (for the label to remove a file)
> 
>     2. `Upload a file` (for the button label)
> 
>     3. `No file uploaded yet` (as a placeholder)
> 
> 
> Let me know what you think.

Those look good to me @ismay 👍

PR #65, opened 2020-05-12
fix(button): use correct padding when displaying only the icon

Mohammer5:
This PR will

* fix the padding issue when displaying an icon but no label
* "fix" the cypress test, waiting for the button to be focused
* remove the `ButtonBase` component and put the code back to the `Button` component


packages/core/src/Button/Button.styles.js
```diff
@@ -1,23 +1,228 @@
 import css from 'styled-jsx/css'
```
ismay, line 1: We have quite some duplicate selectors in this file. I would love it if we could deduplicate them, would really simplify reading through the styles.

Mohammer5:
@cooper-joe I was looking at the styles that @ismay was talking about. I thought about replacing the values for the colors with values from our colors/theme palette, but they don't match. Some are just in between the colors while others are just slightly off.

Some gradients have two different colors, but the difference is so minimal, you really need to know that there's a difference otherwise you just don't see it, like in this rule:

```css
background: linear-gradient(180deg, #b81c1c 0%, #b80c0b 100%);
```

Could you have a quick look at the `Button.styles.js` file and if you think these are the correct colors, I'll just keep them as they are, otherwise we could replace them with actual colors from our color palette / theme

cooper-joe:
> @cooper-joe I was looking at the styles that @ismay was talking about. I thought about replacing the values for the colors with values from our colors/theme palette, but they don't match. Some are just in between the colors while others are just slightly off.
> 
> Some gradients have two different colors, but the difference is so minimal, you really need to know that there's a difference otherwise you just don't see it, like in this rule:
> 
> ```css
> background: linear-gradient(180deg, #b81c1c 0%, #b80c0b 100%);
> ```
> 
> Could you have a quick look at the `Button.styles.js` file and if you think these are the correct colors, I'll just keep them as they are, otherwise we could replace them with actual colors from our color palette / theme

I don't think it's an issue that the buttons do not use all `ui` colors. The way I see it these are 'atoms' that do not necessarily inherit anything. The `ui` colors are intended for the interface and I didn't want to force consistency of usage and sacrifice the core button look. If I remember my experiments correctly an example issue would be: background of a page uses `grey100`, if buttons also use `grey100`->`grey200` gradient then everything was too monotonous. 

However, that gradient looks odd, almost no change as you said, so I'll take another look at the button styles to make sure all is as intended.

cooper-joe:
I checked the button colors/styles - all matches the design system specs so this PR looks good to me. Some of the gradients are very subtle, this is mostly on hover styles for `primary` and `destructive` types. I don't see a problem with leaving the styles as they are now unless it would benefit performance wise. As mentioned above atoms are not required to use colors defined in the `theme`.

ismay:
I wouldn't mind having a go at deduplicating these styles if there's time in the future. As it stands it's quite a massive amount of styles and as noted above, a lot of duplicate selectors. How would you all feel about creating an issue for this? I think it'd improve DX.

Mohammer5:
@ismay I tried refactoring this a little last week and I feel like deduplication doesn't really do much here. Also, as it's not being tested right now, it's potentially "dangerous" as we might overlook which styles override other styles, i. e. a button could be `destructive` and `destructive.disabled`, they have a shared `border-color` style, so we could change it to

```css
.destructive, .destructive.disabled {
  border-color: //...
}
```

but then `.destructive.disabled` would NOT overwrite `destructive.active` anymore.

ismay:
> @ismay I tried refactoring this a little last week and I feel like deduplication doesn't really do much here. Also, as it's not being tested right now, it's potentially "dangerous" as we might overlook which styles override other styles, i. e. a button could be `destructive` and `destructive.disabled`, they have a shared `border-color` style, so we could change it to
> 
> ```css
> .destructive, .destructive.disabled {
>   border-color: //...
> }
> ```
> 
> but then `.destructive.disabled` would NOT overwrite `destructive.active` anymore.

Yeah I mean the brittleness is why I'd prefer eventually addressing it. To me duplicate selectors are a bit of an anti-pattern. So maybe a proper follow up issue would be: 

- add tests
- simplify styles

I also wouldn't address it in this PR, seems out of scope.

Mohammer5:
@HendrikThePendric any idea why the fileinput test would fail?

HendrikThePendric:
> any idea why the fileinput test would fail?

I've looked at the failing test and can't offer a precise answer. But given that the `FileInput` uses the `Button` internally, and this PR includes changes to the `Button`, I think there are two likely culprits:
- The failing `FileInput` test exposes a regression in the `Button` that wasn't tested directly
- The `FileInput` test needs to be updated slightly, perhaps because of changed data-test attribute or sth?

Let me know if you need any help with this

Mohammer5:
Fixed this. I just added a `cy.focused()` so cypress waits until it finds a focused element to avoid race conditions.

Issue #86, opened 2020-05-20
Use less (or simpler) test helpers


HendrikThePendric:
This issue was created as a result of this https://github.com/dhis2/ui/pull/66#issuecomment-631385176. It would be good to start using less helpers in our tests code. I know that quite a lot of components have a `position.feature` file, and all of these make quite extensive usage of a helper. So that's one part that would definetely have to be addressed, but quite likely there's more.

ismay:
To expand a bit on this, in the linked PR I was trying to follow up on the recommendations in this article: https://mtlynch.io/good-developers-bad-tests. Though factoring out often used code makes sense for production code, for tests it can be confusing. What I did in the PR was use a more basic helper, and copy/paste instead of abstracting. That (in my opinion) made the tests easier to read and reason about. I would propose we do this for all our tests.

ismay:
On that note, I've introduced a `getAll` helper, but see that we have an `all` helper as well. At a glance, it seems like the `all` helper is a bit more useful (it allows more than just a get). If the `all` helper is stable I'd recommend removing the `getAll` helper whilst addressing this issue.

HendrikThePendric:
@ismay I'm not sure if we should keep this issue open, and to which milestone we should add it. A lot of problems have been addressed already in https://github.com/dhis2/ui/pull/102, but that PR didn't really focus fully on removing the helper functions.

ismay:
I'd vote to keep it open until it's been fully addressed. There are still test helpers remaining that we could potentially remove. Whether or not we should we should still discuss of course, but until then I'd say we could keep this open.

ismay:
I'll add this to 5.1. Not really relevant to a specific release, but fits as a post major release task.

varl:
https://jira.dhis2.org/browse/LIBS-57

PR #102, opened 2020-05-21
test: use should for assertions

ismay:
Since we still see assertions failing randomly, I thought I'd change the assertions to use .should now instead of later.

I've noticed that when moving to `.should`, tests start failing if there's a `wrap` call in the callback passed to should. Moving the call to `wrap` outside of `should` fixes it. Maybe this is because `wrap` can be considered a side effect.

I've not touched the tests for our forms components. There are calls to `wrap` nested in `should` there, but it's going to be a fair amount of work to separate those as the tests depends a lot on passing around wrapped values.

### DIFF
--- a/cypress/integration/FileInput/common/index.js
+++ b/cypress/integration/FileInput/common/index.js
@@ -8,7 +8,7 @@ Given('the FileInput does not have any files', () => {
 })
 
 Then('the onChange handler is called', () => {
-    cy.window().then(win => {
+    cy.window().should(win => {
         const calls = win.onChange.getCalls()
         expect(calls).to.have.lengthOf(1)
 
@@ -17,7 +17,5 @@ Then('the onChange handler is called', () => {
 
         const payload = callArgs[0]
         expect(Object.keys(payload)).to.include.members(['files', 'name'])
-
-        cy.wrap(payload).as('payload')
     })
 })
--- a/cypress/integration/HeaderBar/the_headerbar_contains_a_profile_menu/the_user_name_and_email_are_displayed.js
+++ b/cypress/integration/HeaderBar/the_headerbar_contains_a_profile_menu/the_user_name_and_email_are_displayed.js
@@ -2,17 +2,18 @@ import '../common/index'
 import { Then } from 'cypress-cucumber-preprocessor/steps'
 
 Then('contains the user name', () => {
-    cy.get('@meFixture').then(({ name }) => {
-        cy.get('[data-test="headerbar-profile-username"]').then($name => {
+    cy.getAll('@meFixture', '[data-test="headerbar-profile-username"]').should(
+        ([{ name }, $name]) => {
             expect($name.text()).to.equal(name)
-        })
-    })
+        }
+    )
 })
 
 Then('contains the user email', () => {
-    cy.get('@meFixture').then(({ email }) => {
-        cy.get('[data-test="headerbar-profile-user-email"]').then($email => {
-            expect($email.text()).to.equal(email)
-        })
+    cy.getAll(
+        '@meFixture',
+        '[data-test="headerbar-profile-user-email"]'
+    ).should(([{ email }, $email]) => {
+        expect($email.text()).to.equal(email)
     })
 })
--- a/cypress/integration/Layer/stacking/index.js
+++ b/cypress/integration/Layer/stacking/index.js
@@ -34,19 +34,35 @@ Given(
 )
 
 Then('the second layer is on top of the first layer', () => {
-    assertLayerIsOnTop('second')
+    cy.get('body').click()
+    cy.window().should(win => {
+        expect(win.onLayerClick).to.be.calledOnce
+        expect(win.onLayerClick).to.be.calledWith('second')
+    })
 })
 
 Then('the alert layer is on top', () => {
-    assertLayerIsOnTop('alert')
+    cy.get('body').click()
+    cy.window().should(win => {
+        expect(win.onLayerClick).to.be.calledOnce
+        expect(win.onLayerClick).to.be.calledWith('alert')
+    })
 })
 
 Then('the layer with level 1001 is on top', () => {
-    assertLayerIsOnTop('1001')
+    cy.get('body').click()
+    cy.window().should(win => {
+        expect(win.onLayerClick).to.be.calledOnce
+        expect(win.onLayerClick).to.be.calledWith('1001')
+    })
 })
 
 Then('the blocking layer is on top', () => {
-    assertLayerIsOnTop('blocking')
+    cy.get('body').click()
+    cy.window().should(win => {
+        expect(win.onLayerClick).to.be.calledOnce
+        expect(win.onLayerClick).to.be.calledWith('blocking')
+    })
 })
 
 Then('the blocking layer is a child of the alert layer', () => {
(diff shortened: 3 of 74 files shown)


cypress/integration/FileInput/common/index.js
```diff
@@ -8,7 +8,7 @@ Given('the FileInput does not have any files', () => {
 })
 
 Then('the onChange handler is called', () => {
-    cy.window().then(win => {
+    cy.window().should(win => {
```
Mohammer5, line 11: This has nothing to do with our PR in particular, but as you're changing the assertion style here and there anyways, it might be a good time to address this:

I think these assertions could be done in a more `cypress`-ish way:
```js
cy.window()
  .its('onChange')
  .should('be.calledOnce')
  .then(stub => stub.getCalls()[0])
  .its('args')
  .should(args => expect(args).to.have.lengthOf(2))
  .and(args => {
    const [payload] = args
    expect(Object.keys(payload)).to.include.members(['files', 'name'])
  })
```

cypress/integration/HeaderBar/the_headerbar_contains_a_profile_menu/the_user_name_and_email_are_displayed.js
```diff
@@ -2,17 +2,18 @@ import '../common/index'
 import { Then } from 'cypress-cucumber-preprocessor/steps'
 
 Then('contains the user name', () => {
-    cy.get('@meFixture').then(({ name }) => {
-        cy.get('[data-test="headerbar-profile-username"]').then($name => {
+    cy.getAll('@meFixture', '[data-test="headerbar-profile-username"]').should(
```
Mohammer5, line 5: There's a `cy.all` helper that could be used as well:

```js
cy.all(
  () => cy.get('@meFixture'),
  () => cy.get('[data-test="headerbar-profile-username"]')
).should(([{ name }, $name]) => expect($name.text()).to.equal(name)
```

To me this is easier to parse visually

cypress/integration/Layer/stacking/index.js
```diff
@@ -34,19 +34,35 @@ Given(
 )
 
 Then('the second layer is on top of the first layer', () => {
-    assertLayerIsOnTop('second')
+    cy.get('body').click()
+    cy.window().should(win => {
+        expect(win.onLayerClick).to.be.calledOnce
+        expect(win.onLayerClick).to.be.calledWith('second')
+    })
```
Mohammer5, line 41: This wants to be refactored hehe

```js
cy.window()
  .its('onLayerClick')
  .should('be.calledOnce')
  .and('be.calledWith', 'second')
```

cypress/integration/MultiSelect/position/index.js
```diff
@@ -22,6 +22,21 @@ Given(
...
+Given('the input is empty', () => {
+    cy.get('[data-test="dhis2-uicore-select-input"]')
+        .should('exist')
+        .should('have.length', 1)
+        .then(inputs => {
+            const $input = inputs[0]
+            const inputRect = $input.getBoundingClientRect()
+            const height = inputRect.height
+
+            cy.wrap(height).as('emptyInputHeight')
```
Mohammer5, line 34: you could just return the height here and then chain the `.as` command on the outer chain

cypress/integration/MenuItem/position/index.js
```diff
@@ -28,7 +28,10 @@ When('the user hovers over the MenuItem', () => {
 Then(
     'the right of the MenuItem is aligned with the left of the SubMenu',
     () => {
-        getMenuItemAndSubMenuRects().then(([menuItemRect, subMenuRect]) => {
+        cy.getPositionsBySelectors(
+            '[data-test="dhis2-uicore-menuitem"]',
+            '[data-test="submenu"]'
+        ).should(([menuItemRect, subMenuRect]) => {
```
HendrikThePendric, line 34: I like the approach you've taken here. When you reworked the Select position tests, you chose a slightly different approach. Could you align the Select position tests https://github.com/dhis2/ui/pull/66/commits/f355ea6889e7217f91bdbf7cfd932be2bc0063a8 with the current approach?

cypress/integration/Popover/position/index.js
```diff
@@ -48,7 +48,7 @@ Given(
 Given(
     'there is sufficient space between the bottom of the reference element and the bottom of the Popover to show the arrow',
     () => {
-        getRefAndPopoverPositions().then(([refPos, popoverPos]) => {
+        getRefAndPopoverPositions().should(([refPos, popoverPos]) => {
```
HendrikThePendric, line 51: This relates to my previous comment. Here we are now using the `getRefAndPopoverPositions` helper, which just calls `cy.getPositionsBySelectors`. So now we have three flavours of doing these position assertions:
- Using a named helper that calls `cy.getPositionsBySelectors`
- Using `cy.getPositionsBySelectors` directly (I think this is probably the way to go)
- Not using a helper, as in https://github.com/dhis2/ui/commit/f355ea6889e7217f91bdbf7cfd932be2bc0063a8

So it seems we are introducing inconsistency here. It's not a big deal, but I think my personal preference would be that we either:
1. Consistently call `cy.getPositionsBySelectors` directly in this PR, and update the position assertions for the Select too.
2. Focus purely on refactoring `.then()` to `.should()`, and don't touch the position assertions themselves. This would mean reverting some of the changes in this PR too (i.e.`cypress/integration/MenuItem/position/index.js`).

cypress/integration/HeaderBar/the_headerbar_contains_a_profile_menu/the_user_name_and_email_are_displayed.js
ismay: Yeah I agree, I also think all uses of getAll could be moved to `all`

cypress/integration/MultiSelect/position/index.js
ismay: Ah yeah that's better. Fixed in 087b635

cypress/integration/MenuItem/position/index.js
ismay: I've removed this change. Created an issue to address this here: https://github.com/dhis2/ui/issues/106

cypress/integration/Popover/position/index.js
ismay: See https://github.com/dhis2/ui/pull/102#discussion_r429929966

cypress/integration/Layer/stacking/index.js
ismay: Yeah that seems nice. I think that that could be filed under point 1. here: https://github.com/dhis2/ui/issues/103#issuecomment-633571972. I'll address it when fixing that issue if you don't mind.

cypress/integration/HeaderBar/the_headerbar_contains_a_profile_menu/the_user_name_and_email_are_displayed.js
ismay: Noted in this issue: https://github.com/dhis2/ui/issues/86#issuecomment-632147319

cypress/integration/FileInput/common/index.js
ismay: Yup, agreed. Same here as in the other comment, should be addressed when fixing this one: https://github.com/dhis2/ui/issues/103#issuecomment-633571972 (see point 1., I think it would fall under there)

PR #116, opened 2020-05-28
fix(headerbar): hide app links from unprivileged users

HendrikThePendric:


amcgee:
Nice @HendrikThePendric , looks good!  Minor, but do you think we should also have a story for unauthorized users in the demo storybook, rather than only in the test storybook?  I suppose it's not that necessary, but might be nice?

HendrikThePendric:
@amcgee your suggestion made me realise that currently all the demo stories were not showing the messaging- and interpretations-app at all 🤦 . I've now corrected this, and also added a story to demo in which the interpretations-app is hidden because the user only has authority for the messaging-app.

So thanks for suggesting that additional story 😉

PR #129, opened 2020-06-02
test(cypress): update cypress commands

ismay:
I should have updated these before merging https://github.com/dhis2/ui/pull/114.

Closes #123


ismay:
> We'll have to remember to change the port for the demo stories back to 5000. Could you create an issue for that?

@Mohammer5 Actually, it turned out to be more straightforward to just have the separate port addition in this PR as well, things weren't working as I expected them to. Let me know what you think of these changes.

PR #130, opened 2020-06-02
docs(rich-text-parser): add draft feature files

HendrikThePendric:
This is equivalent to this https://github.com/dhis2/ui-widgets/pull/289 one, which I closed because we are archiving `@dhis2/ui-core`.

### DIFF
--- a/cypress/drafts/RichTextParser/Emoticons.feature
+++ b/cypress/drafts/RichTextParser/Emoticons.feature
@@ -1,0 +1,15 @@
+Feature: Emoticon parsing
+
+  Scenario Outline: Emoticons strings
+      Given there is a text string of <string>
+      And the string is not preceded immediately by a non-space character
+      Then the parsed text string will be rendered as <result>
+
+      Examples:
+          | string | result                 |
+          | :)     | <span>\u{1F642}</span> |
+          | :-)    | <span>\u{1F642}</span> |
+          | :(     | <span>\u{2639}</span>  |
+          | :-(    | <span>\u{2639}</span>  |
+          | :+1    | <span>\u{1F44D}</span> |
+          | :-1    | <span>\u{1F44E}</span> |
--- a/cypress/drafts/RichTextParser/Headers.feature
+++ b/cypress/drafts/RichTextParser/Headers.feature
@@ -1,0 +1,17 @@
+Feature: Header items parsing
+
+  Scenario Outline: Headers
+    Given there is a text string of <string>
+    Then the parsed text will be rendered as <result>
+
+    Examples:
+      | string        | result             |
+      | # sky         | <h1>sky</h1>       |
+      | ## sky        | <h2>sky</h2>       |
+      | ### sky       | <h3>sky</h3>       |
+      | #### sky      | <h4>sky</h4>       |
+      | ##### sky     | <h5>sky</h5>       |
+      | ###### sky    | <h6>sky</h6>       |
+      | the # sky     | the # sky          |
+      | #sky          | #sky               |
+      | ### the # sky | <h3>the # sky</h3> |
--- a/cypress/drafts/RichTextParser/Lists.feature
+++ b/cypress/drafts/RichTextParser/Lists.feature
@@ -1,0 +1,73 @@
+Feature: List item parsing
+
+  Scenario: Valid unordered list
+    Given there is a string with content
+      """
+      - one
+      - two
+      - three
+      """
+    Then the parsed text will be rendered as "<ul><li>one</li><li>two</li><li>three</li></ul>"
+
+  Scenario: Invalid unordered list
+    Given there is a string with content
+      """
+      -one
+      -two
+      -three
+      """
+    Then the parsed text will be rendered as "<p>-one -two -three</p>"
+
+  Scenario: Invalid list item inside unordered list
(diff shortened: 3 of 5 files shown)


HendrikThePendric:
Closing this one in favour of https://github.com/dhis2/ui-widgets/pull/289

HendrikThePendric:
Sorry, I closed the wrong PR 🤦 . I've now reopened this one and closed https://github.com/dhis2/ui-widgets/pull/289

cypress/drafts/RichTextParser/Emoticons.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Emoticon parsing
+
+  Scenario Outline: Emoticons strings
+      Given there is a text string of <string>
+      And the string is not preceded immediately by a non-space character
```
Mohammer5, line 5: I don't like this rule. What if I want to have an emoticon at the end of a sentence, or at the end of brackets.

HendrikThePendric:
I've removed myself from the list of assignees and have added @cooper-joe. This seemed more appropriate since he wrote the initial version of this PR. I only moved it to the mono-repo.

cypress/drafts/RichTextParser/Emoticons.feature
cooper-joe: I think the common workaround is to add a space to trigger the emoticon replacement and then backspace and add the extra bracket, comma, etc. Although slightly irritating, I think this is preferable. Interpreting `:)` anywhere as an emoticon would, I think, result in more unintended consequences.

cooper-joe:
> I can see why we would want feature files here, on the other hand, these look like perfect jest candidates, there's no rendering involved, it's pure input->output, while feature files should cover user interaction.
> 
> So I don't think we'd need feature files for this, but rather for something like "given the user highlighted some text. When the user clicks the bold button, then selected text should be bold and not highlighted anymore"

@Mohammer5, I have no opinion about feature files vs. jest, however it would be good to get this PR merged or closed. What do you think would be the best way to move forward here?

Mohammer5:
I think here we have to differentiate between behavior (cypress) and parsing (jest). I think the feature files of this PR are about parsing only, so we shouldn't use cypress.

It's nice to have the feature files though, so maybe we should consider jest-cucumber for the `ui` repository (cc @ismay @amcgee @varl @HendrikThePendric) as these specs are written by @cooper-joe and it's very nice/comfy to just get a list of feature files instead of having to parse a jira issue or google docs file.

ismay:
> I think here we have to differentiate between behavior (cypress) and parsing (jest). I think the feature files of this PR are about parsing only, so we shouldn't use cypress.
> 
> It's nice to have the feature files though, so maybe we should consider jest-cucumber for the `ui` repository (cc @ismay @amcgee @varl @HendrikThePendric) as these specs are written by @cooper-joe and it's very nice/comfy to just get a list of feature files instead of having to parse a jira issue or google docs file.

Personally I would just use jest without cucumber. I think the indirection and added maintenance that jest-cucumber adds isn't worth the improved readability of the specs for me. In my opinion when written properly, jest tests by themselves are already quite readable. Enough that I'd say they don't need to be improved further.

HendrikThePendric:
This presents us with a bit of a dilemma:
- I find that these feature files hold valuable information so they should not be discarded
- On the other had I also agree that these type of tests are unit tests and probably don't need feature files

So ultimately what should happen is that a set of unit tests is written that covers everything that's described in these feature files. But the most sensible time to write these is when we get to work on the component, which I don't think will be very soon....

So perhaps for now we do this:
- Add plentiful comments to these feature files that explain that these feature files should probably be converted to jest test when working on the component (and possibly look into using `jest-cucumber`, I'm not a huge fan).
- Just merge these feature files to the drafts folder

Mohammer5:
@cooper-joe would you be able to write test file skeletons? I mean not the actual test code itself, but rather like this:

```js
describe('Transfer - Highlight range of option', () => {
  it('should select the option after shift clicking the option when not selected', () => {})
  it('should deselect the option after shift clicking the option when selected', () => {})
  // ...
})
```

cooper-joe:
> @cooper-joe would you be able to write test file skeletons and probably understand test files when reading through them when they have actual test code? I mean not the actual test code itself, but rather like this:
> 
> ```js
> describe('Transfer - Highlight range of option', () => {
>   it('should select the option after shift clicking the option when not selected', () => {})
>   it('should deselect the option after shift clicking the option when selected', () => {})
>   // ...
> })
> ```

I'm sure I could learn to do so! Although I do think that the behaviour is easier to define when writing feature files than the above example. If it would simplify this process then I am open to giving it a try.

Mohammer5:
My preference would be to use feature files, but I'm outnumbered :wink: We could schedule a meeting and I can give you an introduction to jest test files

ismay:
> My preference would be to use feature files, but I'm outnumbered

I know it's in jest (💯 pun credits for me), but I wouldn't want to force a decision just through numbers. I would prefer to go with the best solution, even if the majority disagrees. I wouldn't mind taking a deeper look and weighing the pros/cons, there could be things I'm overlooking.

HendrikThePendric:
Closing this due to inactivity

PR #133, opened 2020-06-03
chore(forms): update snapshot to match new type

ismay:
Fixes a jest test failure.

### DIFF
--- a/packages/forms/src/__tests__/__snapshots__/index.test.js.snap
+++ b/packages/forms/src/__tests__/__snapshots__/index.test.js.snap
@@ -45,7 +45,7 @@ Array [
 ]
 `;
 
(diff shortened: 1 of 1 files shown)


ismay:
> I'm pretty sure I addressed this problem in a dependabot PR yesterday already...
> See #126 commit [4bd3ce1](https://github.com/dhis2/ui/commit/4bd3ce1e57a57f67ad333a88370cc9d4c312220f)
> 
> In master this seems to be correct already:
> 
> https://github.com/dhis2/ui/blob/1704fbd17b75c0ddbb326a9361eaf85992e6a6c3/packages/forms/src/__tests__/__snapshots__/index.test.js.snap#L48
> 
> Shall we close this one?

@HendrikThePendric Ah, I see what happened now, my dependencies weren't up to date, so that's why the test was failing. After updating the test passes. Yeah we can close this 👍

Issue #106, opened 2020-05-25
Use uniform approach for position testing in cypress


ismay:
As @HendrikThePendric notes here: https://github.com/dhis2/ui/pull/102#discussion_r429754967, we're now using different ways to test position. We should unify that so that we use a single approach.

HendrikThePendric:
@ismay I have forgotten what our preferred solution would look like. 

Shall we go for the bare metal approach? Like this one👇 
https://github.com/dhis2/ui/blob/1704fbd17b75c0ddbb326a9361eaf85992e6a6c3/cypress/integration/SingleSelect/position/index.js#L32-L48

Or shall we go for the approach [you took here](https://github.com/dhis2/ui/pull/102#pullrequestreview-417485432)?
```javascript
cy.getPositionsBySelectors(
    '[data-test="dhis2-uicore-menuitem"]',
    '[data-test="submenu"]'
).should(([menuItemRect, subMenuRect]) => {
```

I have a slight preference for the second approach but I'm also fine with the first.

Mohammer5:
I think it should be fine to have a mixed approach. We already have the `all` selector, combined, it could look something like this:


```js
cy.all(
  () => cy.get('[data-test="dhis2-uicore-menuitem"]').getPosition(),
  () => cy.get('[data-test="submenu"]').getPosition(),
).should(([menuItemRect, subMenuRect]) => { /* ... */ })
```

The `getPosition` command could stay but works on a `prevSubject` while using the `all` command to do it multiple times "at once".

HendrikThePendric:
I really like @Mohammer5's suggestion. Probably I would go with something like `getRect` instead of `getPosition`. I came up with the name `getPositionsBySelectors`, and the fact that this returns `rects`, and each `rect` doesn't just contain position data (i.e. top, right, bottom, left) but also width and height has bothered me ever since.

ismay:
I'd personally prefer going with the bare metal approach. I've not seen _that_ much position testing in UI, so for me the added indirection and maintenance don't seem worth the convenience of abstraction.

HendrikThePendric:
> I've not seen that much position testing in UI

There are 34 scenarios that assert a position. I think that's quite a lot, right?

ismay:
> > I've not seen that much position testing in UI
> 
> There are 34 scenarios that assert a position. I think that's quite a lot, right?

[...]

Issue #83, opened 2020-05-19
Add e2e tests for new popper & popover features


HendrikThePendric:
We've added a few things to the Popper and Popover which should be tested:
- Accepting different types of variables for the `reference` prop, i.e. React.ref, DOM node, and virtual element
- Arrow directions for the popover. We should test all 4 main directions (top/bottom/left/right) + complex placement (i.e. bottom-start) + verify that the arrow shifts a little bit when the reference element is narrower than the popper element.

PR #141, opened 2020-06-04
test(popper): additional popper and popover e2e tests

HendrikThePendric:
This adds some tests to check the arrow positions for the `Popover` and different reference types for the `Popper`. When merged, issue https://github.com/dhis2/ui/issues/83 can be closed.

Along the way I also fixed a few typos and a bug (see commit 209e5be and this issue popperjs/react-popper#354)

### DIFF
--- a/cypress/integration/Popover/arrow_position/index.js
+++ b/cypress/integration/Popover/arrow_position/index.js
@@ -1,0 +1,121 @@
+import { Given, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given('a Popover is rendered with placement top', () => {
+    cy.visitStory('Popover', 'Placement Top')
+})
+Given('a Popover is rendered with placement right', () => {
+    cy.visitStory('Popover', 'Placement Right')
+})
+Given('a Popover is rendered with placement bottom', () => {
+    cy.visitStory('Popover', 'Placement Bottom')
+})
+Given('a Popover is rendered with placement left', () => {
+    cy.visitStory('Popover', 'Placement Left')
+})
+Given('a Popover with position left is shifted into view', () => {
+    cy.visitStory('Popover', 'Shifted Arrow')
+})
+
+Then('the Arrow is horizontally aligned with the Popper', () => {
+    cy.all(
+        () => cy.get('[data-test="dhis2-uicore-popover"]'),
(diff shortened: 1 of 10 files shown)


HendrikThePendric:
BTW: Can I just keep these commit as the are? Or should I squash them before merging?

cypress/integration/Popover/arrow_position/index.js
```diff
@@ -0,0 +1,101 @@
...
+Given('a Popover is rendered with placement right', () => {
+    cy.visitStory('Popover', 'Placement Right')
+})
+Given('a Popover is rendered with placement bottom', () => {
+    cy.visitStory('Popover', 'Placement Bottom')
+})
+Given('a Popover is rendered with placement left', () => {
+    cy.visitStory('Popover', 'Placement Left')
+})
+Given('a Popover with position left is shited into view', () => {
```
ismay, line 15: shifted

cypress/integration/Popover/arrow_position/index.js
```diff
@@ -0,0 +1,101 @@
...
+    cy.visitStory('Popover', 'Shifted Arrow')
+})
+
+Then('the Arrow is horizontally aligned with the Popper', () => {
+    cy.getPositionsBySelectors(
```
ismay, line 20: Maybe we should resolve this first: https://github.com/dhis2/ui/issues/106

ismay:
> BTW: Can I just keep these commit as the are? Or should I squash them before merging?

I would use the `test` prefix for all changes related to the tests. The `refactor` commit type shows up in the changelog for instance.

I you ask me, I'd say squash to a commit that changes the popper test, a commit that does the popover tests, and the one with the sharedPropType fix. Unless they're all somehow related.

ismay:
Looks good to me. I personally think it'd be nice to resolve #106 if we can, before merging this. That way we won't be introducing more tests that need to be refactored later.  Other than that and the typo this looks good to merge to me.

cypress/integration/Popover/arrow_position/index.js
Mohammer5: :laughing:

HendrikThePendric:
I've squashed all my commits, fixed the typo and have implemented the position assertions as discussed in #106 (and a little bit on Slack). So I think this is now ready for a re-review.

PR #161, opened 2020-06-11
docs(getting-started): fix badge and polyfills section

ismay:
If we want to add a polyfills section we can always do that later. I don't see any added value in having the TODO there, so I removed it.

The badge was pointing to ui-core, so I fixed that as well.


ismay:
> 👍
> Todo stuff should probably be issues instead of existing inside published documents.

I agree 👍

Issue #119, opened 2020-05-28
Documentation tasks following the v5 release


ismay:
### Done

- [x] [Docs landing page](https://ui.dhis2.nu/) has `root` as the title in several places, I'd expect UI, or something of that nature.
- [x] The [getting started page](https://ui.dhis2.nu/#/getting-started) lists ui-core installation instructions, not ui (done, see #158)
- [x] The getting started page has a polyfills section, under which it just says TODO (done, see https://github.com/dhis2/ui/pull/161)
- [x] Storybook does not show syntax (done, fixed in https://github.com/dhis2/ui/pull/162)
- [x] Post on COP once changelog and blogpost have been finalized (done, see here: https://community.dhis2.org/t/release-post-for-dhis2-ui-version-5/39475)
- [x] The [concepts section](https://ui.dhis2.nu/#/concepts) mentions under construction and alpha version, we should check this and rewrite outdated areas to reflect the current situation (done, see: https://github.com/dhis2/ui/pull/166)
- [x] Publish an article about the v5 release, explaining the move and also explaining the changes as we did in for the [v4 release](https://dhis2.nu/2019/12/ui-core-4/). (done, see here: https://developers.dhis2.org/2020/05/ui-5)

### Ready for review

- [x] [Storybook](https://ui.dhis2.nu/demo): widgets is the only uncapitalized folder in the tree (I know, nitpick), plus the nesting seems off (PR: https://github.com/dhis2/ui/pull/167)
- [x] Add styled-jsx gotcha to UI docs instead of notes (PR here: https://github.com/dhis2/ui/pull/168)
- [x] update the links on developers.dhis2.org (PR: https://github.com/dhis2/dhis2.github.io/pull/34)
- [x] Update the _Component & Helper Libraries_ section on the [Libraries page](https://dhis2.nu/libs/) (move outdated libs to Legacy libraries section below) (assuming this is the same issue as the above, PR: dhis2/dhis2.github.io#34)
- [x] Redirect ui-core.dhis2.nu and ui-widgets.dhis2.nu to ui

### Todo

- [ ] Add a link to ui in the docsite navbar
- [ ] Link to release/migration post from the docs, maybe have a Migration Guide docs page (or something) that links to the blogpost for each major release?
[...]

PR #168, opened 2020-06-15
docs(styled-jsx): add documentation on styled-jsx override

ismay:
Ref: #119

This adds the styled-jsx override example from https://github.com/dhis2/notes/pull/110 here. It's advanced usage, and discouraged, but we've received enough question about it that we should document it clearly somewhere, so that the dangers are also clear to everyone.

Closes: https://github.com/dhis2/notes/pull/110, https://github.com/dhis2/notes/issues/108

Note: I think that Viktor wanted to add this override somewhere else. I think I saw a PR or issue where he added this. So consider this a proposal, we should take a look at if this matches what Viktor wanted to do as well.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -1,0 +1,35 @@
+# Overriding the styled-jsx from our libraries
+
+### Recommendations
+
+Usually there are better ways to achieve what you want than overriding the styled-jsx we use to style our library components. Consider the following alternatives:
+
+1. If the component has a bug in the css, open an issue or PR here https://github.com/dhis2/ui.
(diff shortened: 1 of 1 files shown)


ismay:
> I think this should include the solution to deal with the specificity as well. The example in the current state uses one class name which won't work 100%.
> 
> I think `ui` is the proper place to put these docs, but then we should close [dhis2/notes#110](https://github.com/dhis2/notes/pull/110)

Yeah that's why I added the !important (and the explanation/disclaimer). I think it's a cleaner solution than artificially bumping the specificity by including extra selectors.

ismay:
I'll go ahead and merge this, and if it turns out this does not align with what Viktor prefers we can always revert it.

Issue #176, opened 2020-06-17
hideDivider prop of SectionHeader is inverted


amcgee:
The divider is shown iff `hideDivider` is specified, when it should be the opposite.

![Screen Shot 2020-06-17 at 5 06 39 PM](https://user-images.githubusercontent.com/947888/84915720-723b0700-b0bd-11ea-9c46-2ba286b80654.png)
![Screen Shot 2020-06-17 at 5 06 25 PM](https://user-images.githubusercontent.com/947888/84915723-736c3400-b0bd-11ea-95f6-28bbf7eb440e.png)

HendrikThePendric:
I expect something else is going on @amcgee 
1. The logic in the `MenuSectionHeader` is correct, see:
https://github.com/dhis2/ui/blob/35b2964e13371998734f80e48ee8d0a3478f2ef9/packages/core/src/MenuSectionHeader/MenuSectionHeader.js#L26

2. But at the last moment we decided to add some magic to the `Menu`
https://github.com/dhis2/ui/blob/35b2964e13371998734f80e48ee8d0a3478f2ef9/packages/core/src/Menu/Menu.js#L23-L27

So the situation currently is as follows:
- By default, if the first `li` in a menu `ul` is a menu-section-header, the divider will be hidden, because it doesn't look good [*] to start the menu with a divider. 
- However, the current implementation does allow for that behaviour to be overridden by explicitly setting `hideDivider` to `false`.
[*] I think @ismay, @Mohammer5 and myself came to this conclusion during a meeting or a Slack conversation, because I can't seem to find a mention of this in the PR conversation here https://github.com/dhis2/ui/pull/44...

Given the confusion this "feature" has already caused, I wonder if we would perhaps want to get rid of it. But in any case, the above should at least explain what is causing the behaviour you're seeing.

ismay:
It's not really clear what the intention of that bit of logic in the menu is. Is it:

* If hideDivider is not set, and it's the first item in the list, hide the divider
* If hideDivider is not set, and it's not the first item in the list, fall back to the dense prop

That second bit is confusing to me. Why is it falling back to the setting of the dense prop?

Mohammer5:
The `dense` part doesn't make much sense to me either right now.
I think it should be

```js
hideDivider: index === 0
  // if it's the first item, hide the divider if hideDivider is not specified
  ? typeof props.hideDivider !== 'boolean' || props.hideDivider
  // if it's not the first item, show the divider if hideDivider is not specified
[...]

PR #177, opened 2020-06-18
fix(menu): stop using child's dense prop to set value of hideDivider

HendrikThePendric:
This relates to #176. We are dealing with two things here:
- A potentially confusing feature, hiding the divider from the menu-section-header if it's the first child of a menu. In most cases I think this behaviour does make (some) sense and it's possible to "opt out" of this behaviour by explicitly setting `hideDivider` to `false`.
- A bug in the implementation of this feature.

Currently all I've done in this PR is to fix that bug, and introduce an example of the feature.

However, I'd also be fine removing the feature. It seems to be causing more confusion than it brings convenience.

Let me know what you'd prefer.

### DIFF
--- a/packages/core/src/Menu/Menu.stories.js
+++ b/packages/core/src/Menu/Menu.stories.js
@@ -2,6 +2,7 @@ import React from 'react'
 
 import { MenuItem } from '../MenuItem/MenuItem.js'
 import { Menu } from './Menu.js'
+import { MenuSectionHeader } from '../MenuSectionHeader/MenuSectionHeader.js'
 
 export default {
     title: 'Menu',
@@ -19,6 +20,13 @@ export const Dense = () => (
     <Menu dense>
         <MenuItem label="Menu item" />
         <MenuItem label="Menu item" />
+    </Menu>
+)
+
+export const AutoHideFirstSectionHeaderDivider = () => (
+    <Menu>
+        <MenuSectionHeader label="First - no divider above" />
+        <MenuSectionHeader label="Second - with divider above" />
     </Menu>
 )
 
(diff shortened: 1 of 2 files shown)


packages/core/src/Menu/Menu.stories.js
```diff
@@ -22,6 +23,13 @@ export const Dense = () => (
     </Menu>
 )
 
+export const AutoHideFirstSectionHeaderDivider = () => (
+    <Menu>
+        <MenuSectionHeader label="First - no divider above" />
+        <MenuSectionHeader label="Second - with divider above" />
+    </Menu>
+)
+
```
Mohammer5, line 32: I think this should go into the `MenuSectionHeader.stories.js`

HendrikThePendric: Yeah I started of having it there. But in the end I moved it, because this is actually a feature of the Menu.

Having said that, I didn't realise that the JSX panel was showing the decorator markup too. I suppose moving it there does make sense after all. I'll do that.

HendrikThePendric: Correction. I tried that and because we're using a decorator in that story already it's only possible to render two headers in a fragment, and that breaks the logic, since the menu is expecting its children to be the items. So to change this, I'd have to remove the decorator from the MenuSectionHeader demo and update all the stories there.

It's not that the above would be much work, but to me it just underpins that this is actually a feature of the Menu, and should be demoed as such.

Mohammer5: Hmm to me, all these different story "namespaces" are quite confusing. I think I'd prefer if everything could be found under `Menu`

Mohammer5 (approved):

PR #188, opened 2020-06-24
test(box): add test for multiple props

ismay:
I noticed that the box isn't applying multiple props. Single props (just marginTop, or maxWidth) are applied correctly, but if you pass multiple props they aren't applied at all. This adds a (currently failing) test that demonstrates the problem.

### DIFF
--- a/cypress/integration/Box/accepts_multiple_props.feature
+++ b/cypress/integration/Box/accepts_multiple_props.feature
@@ -1,0 +1,5 @@
+Feature: The Box accepts multiple props
+
+    Scenario: A Box with multiple props
+        Given a Box with multiple props is rendered
+        Then the styles are all applied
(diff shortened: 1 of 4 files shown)


varl:
@ismay there is an "oopsie" on this line: https://github.com/dhis2/ui/blob/master/packages/core/src/Box/Box.js#L27

Notice the missing semicolon in the CSS statement, this would cause none of the other CSS rules to be applied.

ismay:
> @ismay there is an "oopsie" on this line: https://github.com/dhis2/ui/blob/master/packages/core/src/Box/Box.js#L27
> 
> Notice the missing semicolon in the CSS statement, this would cause none of the other CSS rules to be applied.

Ah I missed that one, thanks! Should be fixed now.

PR #190, opened 2020-06-30
docs(readme): update packages links

erikarenhill:


amcgee:
Looks like there's a lint issue?

erikarenhill:
> Looks like there's a lint issue?

yeah, really exciting as I only removed "ui-" in some links... I'll check what it might be

Maybe some new rules introduced after the first readme was pushed.

PR #196, opened 2020-07-02
fix(switch): use input value instead of checked for the value

ismay:
My earlier fix unearthed a different issue, namely that we're passing `input.checked` to the Switch's `value` prop. According to the MDN docs on the [checkbox input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox), the input property is just for specifying a string that will be submitted to the backend if the checkbox is checked. Final-form seems to comply with this as well:

Screenshot 2020-07-02 at 13 54 17

> https://final-form.org/docs/react-final-form/types/FieldProps

 So instead of `checked` we should be passing `value` here. I see we already do this with the RadioFieldFF and CheckboxFieldFF, so I think this was just a typo.


ismay:
> 👏

Senks 😊

PR #200, opened 2020-07-03
feat(tech 378): add scrollable area

Mohammer5:
This will add a component that notifies the consumer about what scroll position is visible.
Currently there are 3 areas: `top`, `middle` & `end` (100% open to renaming these)

* [x] Add component
* [x] Add stories
* [ ] Write cypress tests
* [ ] Discuss edge cases
   * [ ] What if both the `top` and `end` screen are visible right from the start
     *See below*
   * [ ] What if the `top` and `end` screen overlap
     *See below*

Edge cases:
Currently, if the `end` part is visible, that part will be the initial area. If that's not the case but the top part is visible (which should always be the case), the `top` will be the initial area


Mohammer5:
Closed in favor of #204

PR #204, opened 2020-07-07
feat(intersection detector): add component (TECH-378)

Mohammer5:
Relates to [TECH-378](https://jira.dhis2.org/browse/TECH-378)

This will add a component that can be used to track whether it's inside the viewport.

- [x] Add component
- [x] Add stories
- [x] Add cypress tests

### DIFF
--- a/packages/core/src/IntersectionDetector/IntersectionDetector.js
+++ b/packages/core/src/IntersectionDetector/IntersectionDetector.js
@@ -1,0 +1,116 @@
+import React, { useEffect, useRef } from 'react'
+import { PropTypes } from '@dhis2/prop-types'
+;('') // TODO: https://github.com/jsdoc/jsdoc/issues/1718
+
+/**
+ * @module
+ *
+ * @param {IntersectionDetector.PropTypes} props
+ * @returns {React.Component}
+ */
+export const IntersectionDetector = ({
+    threshold,
+    onChange,
+    children,
+    className,
+    rootRef,
+}) => {
+    // Use useRef instead of useState to prevent unnecessary re-render:
+    //   The state changes won't be reflected in what this component renders,
+    //   so there's no need for re-rendering the (potentially computational
+    //   heavy) children.  Also: If the parent re-renders (e. g. due to a state
+    //   change), then this component will re-render as well.
+
+    // @var {Object}
+    // @prop {IntersectionObserver} current
+    const observer = useRef()
+
+    // @var {Object}
+    // @prop {bool} current
+    const isIntersecting = useRef()
+
+    // @var {Object}
+    // @prop {HTMLElement} current
+    const intersectionRef = useRef()
+
+    useEffect(() => {
+        const rootEl = rootRef.current
+        const intersectionEl = intersectionRef.current
+
+        if (rootEl && intersectionEl && !observer.current) {
+            const onIntersection = entries => {
+                // Currently there's no way to supply multiple thresholds,
+                // so a single entry can be assumed safely
+                const [entry] = entries
+
+                // Make sure the callback is not called multiple times
+                // if there is no change
+                const intersectionChange =
+                    entry.isIntersecting !== isIntersecting.current
+
+                if (intersectionChange) {
+                    isIntersecting.current = entry.isIntersecting
+                    onChange({
+                        isIntersecting: entry.isIntersecting,
+                    })
+                }
+            }
+
+            const observerOptions = { root: rootEl, threshold }
+            const intersectionObserver = new IntersectionObserver(
+                onIntersection,
+                observerOptions
+            )
+
+            intersectionObserver.observe(intersectionEl)
+            observer.current = intersectionObserver
+
+            // Make sure to clean up everything when un-mounting.
+            // Using an arrow function instead of just returning
+            // the disconnect function for better readability.
+            return () => intersectionObserver.disconnect()
+        }
+    }, [rootRef.current, intersectionRef.current])
+
+    return (
+        <div ref={intersectionRef} className={className}>
+            {children}
+
+            <style jsx>{`
+                // Enables compatibility with "position: absolute;" containers.
+                // If the parent does not have a height, this will have no
+                // effect, which makes it work with standard containers.
+                div {
+                    height: 100%;
+                }
+            `}</style>
+        </div>
+    )
+}
+
+IntersectionDetector.defaultProps = {
+    threshold: 0,
+}
+
+/**
+ * @typedef {Object} PropTypes
+ * @static
+ *
+ * @prop {Object} rootRef
+ * @prop {HTMLElement} [rootRef.current]
+ * @prop {Function} onChange
+ * @prop {any} [children]
+ * @prop {string} [className]
+ * @prop {number} [threshold]
+ */
+IntersectionDetector.propTypes = {
+    rootRef: PropTypes.shape({
+        // not required so `current` can be `null`
+        current: PropTypes.instanceOf(HTMLElement),
+    }).isRequired,
+    onChange: PropTypes.func.isRequired,
+
+    children: PropTypes.any,
+    className: PropTypes.string,
+    threshold: PropTypes.number,
+}
--- a/packages/core/src/IntersectionDetector/IntersectionDetector.stories.e2e.js
+++ b/packages/core/src/IntersectionDetector/IntersectionDetector.stories.e2e.js
@@ -1,0 +1,61 @@
+import React, { useRef } from 'react'
+import { IntersectionDetector } from './IntersectionDetector.js'
+
+window.onChange = window.Cypress ? window.Cypress.cy.stub() : () => null
+
+export default { title: 'IntersectionDetector' }
+
+export const OutOfView = () => {
+    const rootRef = useRef()
+
+    return (
+        <div
+            ref={rootRef}
+            data-test="scroll-container"
+            style={{ width: 200, height: 300, overflow: 'auto' }}
+        >
+            <div
+                // spacer to push indicator out of the intersecting area
+                style={{
+                    width: 200,
+                    height: 310,
+                    overflow: 'auto',
+                    background: 'blue',
+                }}
+            />
+
+            <IntersectionDetector rootRef={rootRef} onChange={window.onChange}>
+                <div style={{ background: 'red', height: 200 }} />
+            </IntersectionDetector>
+        </div>
+    )
+}
+
+export const InView = () => {
+    const rootRef = useRef()
+
+    return (
+        <div
+            ref={rootRef}
+            data-test="scroll-container"
+            style={{ width: 200, height: 300, overflow: 'auto' }}
+        >
+            <div style={{ background: 'red', height: 200 }}>
+                <IntersectionDetector
+                    rootRef={rootRef}
+                    onChange={window.onChange}
+                />
+            </div>
+
+            <div
+                // spacer to push indicator out of the intersecting area
+                style={{
+                    width: 200,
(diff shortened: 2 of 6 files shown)


cypress/integration/IntersectionDetector/Visibility_notification/index.js
```diff
@@ -0,0 +1,43 @@
...
+    cy.get('[data-test="scroll-container"]').scrollTo(0, 201)
+})
+
+Then('the callback passed to onIsVisible should be called', () => {
+    cy.get('@onVisibilityChangeStub').should(stub => {
+        expect(stub).to.be.calledTwice
+    })
+})
+
+Then(/the visible prop of its payload should be .*/, bool => {
```
ismay, line 40: Can this be done without a regex? I don't find them the most readable personally.

packages/core/src/IntersectionDetector/IntersectionDetector.js
```diff
@@ -0,0 +1,105 @@
...
+import { PropTypes } from '@dhis2/prop-types'
+;('') // TODO: https://github.com/jsdoc/jsdoc/issues/1718
+
+/**
+ * @module
+ *
+ * @param {IntersectionDetector.PropTypes} props
+ * @returns {React.Component}
+ */
+export const IntersectionDetector = ({
```
ismay, line 11: Personally I think I'd go with `VisibilityDetector` here, as most of the terminology used in this component mentions visibility. `IntersectionDetector` could be a useful name as well, but if we're going to go with `Intersection` over `Visibility` as the core responsibility of this component, I'd move all the other terminology in this component over to `Intersection` as well. Though in that case, I'd prefer just going with `IntersectionObserver`.

packages/core/src/IntersectionDetector/IntersectionDetector.js
```diff
@@ -0,0 +1,105 @@
...
+    threshold,
+    onVisibilityChange,
+    children,
+    className,
+    rootRef,
+}) => {
+    // use ref to prevent unnecessary re-render
```
ismay, line 18: I think it'd be useful to elaborate slightly here with what you told me on slack (why not state, etc.), so that other devs understand the why.

packages/core/src/IntersectionDetector/IntersectionDetector.js
```diff
@@ -0,0 +1,105 @@
...
+                    entry.isIntersecting !== isIntersecting.current
+
+                if (intersectionChange) {
+                    isIntersecting.current = entry.isIntersecting
+                    onVisibilityChange({ visible: entry.isIntersecting })
+                }
+            }
+
+            const observerOptions = { root: rootEl, threshold }
+            const intersectionObserver = new IntersectionObserver(
```
HendrikThePendric, line 54: For `ResizeObserver` we decided to add a polyfill. Perhaps we should do the same here. Support seems similar (to be honest, support seems pretty decent for both these `*Observer` APIs, so I'm not 100% sure why we decided to add the polyfill for the `ResizeObserver` in the first place... Quite possibly because the specifications for the API are still in draft status 🤷 )

cypress/integration/IntersectionDetector/Visibility_notification.feature
```diff
@@ -0,0 +1,13 @@
+Feature: The IntersectionDetector notifies the consumer when becoming visible
+
+    Scenario: The user scrolls the IntersectionDetector into the view
+        Given the detector is not visible
+        When the user scrolls the detector into the view
+        Then the callback passed to onIsVisible should be called
```
HendrikThePendric, line 6: The component has a prop called `onVisibilityChange` but here in the scenarios you are talking about `onIsVisible`. I found that a little confusing.

packages/core/src/IntersectionDetector/IntersectionDetector.js
HendrikThePendric: I agree with @ismay that we should try to avoid be mixing terminology, and either refer to `visibility` or `intersection` consistently. I'm not 100% sure which one I would prefer here:
- `visibility` is nice, because I expect that in 99% of cases this is probably the type of logic you would want to implement with this component: when something becomes is in view, do A, when it moves out of view, do B. Unfortunately the word `visible` is also quite ambiguous and not entirely accurate. What happens with elements that have `visibility: hidden`, or `opacity: 0`, or are covered by another element? Will the `onVisibilityChange` callback be executed when these type of elements scroll into view? IMO it's not entirely clear...
- On the other hand, `intersection` is much more clear. It hints at the API being used under the hood, which can help manage the developer's expectations: it is 100% clear that this logic is based on intersections/overlaps.

🤔  On balance, my preference would probably be to settle on `IntersectionDetector` for the component and rename the prop from `onVisibilityChange` to `onIntersectionChange`. 

I also think this is probably more future proof: if we start adding support for more things the `IntersectionObserver` can do, this type of naming will make more sense.

cypress/integration/IntersectionDetector/Visibility_notification/index.js
Mohammer5: Split into two step definitions

packages/core/src/IntersectionDetector/IntersectionDetector.js
Mohammer5: That's very good feedback. I've decided to go for @HendrikThePendric's suggestion.

packages/core/src/IntersectionDetector/IntersectionDetector.js
```diff
@@ -0,0 +1,105 @@
...
+                onIntersection,
+                observerOptions
+            )
+
+            intersectionObserver.observe(intersectionEl)
+            observer.current = intersectionObserver
```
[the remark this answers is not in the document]
Mohammer5: Good catch, added

cypress/integration/IntersectionDetector/Visibility_notification.feature
Mohammer5: :+1:

packages/core/src/IntersectionDetector/IntersectionDetector.js
Mohammer5: It's supported in all browsers that we support, wondering why you added a polyfill.
I'd prefer not to add one as that's ultimately the responsibility of the app

Mohammer5:
@HendrikThePendric @ismay Thanks for the feedback guys, I've worked on all comments or added a response

packages/core/src/IntersectionDetector/IntersectionDetector.js
ismay: > It's supported in all browsers that we support, wondering why you added a polyfill.
I'd prefer not to add one as that's ultimately the responsibility of the app

I second this, and maybe we should remove the resizeobserver polyfill. I think we should also add https://github.com/amilajack/eslint-plugin-compat and lint browser support for the APIs we use. We should set violations to errors so that we can catch usage of unsupported APIs.

ismay:
> We should still try to avoid external dependencies if at all possible, as per @varl's recommendations. And in my personal opinion this is one of those cases where using a DYI solution based on corresponding web API makes a lot of sense. We only have 30 lines of code that deals with element-intersection and I would argue this is well below the threshold of "re-inventing the wheel".

Maybe we need to establish that rule more clearly, because I think it's not as clear to me where the cut-off lies. My intention with suggesting a third party lib is to cut down on maintenance. But I feel like we've also discussed that subject a couple times before, so I'll concede. I would like to just state for the record that I think that using a third party library here would be a more efficient use of our time.

packages/core/src/IntersectionDetector/IntersectionDetector.js
ismay: What do you think of this suggestion? I see it was resolved but am not sure what the resolution was.

packages/core/src/IntersectionDetector/IntersectionDetector.js
HendrikThePendric: > I second this, and maybe we should remove the resizeobserver polyfill.
👍

Mohammer5:
@ismay is there anything left that I need to change? If no, could you approve the PR so I can merge it?

cypress/integration/IntersectionDetector/Visibility_notification.feature
```diff
@@ -0,0 +1,13 @@
+Feature: The IntersectionDetector notifies the consumer when becoming visible
```
ismay, line 1: Maybe we should update the wording here as well, since we changed all the terms from visibility to intersection.

ismay:
> @ismay is there anything left that I need to change? If no, could you approve the PR so I can merge it?
> EDIT: I also have to fix the failing cypress test

I checked this at the end of the working day, so I hope I didn't miss anything but it looks good. Only thing I think is good to address is the wording of the tests (see comment). Let me know what you think!

cypress/integration/IntersectionDetector/Visibility_notification.feature
Mohammer5: Done

Mohammer5:
@ismay I adjusted the feature file & fixed the tests.

cypress/integration/IntersectionDetector/Visibility_notification.feature
```diff
@@ -0,0 +1,13 @@
...
+
+    Scenario: The user scrolls the IntersectionDetector into the view
```
ismay, line 3: The user scrolls the IntersectionDetector into ~~the~~ view

cypress/integration/IntersectionDetector/Visibility_notification.feature
```diff
@@ -0,0 +1,13 @@
...
+        Given the detector is not intersection with the root
```
ismay, line 4: intersection -> intersecting

cypress/integration/IntersectionDetector/Visibility_notification.feature
```diff
@@ -0,0 +1,13 @@
...
+        When the user scrolls the detector into the view
```
ismay, line 5: https://github.com/dhis2/ui/pull/204/files#r459395986

packages/core/src/IntersectionDetector/IntersectionDetector.js
```diff
@@ -0,0 +1,126 @@
...
+    threshold,
+    onChange,
+    children,
+    className,
+    rootRef,
+}) => {
+    // Use useRef instead of useState to prevent unnecessary re-render:
+    //   The state changes won't be reflected in what this component renders,
+    //   so there's no need for re-rendering the (potentially computational
+    //   heave) children.  Also: If the parent re-renders (e. g. due to a state
```
ismay, line 21: heave -> heavy

packages/core/src/IntersectionDetector/IntersectionDetector.js
```diff
@@ -0,0 +1,126 @@
...
+            !!(rootEl && intersectionEl && !observer.current)
+        )
+
+        if (rootEl && intersectionEl && !observer.current) {
+            const onIntersection = entries => {
+                // Currently there's no way to supply multiple thresholds,
+                // so a single entry can be assumed safely
+                const [entry] = entries
+
+                // Make sure the callback is not called multilpe times
```
ismay, line 54: multilpe -> multiple

Mohammer5:
@ismay typos & `console.log`s removed

Mohammer5:
@ismay Now also successfully rebased onto master. Should be good now

Mohammer5:
@ismay I've updated the commit messages

PR #211, opened 2020-07-10
feat(table): add alternateRowBgColor prop (default: true) (TECH-416)

### LINKED ISSUE #207: Table can't disable zebra striping
`Table` is lacking an option to turn off the zebra striping, shown here and in the screenshot below: https://ui.dhis2.nu/demo/?path=/story/table--static-layout.
This means that all use-cases are forced to display the zebra striping, even if this is not intended.

JIRA ticket: https://jira.dhis2.org/browse/TECH-416

image

Mohammer5:
Fixes #207 
Fixes TECH-416

Adds the `alternateRowBgColor` prop that defaults to true.
Can be set globally on the `Table` component and/or individually on `TableRow` components.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,13 @@
         "url": "https://github.com/dhis2/ui.git"
     },
     "author": "Viktor Varland <viktor@dhis2.org>",
+    "contributors": [
+        {
+            "name": "Jan-Gerke Salomon",
+            "email": "jan-gerke@dhis2.org",
+            "url": "http://functionalprogramming.ninja"
+        }
+    ],
     "license": "BSD-3-Clause",
     "private": true,
     "workspaces": {
--- a/packages/core/src/Table/Table.js
+++ b/packages/core/src/Table/Table.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import css from 'styled-jsx/css'
 import propTypes from '@dhis2/prop-types'
+import { Provider } from './TableContext.js'
 
 const tableStyles = css`
     table {
@@ -20,12 +21,19 @@ const tableStyles = css`
  * @example import { Table } from '@dhis2/ui-core'
  * @see Live demo: {@link /demo/?path=/story/table--static-layout|Storybook}
  */
-export const Table = ({ children, className, dataTest }) => (
-    <table className={className} data-test={dataTest}>
-        {children}
+export const Table = ({
+    children,
+    className,
+    dataTest,
+    suppressZebraStriping,
+}) => (
+    <Provider value={{ suppressZebraStriping }}>
+        <table className={className} data-test={dataTest}>
+            {children}
 
-        <style jsx>{tableStyles}</style>
-    </table>
+            <style jsx>{tableStyles}</style>
+        </table>
+    </Provider>
 )
 
 Table.defaultProps = {
@@ -38,9 +46,11 @@ Table.defaultProps = {
  * @prop {TableHead|TableBody|TableFoot|Array.<TableHead|TableBody|TableFoot>} [children]
  * @prop {string} [className]
  * @prop {string} [dataTest]
+ * @prop {bool} [suppressZebraStriping]
  */
 Table.propTypes = {
     children: propTypes.node,
     className: propTypes.string,
     dataTest: propTypes.string,
+    suppressZebraStriping: propTypes.bool,
 }
--- a/packages/core/src/Table/TableRow.js
+++ b/packages/core/src/Table/TableRow.js
@@ -1,9 +1,11 @@
-import React from 'react'
+import React, { useContext } from 'react'
 import css from 'styled-jsx/css'
 import propTypes from '@dhis2/prop-types'
+import cx from 'classnames'
+import { TableContext } from './TableContext'
 
 const tableRowStyles = css`
-    tr:nth-child(even) {
+    .zebraStriping:nth-child(even) {
         background: #fbfcfd;
     }
 `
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -6,6 +6,13 @@
         "url": "https://github.com/dhis2/ui.git"
     },
     "author": "Viktor Varland <viktor@dhis2.org>",
+    "contributors": [
+        {
+            "name": "Jan-Gerke Salomon",
+            "email": "jan-gerke@dhis2.org",
+            "url": "http://functionalprogramming.ninja"
+        }
+    ],
```
Mohammer5, line 15: :sunglasses:

ismay: Should we also add the `"awesome": "yes"` field here?

Mohammer5: Haha, good idea! Jokes aside, I could add all contributors to this field, using their dhis2 address?

ismay: For me personally that's not necessary. But I don't know how the others feel.

Mohammer5: Not necessary or not wanted?

ismay:
Sidenote, this article makes some interesting point on components that do switching based on boolean props: https://spicefactory.co/blog/2019/03/26/how-to-avoid-the-boolean-trap-when-designing-react-components. I'm not sure whether I agree with everything, but I do kind of agree with the point that the author makes on separating out a component instead of creating a flag on an existing one.

package.json
ismay: Well, I don't really have a strong opinion on it, but I feel like the entire organisation owns this library. Having a couple people as the designated contributors maybe communicates a conflicting message. But as I said, I don't really have too strong of an opinion on it.

Mohammer5:
> but I do kind of agree with the point that the author makes on separating out a component instead of creating a flag on an existing one.

I don't agree with this. It's like inheritance. In case of the `Table` component we could get away with it. But when thinking about other components that can have several variants of different kinds, this gets way more complicated. The `Button` component is an excellent example for this: If we want to be consistent with this, then we'd end up with a `SmallButton` & `LargeButton`, as well as `DestructiveButton`, `PrimaryButton`, `SecondaryButton` & `DefaultButton`. Do we then also need a `SmallDestructiveButton` & `LargeDestructiveButton`? Or do we keep the variation prop for the size but use different components for the style? This will 100% introduce the following issue:

**What prop should be used to create several components & which props should stay variation props?**

It will inevitably introduce inconsistency & confusion, both for the developers as well as consumers. I def. agree with having a variation prop, but I'm struggling with how to implement that with the table component's zebra stripping.

<hr />

> I would go for a flag that is false by default, say `suppressZebraStriping`.

This is def. a good idea, but I wouldn't know how it'd simplify the internal logic?

I'm not a big fan of the name `zebraStriping` as it's not automatically clear what it means, but you're already the second person to suggest that name. Do you prefer it over `suppressAlternatingRowColor`?

ismay:
> It will inevitably introduce inconsistency & confusion, both for the developers as well as consumers. I def. agree with having a variation prop, but I'm struggling with how to implement that with the table component's zebra stripping.

Yeah I see what you mean. I think that his point on how certain flags will interact and create implicit hierarchies and require custom propType checks is spot on though. But that's also something we've already discussed.

> I'm not a big fan of the name zebraStriping as it's not automatically clear what it means, but you're already the second person to suggest that name. Do you prefer it over suppressAlternatingRowColor?

Ok, yeah I'm not sure how common the term is. I feel it's fairly common: https://alistapart.com/article/zebrastripingdoesithelp. I think I'd prefer it yeah, but I think the most important part of the suggestion to me is not the specific term but the `supress-` part.

Mohammer5:
I've renamed the prop to `suppressZebraStriping`, seems like it's indeed a commonly used name for alternating colors on data tables.

packages/core/src/Table/TableRow.js
```diff
@@ -15,13 +17,29 @@ const tableRowStyles = css`
...
+        suppressZebraStriping: suppressZebraStripingFromContext,
+    } = useContext(TableContext)
 
-        <style jsx>{tableRowStyles}</style>
-    </tr>
-)
+    const zebraStriping =
+        typeof suppressZebraStriping !== 'undefined'
+            ? !suppressZebraStriping
+            : !suppressZebraStripingFromContext
```
ismay, line 33: I would expect a table to have (or not have) zebra striping in its entirety. Is there a reason that the setting can be overridden with a prop on an individual TableRow as well? I would expect users to always use the TableRow with our Table.

packages/core/src/Table/Table.js
```diff
@@ -20,12 +21,19 @@ const tableStyles = css`
...
-export const Table = ({ children, className, dataTest }) => (
-    <table className={className} data-test={dataTest}>
-        {children}
+export const Table = ({
+    children,
+    className,
+    dataTest,
+    suppressZebraStriping,
+}) => (
+    <Provider value={{ suppressZebraStriping }}>
```
ismay, line 30: I was thinking whether it makes sense to set a default value for suppressZebraStriping in the Table's defaultProps, instead of in the context. Instinctively I'd have gone with defaultProps since it feels like it belongs more to the Table to me. What are your thoughts on that @Mohammer5?

packages/core/src/Table/TableRow.js
Mohammer5: It's because we've defined that we won't use context unless it's overwritable by props and that props always have priority over context. @varl didn't want to have context in the core library at first until we agreed on the compromise to allow context as long as it can be overwritten with props.

packages/core/src/Table/Table.js
Mohammer5: That's what I did initially when the default value was `true`. But after having changed the prop, I thought it be better to be consistent with all our other components where a missing flag means `false` and we don't "mention" that in the default props.

packages/core/src/Table/TableRow.js
ismay: Ah, I didn't know that. Ok, I understand the intention. It keeps our components under user control at the component-level if they want that.

I do think that it makes for a bit of a confusing api though. It feels like we're using context to allow our components to communicate, so that we can have a simple top-level api on the Table component. Feels like that design intention runs kind of contrary to having components that are completely isolated. A bit like we're designing two different types of APIs.

I think we have that issue in some of our other components as well. Maybe we should discuss that on a more global level. Maybe you could say that one is the expected usage, and the other an escape hatch for more experienced devs? Though I wonder in what situation a dev would want to override the context and set the prop themselves.

packages/core/src/Table/Table.js
ismay: Ok, so the default in the context is not as necessary. All right, yeah I guess there's not really any real harm in having it, other than that there's an implicit default in the Table and an explicit one in the context.

PR #212, opened 2020-07-10
feat(transfer): add prop selected to renderOption (TECH-393)

### LINKED ISSUE #105: Transfer: Support custom style for selected (right side) items
https://jira.dhis2.org/browse/TECH-393

As per the design for the new dimension selection modals ([design spec](https://www.sketch.com/s/bf6d1509-6ff2-4b0c-998d-3b8cd206a47b/a/Qqbeyx) partly attached below) the selected items (on the right side) should use a separate style compared to the available items (on the left side). 

This is currently not supported, as all custom items are passed as children to `Transfer`, which in turn internally handles the population of the right side items (using the same custom items that were used for the left side). In addition, the `selected` prop only receives `label` / `value`, not the custom element itself. This leaves no intuitive way to add custom styling to the right side items.

-----------------------
Part of the design spec
image

martinkrulltott:
Fixes #105 
Fixes [TECH-393](https://jira.dhis2.org/browse/TECH-393)

Adds a `selected` prop (bool) to the `Transfer` component's `renderOption`. This enables consumers to use `selected` in the `renderOption` function to provide separate styles for the source options and picked options, in addition to the already existing `highlighted` prop.

-----------------

_Story example: Custom List Options_
image

### DIFF
--- a/packages/widgets/src/Transfer/Transfer.js
+++ b/packages/widgets/src/Transfer/Transfer.js
@@ -218,6 +218,7 @@ export const Transfer = ({
                                         toggleHighlightedSourceOption
                                     ),
                                     highlighted,
+                                    selected: false,
                                 })}
                             </Fragment>
                         )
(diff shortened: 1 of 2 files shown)


martinkrulltott:
@Mohammer5, added 2 pre-selected options to the story 👍 
image

PR #216, opened 2020-07-21
refactor(button): render data-test attribute only on button

### LINKED ISSUE #215: Button dataTest prop is broken
The `Button` component accepts a `dataTest` prop, but passes it to the `ButtonBase` component as a regular html data-attribute:

https://github.com/dhis2/ui/blob/master/packages/core/src/Button/Button.js#L83

The `ButtonBase` component also does not accept or render the `dataTest` prop. This should be fixed so that a `dataTest` prop can be passed to the `Button` and will be rendered as a `data-test` data attribute for testing.

ismay:
Closes #215

### DIFF
--- a/packages/core/src/Button/Button.js
+++ b/packages/core/src/Button/Button.js
@@ -80,7 +80,7 @@ export class Button extends Component {
                 destructive={destructive}
                 small={small}
                 large={large}
-                data-test={dataTest}
+                dataTest={dataTest}
                 disabled={disabled}
                 name={name}
                 onBlur={this.handleBlur}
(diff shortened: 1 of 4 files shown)


ismay:
> code looks good to me, but I'd consider it a `refactor` rather than a `fix` as it has worked before already (albeit unintentionally) as the button base forwarded all unknown props directly to the button element.

I don't know, I can see how it _should_ work, but in the scheduler it didn't work for me. Passing a dataTest prop. didn't render a data-test attribute on the ButtonBase for me there.

PR #220, opened 2020-07-28
feat(transfer): add requested improvements

Mohammer5:
* [x] TECH-378: Add `onSourceEndReached & `onPickedEndReached` callbacks
* [x] TECH-379: Add `loadingSource` & `loadingPicked` flags
   *Rename to `loadingOptions` & `loadingSelected` for consistency?*
* [x] Add filtering mechanism to picked options side
* [x] Cover new functionality with cypress tests
  * [x] TECH-378
  * [x] TECH-379
* [x] Discuss inconsistencies I found (require a breaking change..)
  * `option + selected` vs `source + picked` vs mixtures of the two
  * Props `sourceEmptyPlaceholder` & `selectedEmtpyComponent`
    * `source` & `selected`
    * `Placeholder` & `Component`
  * filtering props for the selected list end with `Picked`
    * Should it end with `Selected`?
    * Other props have `Selected`/`Picked` not at the end, but in the middle of the prop name
  * Some props use `left`/`right`, not `option`/`selected` (possible confusion once component has been adjusted to adapt to smaller screen sizes)


packages/widgets/src/Transfer/SourceOptions.js
```diff
@@ -1,23 +1,39 @@
...
 import React from 'react'
 import propTypes from '@dhis2/prop-types'
 
-import { spacers } from '@dhis2/ui-constants'
+import { EndIntersectionDetector } from './EndIntersectionDetector.js'
 
 export const SourceOptions = ({
     children,
     dataTest,
     sourceEmptyPlaceholder,
```
ismay, line 10: For the `PickedOptions` this seems to be called `selectedEmptyComponent`, here it's `sourceEmptyPlaceholder`. I would expect them to have similar types of names.

Mohammer5: Yes, I've listed some of the inconsistencies in the PR's description that we need to talk about and whether we want to introduce a breaking change

packages/widgets/src/Transfer/Transfer.js
```diff
@@ -108,6 +119,16 @@ export const Transfer = ({
...
 
+    const [internalFilterPicked, setInternalFilterPicked] = useState(
+        initialSearchTermPicked
+    )
+    const actualFilterPicked = onFilterPickedChange
+        ? searchTermPicked
+        : internalFilterPicked
+    const actualFilterPickedCallback = filterablePicked
+        ? filterCallbackPicked
+        : identity
```
ismay, line 130: It's a bit unclear what's going on in this area.

packages/widgets/src/Transfer/features/notify_at_end_of_list/index.js
```diff
@@ -0,0 +1,84 @@
+import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Cypress.on('uncaught:exception', err => {
+    // This prevents a benign error:
+    //   This error means that ResizeObserver was not able to deliver all
+    //   observations within a single animation frame. It is benign (your site
+    //   will not break).
+    //
+    // Source: https://stackoverflow.com/a/50387233/1319140
```
ismay, line 9: I checked out the answer. While it might not be indicative of a problem in the usecase mentioned on stack overflow, it might be in our case. Lea Verou has posted a related issue in the Resize Observer spec repo here that's interesting: https://github.com/WICG/resize-observer/issues/38. I would still be interested in what's causing the error exactly before we ignore it.

packages/widgets/src/Transfer/SourceOptions.js
ismay: Ah sorry, I missed that. 👍

packages/widgets/src/Transfer/OptionsContainer.js
```diff
@@ -0,0 +1,144 @@
+import { CircularLoader } from '@dhis2/ui-core'
+import { spacers } from '@dhis2/ui-constants'
+import React, { Fragment, useEffect, useRef, useState } from 'react'
+import propTypes from '@dhis2/prop-types'
+
+import { EndIntersectionDetector } from './EndIntersectionDetector.js'
+
+export const OptionsContainer = ({
```
ismay, line 8: I know we talked about the IntersectionDetector not working properly in some cases. I suspect that you've addressed that issue here with the resize observer? Or is it still an issue?

I think what I'd prefer as a reviewer to fix the intersection detector bug is a separate PR with a failing test as the initial commit. So that we can then test our solutions against that. From the reviewer's (or my) perspective that'd be a nice approach.

One other thing I'm still skeptical about with the IntersectionDetector component by the way is that the name is directly tied to the api. The way I see it, we're interested in detecting visibility. That's why I preferred VisibilityDetector in the other PR for this component, since it seemed like that was what it had to do.

Now instead of the function we've tied the naming to the api. And now that we're encountering issues where the api isn't fulfilling our needs of detecting visibility we can't fix it within the component, because it doesn't fit the scope of it. I think I'd still prefer a visibility detector component that neatly packages that functionality. That way we can just use that component to detect when the end of the list has been reached, and fix bugs within that component.

packages/widgets/src/Transfer/Transfer.js
Mohammer5: I've tried to make the code a bit more clear. Is it better now?

packages/widgets/src/Transfer/features/notify_at_end_of_list/index.js
Mohammer5: It's working for now. I think we should investigate the exact cause, but for this PR I think it should be fine?

packages/widgets/src/Transfer/OptionsContainer.js
Mohammer5: Yes, I've addressed it already. The problem with the visibility component in this case is, that the fix had to be done for a DOM element that's specific to the transfer.

I'd prefer to stay close to the actual api the component is implementing. The component is actually testing an intersection. The issue that came up wasn't an issue of the intersection detector alone, but it interacting with the resize observer.

I think the only possible solution I can think of is a `VisibleContainer` and a `VisibilityDetector` which have to be used in combination, as the setup will always be something like this:


```
container
  > some custom content
    > detector
```

The issue is that the `detector` needs to be a child of the custom content, which would increase the complexity of the component. I prefer to stick to the actual functionality of the component as it's clear what you can use it for. If we ever want to have some kind of visibility detector component, it should be separate of course

martinkrulltott:
Regarding the naming convention and possibly renaming props, I don't mind a breaking change to increase the consistency and to make it easier to use / understand. However, as the 2.35 soft freeze is only 2 weeks away and we've had issues with breaking changes in `ui` before, I'd prefer if we could refrain from breaking changes for the changes needed for the 2.35 release. Ping @janhenrikoverland for second opinion on BC's

PR #224, opened 2020-07-31
docs(getting-started): adds text for peer dependency installation 

paschalidi:
Hey 👋 

This PR adds a small section in the getting started section. This is because while we were updating the `capture-app` to use the `@dhis2/ui` we got into the scenario that we had to install as peer dep the `styled-jsx` 

Happy to hear your feedback on the text :)

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -8,6 +8,12 @@
 yarn add @dhis2/ui
 ```
 
+In some cases you might need to install `styled-jsx` as peer dependency. To do so run
+
+```bash
+yarn add styled-jsx -P
+```
+
 ## Requirements
 
 ### React >= 16.3
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -8,10 +8,16 @@
 yarn add @dhis2/ui
 ```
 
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -8,6 +8,12 @@
 yarn add @dhis2/ui
 ```
 
+In some cases you might need to install `styled-jsx` as peer dependency. To do so run
```
Mohammer5, line 11: I think we can be more precise here:

> When not using the dhis2 app platform, then the ui library requires you to install some peer dependencies:
> * React
> * ReactDom
> * styled-jsx
> 
> Using the dhis2 app platform you won't have to install them as they're bundled with the platform scripts

docs/getting-started.md
```diff
@@ -8,6 +8,12 @@
...
+
+```bash
+yarn add styled-jsx -P
+```
```
Mohammer5, line 15: With regards to the comment above, we could add all three peer dependencies

paschalidi: yep

paschalidi:
@Mohammer5 could you have another look when you have time? :)

PR #238, opened 2020-08-13
fix(button): toggled button disabled style

### LINKED ISSUE #237: Button: Incorrect hover style when disabled and toggled
Hovering a disabled toggled Button displays a dark hover state, as seen in [this example](https://codesandbox.io/s/ui5-button-5jekn?file=/src/App.js).

image

Birkbjo:
Forgot to add `disabled`-styles to toggled-button in #234 .  

Also adds some minor style fixes; using off-white color/fill. Keeps border the same on state-changes.

Fixes #237


Birkbjo:
> Looks good. Just `cursor: not-allowed` is missing
 that is inherited from `button:disabled`. Other `disabled:`-states does not specify that either.

HendrikThePendric:
I haven't ran this branch locally, but I don't think we need to explicitly add "cursor: not-allowed", because we already have this:
https://github.com/dhis2/ui/blob/f2542a1ef8aa302ecaff33d7e27730c732e46419/packages/core/src/Button/Button.styles.js#L31-L33

PR #249, opened 2020-08-20
fix(transfer): pass correct selected value to renderOption

### LINKED ISSUE #247: Transfer: selected prop not provided anymore
As per [ui@5.1.2](https://github.com/dhis2/ui/pull/212), a `selected` prop was provided to the custom options to enable specific styling for the right side items. However, as per the latest ui release (ui@5.5.0 at the time of writing), this prop isn't provided anymore.

Currently, the [demo for custom options ](https://ui.dhis2.nu/demo/?path=/story/transfer--custom-list-options) looks like this (screenshot below), which doesn't include the orange styling for the right side (as the selected prop isn't provided anymore.
image

Compared to 5.1.2:
image

Live example here: https://codesandbox.io/s/ui5-transfer-right-side-style-bkgrc
Change the dependency to 5.5.0 to see the difference in the styling.

This creates a bug in [DHIS2-8807](https://jira.dhis2.org/browse/DHIS2-8807) which is due to be released for 2.35 (soft freeze 25/8)

Mohammer5:
Fixed #247 

When adding the last couple of features, I've refactored the code and a bug snack in.
I've covered the functionality with a test so this particular bug should never appear again.

### DIFF
--- a/packages/widgets/src/Transfer/OptionsContainer.js
+++ b/packages/widgets/src/Transfer/OptionsContainer.js
@@ -14,6 +14,7 @@ export const OptionsContainer = ({
     loading,
     renderOption,
     options,
+    selected,
     selectionHandler,
     toggleHighlightedOption,
 }) => {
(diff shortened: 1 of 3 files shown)


martinkrulltott (approved):
LGTM

Issue #89, opened 2020-05-20
Statusicon rendering causing alignment issues


ismay:
See this thread: https://dhis2.slack.com/archives/CBM8LNEQM/p1589961187000500

Two related problems:

### 1. Some of our field components have a margin-right to space the input and a potential icon:

- https://github.com/dhis2/ui-core/blob/05512df65f1a7f59dc4ee38b28927c1514896658/src/SingleSelect.js#L110
- https://github.com/dhis2/ui-core/blob/05512df65f1a7f59dc4ee38b28927c1514896658/src/MultiSelect.js#L110 (and maybe others)

This margin is always present, even if there's not an icon. We should make it conditional and not set a margin if there's no icon.

### 2. In the input we use a wrapper for the status-icon, which always renders a margin-left, throwing off alignment as well.

We should make this conditional as well, and should probably use the same approach everywhere. (Could be present in other components as well btw.)

https://github.com/dhis2/ui-core/blob/05512df65f1a7f59dc4ee38b28927c1514896658/src/Input.js#L68

See here: https://dhis2.slack.com/archives/CBM8LNEQM/p1589962613009000?thread_ts=1589961187.000500&cid=CBM8LNEQM

HendrikThePendric:
> If you mean a conditional wrapper in the parent then that would be similar to the conditional margin-right, I'd prefer sth in that direction as well

I can't remember if that is exactly what I meant, but after reading this thread again, I do think that this approach makes most sense. Having a margin on the actual icon is probably a bad idea, for the reasons you mention. Having a wrapper with a (fixed) margin and conditionally rendering that would be the cleanest solution IMO, because if you used a dynamic/conditional margin we end up rendering an empty div with `margin-left: 0;` as its only style.

varl:
https://jira.dhis2.org/browse/LIBS-80

PR #262, opened 2020-09-01
 feat: introduce pagination component and unit tests

HendrikThePendric:
**UPDATE**
- I've added feature files, cypress tests and docs, so since all the technical aspects are done, I'm taking this PR out of draft status
- But I think the UI is likely to change a little bit, depending on the input from @cooper-joe 

----

Some notes:
- I still need to add feature files and cypress tests.
- The current implementation is **not according to the design specs**, instead I have tried to implement it according to decisions from [this Slack thread](https://dhis2.slack.com/archives/CBM8LNEQM/p1598867816009500?thread_ts=1598865771.000900&cid=CBM8LNEQM).
- I think there are some issues with the current implementation we should address:
    - The icons don't match the design
    - The component is very text-heavy, I wonder if we can reduce this somehow
    - Due to the fact that there is so much text, the component is not suitable for narrow screens. I've applied a HTML structure and some styles that make this component neatly break into different lines when it doesn't fit. I wouldn't quite call it responsive design 😆 , but at the very least the component remains usable. However in the current storybook, this is already happening at around a 1000px. And that's for the English language. I dread to think how this would look for a language that uses more characters per word. So I reckon we should really try to reduce the amount of text being used.

@cooper-joe could you review the current state of this component and offer guidance on how to improve this? You can preview the current state of this component [here](https://5f4e69b1f2b0d3000742f445--dhis2-ui.netlify.app/demo/?path=/story/pagination--default).

_Suggestion:_
How about we change:
- `Number of items per page` to `Items per page`
- `Viewing page ## of ##, item ###-### of ####` to `Page ## of ##, item ###-### of ####`

That will shave off a few characters 😄


packages/widgets/i18n/en.pot
```diff
@@ -55,3 +55,20 @@ msgstr ""
 
 msgid "Could not load children"
 msgstr ""
+
+msgid "Number of items per page"
```
cooper-joe, line 59: "Items per page" will reduce the character count.

packages/widgets/i18n/en.pot
```diff
@@ -55,3 +55,20 @@ msgstr ""
...
+msgstr ""
+
+msgid ""
+"Viewing page {{page}} of {{pageCount}}, item {{firstItem}}-{{lastItem}} of "
+"{{total}}"
```
cooper-joe, line 64: I suggest removing the "Viewing page 1 of 10...", I'm not sure that is really that relevant. It could also be displayed as the selected item in the 'go-to page' select if it is important (which I think is already being done with `availablePages`?).

So this could then be reworked to "25–50 of 300"

packages/widgets/src/Pagination/Pagination.styles.js
```diff
@@ -0,0 +1,34 @@
...
+import { colors, spacers } from '@dhis2/ui-constants'
+
+export default css`
+    .container {
+        display: flex;
+        flex-wrap: wrap;
+        justify-content: space-between;
+        color: ${colors.grey700};
+        padding: ${spacers.dp12} 0 ${spacers.dp12} ${spacers.dp12};
+    }
```
cooper-joe, line 11: let's also set `font-size: 14px;`

cooper-joe (commented):
I've added some comments to the relevant code. Other comments:

- Could the second `section` with the text "1–20..." be aligned right? I think that would reduce the overwhelming feeling. `24px` margin-right to the buttons should be ok.

- Not all instances of pagination will need all of the features. I think the following should be included as optional props:
`itemsPerPageSelector`
`goToPageSelector`

packages/widgets/i18n/en.pot
HendrikThePendric: We do lose some information when we get rid of this text. The current page can be seen directly in the dropdown, but information about the total number of pages is hidden. The user would have to open the "Go to page" select and see the last option to know how many pages there are.

Plus, the "Go to page" select should be an optional element, so we can't rely on it to be a unique source of information.

I think my preference would be to keep this information but just reduce the character-count a little bit. How about:
`Page 1 of 10, item 25-50 of 300`?

HendrikThePendric:
> * Not all instances of pagination will need all of the features. I think the following should be included as optional props:
>   `itemsPerPageSelector`
>   `goToPageSelector`

Good point! I hadn't implemented that part of the specs at all yet 😊 

I agree with the above, but I think we should assume they want to be displayed by default, and should be hidden explicitly, so I'd probably suggest: `hidePageSizeSelect` and `hideGoToPageSelect`.

packages/widgets/i18n/en.pot
cooper-joe: Yeah, that's a good point that the select might not always be used. I can't think of any usecase where it's really useful to know the page number, but I'm sure there are out there. I agree with your conclusion and suggestion of `Page 1 of 10, items 25-50 of 300`.

cooper-joe:
> I agree with the above, but I think we should assume they want to be displayed by default, and should be hidden explicitly, so I'd probably suggest: hidePageSizeSelect and hideGoToPageSelect.

That's fine by me 👍

HendrikThePendric:
OK @ismay and/or @Mohammer5, I think everything is ready for final review now:
- The component is working
- Cypress tests are in place
- Docs are in place
- Joe has reviewed the UI, and I've implemented his feedback
- Commits have been squashed

Looking forward to your comments!

packages/widgets/src/Pagination/Pagination.js
```diff
@@ -0,0 +1,161 @@
...
+ * @returns {React.Component}
+ *
+ * @example import { Pagination } from @dhis2/ui
+ * @see Specification: {@link https://github.com/dhis2/design-system/blob/master/molecules/pagination.md|Design system}
+ * @see Live demo: {@link /demo/?path=/story/pagination--default|Storybook}
+ */
+const Pagination = ({
+    dataTest,
+    hidePageSizeSelect,
+    hideGoToPageSelect,
```
ismay, line 22: This feels like a little bit of an awkward API to me. Is it the intention to allow the dev to make the component more compact?

If so, I'd remove these in favor of a `compact` flag. That's inline with what some of our other components do, and seems to reflect the intention better. Also, it wouldn't feel as much like a double negative, which always makes it easier to understand at a glance.

packages/widgets/src/Pagination/Pagination.js
```diff
@@ -0,0 +1,161 @@
...
+    pageSize,
+    total,
+    pageSizes,
+    onPageChange,
+    onPageSizeChange,
+}) => {
+    const firstItem = Math.min((page - 1) * pageSize + 1, total)
+    const lastItem = Math.min(firstItem + pageSize - 1, total)
+    const availablePages = Array.from({ length: pageCount }, (_x, i) =>
+        (i + 1).toString()
```
ismay, line 34: This seems ripe for abstracting into a separate file and then unit testing them.

packages/widgets/src/Pagination/Pagination.js
```diff
@@ -0,0 +1,161 @@
...
+                        {pageSizes.map(length => (
+                            <SingleSelectOption
+                                key={length}
+                                value={length}
+                                label={length}
+                            />
+                        ))}
+                    </SingleSelect>
+                </div>
+            )}
```
ismay, line 63: It would be more readable for me if areas like this were separate components, so say `PageSizeSelect`, `Summary`, `PageControls` and `PageSelect` for example. Then they'd also be easier to test separately with jest.

packages/widgets/src/Pagination/Pagination.styles.js
```diff
@@ -0,0 +1,45 @@
+import css from 'styled-jsx/css'
+import { colors, spacers } from '@dhis2/ui-constants'
+
+export default css`
+    .container {
+        display: flex;
+        justify-content: flex-end;
+        flex-wrap: wrap;
+        padding: ${spacers.dp12};
```
ismay, line 9: This padding seems like it is on the outside container:

image

Did we not agree to only apply spacing inside a component? Maybe something else has been decided in the meantime, but I still think that it's a good idea to not having spacing outside of a library component.

packages/widgets/src/Pagination/Pagination.styles.js
```diff
@@ -0,0 +1,45 @@
+import css from 'styled-jsx/css'
+import { colors, spacers } from '@dhis2/ui-constants'
+
+export default css`
+    .container {
+        display: flex;
+        justify-content: flex-end;
+        flex-wrap: wrap;
```
ismay, line 8: The wrapping for smaller viewports seems a bit awkward. Maybe we could make this a bit nicer without resorting to media queries (since mq's don't seem like a good idea from a library perspective?):

Screenshot 2020-09-08 at 12 11 22

ismay:
Visually, the design feels a bit difficult to parse for me. I think it's due to two things:

* The mixing of different control patterns (select and buttons)
* The information text about the page/total page, amount of items, etc.

I understand that the controls are grouped according to function, and not visually, which makes a lot of sense of course, but it does seem to cause a bit of visual clutter. I tried moving things around to where it seemed the simplest to me purely from how it looks:

image

(That's missing the information though and groups things differently.)

packages/widgets/src/Pagination/Pagination.js
HendrikThePendric: In general, I think we are adding these props mainly for the reason you suggested, to present users with a cleaner and more compact UI. In theory it could also be because a certain resource doesn't support a pageSize parameter... Probably this is only theoretical...

I agree the prop names are a bit awkward, but don't think you suggestion works because these are two elements and I think it needs to be possible to hide each one individually. Perhaps @cooper-joe can confirm.

If the above is true, the current prop names are the best I can come with, suggestions would be welcome 😄 ...

You mentioned something about double negatives, and I think this is a relative point to consider. First of all, this is not a double negative, even if you want to classify `hiding` as a negation then the words still only contain a single negative. But I do agree that having property names that instruct a component what it **should do** processes a lot better than ones that instruct the component what it **should not do**.....

In the past I have used positive words and `defaultProps` to sort of work around this. So then you'd get `showPageSizeSelect` with a default value of true, and the consumer of the component would have to set its value to `false` to hide that section. I suppose we could do that here too....

packages/widgets/src/Pagination/Pagination.styles.js
HendrikThePendric: @cooper-joe I agree with @ismay on this: I think it'd be better to remove this. Would you be OK with too?
Since t's coming from [the design system](https://www.sketch.com/s/2f0dba87-fb82-415b-863d-0c2ea63ee77b/a/pmVrZo), so I'm keeping it here until you confirm.

packages/widgets/src/Pagination/Pagination.styles.js
HendrikThePendric: 🤔  hmmm thanks for identifying this issue. I had actually looked into this before I implemented Joe's input (all his comments are gone because I squashed my commits 😢 ), and it looked a lot better before (not great but at least better). When I've started aligning stuff on the right this must have gotten this way. I'll look into it.

packages/widgets/src/Pagination/Pagination.js
ismay: Yeah I meant double in the way that hide... means _don't_ display, and the default is false. So the default is don't don't display componentX.

> In the past I have used positive words and defaultProps to sort of work around this. So then you'd get showPageSizeSelect with a default value of true, and the consumer of the component would have to set its value to false to hide that section. I suppose we could do that here too....

Yeah, that also seems like an awkward api. I wouldn't go in that direction either. If it's possible to rework the props to a positively worded one where the default is false, I'd prefer that. Otherwise I wouldn't know a better alternative either.

packages/widgets/src/Pagination/Pagination.styles.js
HendrikThePendric: Fixed the alignment itself in eecd7b9 but this produced a problem when `PageSizeSelect` is not rendered (I guess that's why I was aligning at the flex-end). So I fixed that with a spacer div in ab1e6d5

Mohammer5:
In the two stories `Without page size select` and `Without go to page select` there's a difference in space on the right side. In the former story, there are 4 more pixels which come from the select's input.
I think we should add this rule:

```css
.root-input:last-child {
    margin-right: 0;
}
```

Do you guys agree and - if yes - think we should address that in this PR or in a new issue + PR

packages/widgets/src/Pagination/Pagination.js
```diff
@@ -0,0 +1,124 @@
...
+                />
+            )}
+            <PageSummary
+                dataTest={dataTest}
+                page={page}
+                pageCount={pageCount}
+                pageSize={pageSize}
+                total={total}
+            />
+            <div className="section">
```
Mohammer5, line 57: Could/should be its own component (e. g. `PaginationSection`) which would reduce the complexity of importing the section styles in multiple places which causes multiple style tags and adheres to the styled-jsx philosophy

cooper-joe:
> Visually, the design feels a bit difficult to parse for me.

Thanks for raising your concerns @ismay, looking again now I agree that there is something that is making the full pagination component (using all controls) difficult to parse.

After some experiments, I think the issue could be the mixing of static text with controls. It could be confusing what is clickable and what is not, what is a label for a control and what is an information label. I've moved some things around, switched component variations, and come up with this idea:
image

I think these adjustments have a number of improvements:
- Using selects with prefix-labels means that the only static text that is not inside a control is the page and item number information,
- The previous/next buttons are moved to always be the most right-aligned element, this is where users expect to find these controls,
- The component feels less text heavy

What do you guys think? I hope these changes are not too disruptive to that which has been built already, but I do think it is worth the improvement. 
And also, thanks for taking time to discuss the component design, it will definitely result in a better final product 😃

packages/widgets/src/Pagination/Pagination.styles.js
cooper-joe: Yeah, I agree with removing this. I think the spec was made before we made that decision. Sorry for the out-of-date spec.

packages/widgets/src/Pagination/PageSummary.js
```diff
@@ -0,0 +1,50 @@
...
+    const lastItem = Math.min(firstItem + pageSize - 1, total)
+
+    return { firstItem, lastItem }
+}
+
+const PageSummary = ({ dataTest, page, pageCount, pageSize, total }) => {
+    const { firstItem, lastItem } = getItemRange(page, pageSize, total)
+
+    return (
+        <div className="section" data-test={`${dataTest}-summary`}>
```
Mohammer5, line 19: If I remember correctly, the preferred approach is to supply the data-test values in the root component. I can see that we have mixed approaches in our codebase, so I'm not sure I remember the correct variant. @ismay do you remember? I think you were the one advancing a standard here. It's currently a convention so I'd prefer to not address it until we have a linter plugin or so, but I don't know how strongly you feel about this

packages/widgets/src/Pagination/PageSummary.js
```diff
@@ -0,0 +1,50 @@
...
+            <span>
+                {i18n.t(
+                    'Page {{page}} of {{pageCount}}, item {{firstItem}}-{{lastItem}} of {{total}}',
+                    {
+                        page,
+                        pageCount,
+                        firstItem,
+                        lastItem,
+                        total,
+                    }
```
Mohammer5, line 29: To me this makes the jsx dom tree harder to parser. I'd prefer this translation payload to be outside of the return statement. But it's just a personal preference, so I leave it up to you and what you prefer

packages/widgets/src/Pagination/PageSummary.js
```diff
@@ -0,0 +1,50 @@
...
+}
+
+const PageSummary = ({ dataTest, page, pageCount, pageSize, total }) => {
+    const { firstItem, lastItem } = getItemRange(page, pageSize, total)
+
+    return (
+        <div className="section" data-test={`${dataTest}-summary`}>
+            <span>
+                {i18n.t(
+                    'Page {{page}} of {{pageCount}}, item {{firstItem}}-{{lastItem}} of {{total}}',
```
cooper-joe, line 22: I think `item` should be `items` here. There is a chance there is only a single item, but more often there will be multiple I think.

packages/widgets/src/Pagination/PageSummary.js
Mohammer5, line 19: See `Pagination.js`'s comment about the section component

packages/widgets/src/Pagination/PageSummary.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+import propTypes from '@dhis2/prop-types'
+import i18n from '@dhis2/d2-i18n'
+import { spacers } from '@dhis2/ui-constants'
+
+import { sectionStyles } from './Pagination.styles.js'
+
+const getItemRange = (page, pageSize, total) => {
+    const firstItem = total === 0 ? 0 : (page - 1) * pageSize + 1
+    const lastItem = Math.min(firstItem + pageSize - 1, total)
```
Mohammer5, line 10: Could you write comments about what's happening here, I have a hard time understanding what's going on without reading into the code and it's logic

packages/widgets/src/Pagination/PageSizeSelect.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import propTypes from '@dhis2/prop-types'
+import i18n from '@dhis2/d2-i18n'
+import { SingleSelect, SingleSelectOption } from '@dhis2/ui-core'
+import { spacers } from '@dhis2/ui-constants'
+import { sectionStyles, selectStyles } from './Pagination.styles'
+
+const PageSizeSelect = ({ dataTest, pageSize, pageSizes, onChange }) => (
+    <div className="section" data-test={`${dataTest}-pagesize`}>
```
Mohammer5, line 9: See `Pagination.js`'s comment about section

Mohammer5: There are additional styles here which I think should have their own css rule inside this file that's provided to the `PaginationSection` component I'm proposing, that would add meaning to the styles through the rule's name

packages/widgets/src/Pagination/PageSizeSelect.js
```diff
@@ -0,0 +1,44 @@
...
+        {i18n.t('Items per page')}
+        <SingleSelect
+            dense
+            selected={pageSize.toString()}
+            onChange={({ selected }) => onChange(parseInt(selected))}
```
Mohammer5, line 14: The `parseInt` call lacks the radix (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt)

packages/widgets/src/Pagination/PageSelect.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react'
+import propTypes from '@dhis2/prop-types'
+import i18n from '@dhis2/d2-i18n'
+import { SingleSelect, SingleSelectOption } from '@dhis2/ui-core'
+import { spacers } from '@dhis2/ui-constants'
+import { selectStyles } from './Pagination.styles.js'
+
+const createAvailablePages = length =>
+    Array.from({ length }, (_x, i) => (i + 1).toString())
```
Mohammer5, line 9: I think it's always better to not use abbreviations. They just save a very small amount of coding time but increase the cognitive parsing time every time someone has to read this code, it's prefer the following version:

```js
const createAvailablePages = length =>
    Array
        .from({ length })
        .map((_, index) => (index + 1).toString())
```

To me it's a lot easier to understand what's going on this way

Mohammer5: I don't feel very strong about this though, if you prefer a short version, just keep it the way it is and resolve this remark

Mohammer5 (changes requested):
Generally I think ALL texts should come from props and should have their default translation as the default value for that prop. That way apps can add languages that have not been covered by this library or change if when necessary. As this component is already using `flex-wrap: wrap` it shouldn't be an issue if translations become too long which could happen with our translations as well anyways

HendrikThePendric:
> In the two stories `Without page size select` and `Without go to page select` there's a difference in space on the right side. In the former story, there are 4 more pixels which come from the select's input.
> I think we should add this rule:
> 
> ```css
> .root-input:last-child {
>     margin-right: 0;
> }
> ```
> 
> Do you guys agree and - if yes - think we should address that in this PR or in a new issue + PR

This is due to https://github.com/dhis2/ui/issues/89. This needs to be fixed in the SingleSelect, not in the Pagination.

packages/widgets/src/Pagination/PageSummary.js
HendrikThePendric: a329a8a

packages/widgets/src/Pagination/PageSummary.js
HendrikThePendric: > to supply the data-test values in the root component.

I thought this was the preferred approach for internal components. The root gets the base `dataTest` value and the internal components get values derived from that.

packages/widgets/src/Pagination/Pagination.js
HendrikThePendric: I liked this suggestion, but when I tried implementing it I ran into some issues with overriding the styles for the `SingleSelect`s using `styled-jsx`. Also, there were only three sections and each one is slightly different. Long story short: in the end I duplicated a few lines of CSS in 3 components and got rid of the shared styles instead, see e03f38b.

I'm resolving this comment, because I already discussed this briefly on Slack

packages/widgets/src/Pagination/Pagination.styles.js
HendrikThePendric: Removed in d362a52

packages/widgets/src/Pagination/PageSizeSelect.js
HendrikThePendric: Good catch, I always forget them. Didn't we used to have a linter rule that would pick this up? 2286e50

packages/widgets/src/Pagination/PageSummary.js
ismay: Yeah what Hendrik is saying. https://github.com/dhis2/ui/blob/master/packages/core/src/MultiSelect/MultiSelect.js#L97

Though I see it's not the case everywhere. But the above was the strategy when I originally added them.

ismay:
> Generally I think ALL texts should come from props and should have their default translation as the default value for that prop. That way apps can add languages that have not been covered by this library or change if when necessary.

Doesn't that go against our strategy of using transifex?

ismay:
> > Visually, the design feels a bit difficult to parse for me.
> 
> Thanks for raising your concerns @ismay, looking again now I agree that there is something that is making the full pagination component (using all controls) difficult to parse.
> 
> After some experiments, I think the issue could be the mixing of static text with controls. It could be confusing what is clickable and what is not, what is a label for a control and what is an information label. I've moved some things around, switched component variations, and come up with this idea:
> image
> 
> I think these adjustments have a number of improvements:
> 
>     * Using selects with prefix-labels means that the only static text that is not inside a control is the page and item number information,
> 
>     * The previous/next buttons are moved to always be the most right-aligned element, this is where users expect to find these controls,
> 
>     * The component feels less text heavy
> 
> 
> What do you guys think? I hope these changes are not too disruptive to that which has been built already, but I do think it is worth the improvement.
> And also, thanks for taking time to discuss the component design, it will definitely result in a better final product 😃

Ah yes, I forgot about the internal prefixes for the select. Feels easier to parse this way to me, thumbs up from me for this version 👍. Thanks for taking a look at it!

packages/widgets/src/Pagination/Pagination.js
ismay: > each one is slightly different.

That also makes me prefer the direction Hendrik took here 👍

packages/widgets/src/Pagination/PageSummary.js
ismay: Yeah that seems nice to have 👍

packages/widgets/src/Pagination/PageSizeSelect.js
ismay: There are linter rules for this (airbnb enables it for instance), so I assume we've just not enabled that rule.

packages/widgets/src/Pagination/PageSelect.js
ismay: I agree with the sentiment. Though on the other hand i is a so common that it wouldn't confuse me as much. I like using an underscore instead of `_x` though. Seems more obvious that it's unused to me.

ismay:
I was thinking. Would it be useful to get feedback from others, beyond the people who have commented in the PR currently? In the interest of aligning with the ones using this lib in the wild? Maybe post a heads up in #UI?

Mohammer5:
> Doesn't that go against our strategy of using transifex?

I don't think so. Transifex should still translate everything. But 3rd party apps might want to use different wordings here and there and although it's unlikely that Pagination texts are being overridden, I'd prefer to keep things consistent.

But there's a chance of course I just misinterpreted the way we want to handle translations. Are you guys aware of any discussion and reached conclusions or does everyone have his/her own interpretation?

ismay:
> > Doesn't that go against our strategy of using transifex?
> 
> I don't think so. Transifex should still translate everything. But 3rd party apps might want to use different wordings here and there and although it's unlikely that Pagination texts are being overridden, I'd prefer to keep things consistent.
> 
> But there's a chance of course I just misinterpreted the way we want to handle translations. Are you guys aware of any discussion and reached conclusions or does everyone have his/her own interpretation?

Actually yeah, I agree. It's the way we've dealt with translations in other areas as well (see the translations for the other components in widgets: https://github.com/dhis2/ui/tree/master/packages/widgets/src). All text comes from props, and the default values are our translations. So 👍 from me for your suggestion.

HendrikThePendric:
💯  from me too for mentioning the translations, good catch @Mohammer5!  I will do it the same way as we did for the `*Field` components. I'll have to touch the `translate` helper too since we need to support interpolation now.

packages/widgets/src/Pagination/PageControls.js
```diff
@@ -0,0 +1,66 @@
...
+
+const PageControls = ({
+    dataTest,
+    onClick,
+    nextPageText,
+    page,
+    pageCount,
+    previousPageText,
+}) => (
+    <div data-test={`${dataTest}-pagecontrols`}>
```
cooper-joe, line 17: I think we should use `secondary` type buttons for the 'Previous' and 'Next' buttons.

packages/widgets/src/Pagination/PageSummary.js
HendrikThePendric: aba32bf

HendrikThePendric:
OK, I think I've addressed everything that was mentioned, please review again.

packages/widgets/src/Pagination/PageSelect.js
```diff
@@ -0,0 +1,52 @@
...
+    <div data-test={`${dataTest}-gotopage`}>
+        <SingleSelect
+            dense
+            selected={page.toString()}
+            onChange={({ selected }) => onChange(parseInt(selected, 10))}
+            className="select"
+            dataTest={`${dataTest}-page-select`}
+            prefix={translate(pageSelectText)}
+        >
+            {createAvailablePages(pageCount).map(availablePage => (
```
Mohammer5, line 27: I think `availablePage` is ambiguous and would prefer `pageNum` or sth like that. But it's  not important at all, so take it or leave it

packages/widgets/src/Pagination/PageSummary.js
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Suppose page is 2 and pageSize is 50 the calculations below
+     * will produce: firstItem 51 and lastItem 100.
+     * But if the total is 0 (no items found) firstItem and
+     * lastItem should both be 0.
+     * And if we are on the last page, the lastItem should not
+     * be derived from the firstItem but should be equal to the total.
+     * Math.min helps fixes both these special cases.
+     */
+    const firstItem = Math.min((page - 1) * pageSize + 1, total)
```
Mohammer5, line 17: This was still confusing me. I assume that `page` starts with `1`, not `0`, otherwise this wouldn't make much sense.
The comments I was referring to was more like:

```js
// `page - 1` -> Make 1-based pages 0-based
// `pageSize + 1` ->
//    pageSize would be last item of prev page,
//    pageSize + 1 first item of current page
const firstItem = Math.min((page - 1) * pageSize + 1, total)
```

I like your explanation as well though, so please keep that one! I mostly care about adding the information that page is 1-based.

Mohammer5: you could add steps to make it more clear without having to explain multiple steps in one comment

```js
const pageIndex = page - 1 // 1-based -> 0-based
const offset = pageIndex * pageSize
const firstItem = Math.min(offset + 1, total) // Make 1-based
```

cooper-joe (approved):
LGTM 👍 Great work, thanks for making the request changes (multiple times).

packages/widgets/src/Pagination/PageSelect.js
HendrikThePendric: The problem is that these are string representations of numbers, so introducing number into the variable names makes this confusing in another way...

Mohammer5 (approved):

ismay (approved):

PR #264, opened 2020-09-04
refactor: use different styles for centering CenteredContent/Modal

Mohammer5:
See commit history

### DIFF
--- a/packages/core/src/CenteredContent/CenteredContent.js
+++ b/packages/core/src/CenteredContent/CenteredContent.js
@@ -12,24 +12,31 @@ import propTypes from '@dhis2/prop-types'
  */
 const CenteredContent = forwardRef(
     ({ className, dataTest, children, position }, ref) => (
-        <div className={cx(className, position)} data-test={dataTest} ref={ref}>
-            {children}
+        <div
+            className={cx('centered-content', className, position)}
+            data-test={dataTest}
+            ref={ref}
+        >
+            <div className="centered-inner-content">{children}</div>
+
             <style jsx>{`
-                div {
-                    position: absolute;
-                    left: 50%;
+                .centered-content {
+                    width: 100%;
+                    height: 100%;
+                    display: flex;
+                    justify-content: space-around;
+                    flex-direction: row;
+                    pointer-events: none;
+                    align-items: center;
                 }
-                div.middle {
-                    top: 50%;
-                    transform: translate(-50%, -50%);
+                .centered-content.top {
+                    align-items: flex-start;
                 }
-                div.top {
-                    top: 0;
-                    transform: translate(-50%, 0);
+                .centered-content.bottom {
+                    align-items: flex-end;
                 }
-                div.bottom {
-                    bottom: 0;
-                    transform: translate(-50%, 0);
+                .centered-inner-content {
+                    pointer-events: all;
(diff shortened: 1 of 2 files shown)


HendrikThePendric (commented):
This solution introduces a wrapper div that covers the entire parent. This is probably why the `onClose` test for the `Modal` is now failing. I see the reason moving away from an absolute positioning strategy though. I can think of a solution using pointer-events that would probably solve the click issues and still keep this flexbox implementation. Will produce a commit in a moment.

HendrikThePendric (commented):
Found another potential issue here:
Screenshot 2020-09-07 at 10 02 49
Since the Modal is now using `top` it is displaced from the parent content. I'm not 100% sure this will be an actual issue, or if it just looks funky in DevTools. We could explore just using margins instead for this type of thing...

HendrikThePendric:
> We could explore just using margins instead for this type of thing...

Yeah that seems to work slightly better.

packages/core/src/Modal/Modal.js
```diff
@@ -67,28 +67,34 @@ export const Modal = ({
...
         </CenteredContent>
+
+        <style jsx>{`
+            aside {
+                overflow-y: hidden;
+                height: auto;
+                max-height: calc(100vh - ${2 * spacersNum.dp64}px);
+                max-width: calc(100vw - ${2 * spacersNum.dp64}px);
+                width: 600px;
+                position: relative;
```
ismay, line 79: I'm wondering why there is a `position: relative` here

packages/core/src/CenteredContent/CenteredContent.js
```diff
@@ -12,24 +12,31 @@ import propTypes from '@dhis2/prop-types'
...
+        >
+            <div className="centered-inner-content">{children}</div>
+
             <style jsx>{`
-                div {
-                    position: absolute;
-                    left: 50%;
+                .centered-content {
+                    width: 100%;
+                    height: 100%;
```
ismay, line 25: Dimensions will be relative to the parent element, whereas previously, the dimensions were relative to the first positioned element. That could break layouts that rely on this.

HendrikThePendric: Your observation is correct. I'd say the previous behaviour can be considered a bug. In practice, this component is mostly used in combination with a layer, or a component-cover, so in most cases the UI _should_ remain unaffected.

amcgee: Hey @HendrikThePendric @ismay, I'm not sure I understand... is the "breaking change" here the fact that there is an additional `<div>` in the DOM tree?  How are `Dimensions` relevant here?

amcgee: In any case, since the previous behavior wasn’t documented and caused issues I think labeling it a fix is fine, so long as we do some spot checks on places using it and document the new intentional behavior

HendrikThePendric: @amcgee and I discussed this on Slack using [this example](https://codesandbox.io/s/icy-surf-7nepe?file=/index.html) and decided we can classify this as a bugfix.

martinkrulltott (approved):

PR #272, opened 2020-09-15
fix(transfer): make inifinite loading work when selecting options (#257)

### LINKED ISSUE #257: Transfer: onEndReached not triggered if an item is selected
The `onEndReached` usually triggers when scrolling all the way down in the left side of the `Transfer` component. However if the user selects any item from the left side (i.e. clicks it once), the `onEndReached` doesn't trigger any more.

As seen in [this example](https://codesandbox.io/s/ui5-transfer-onendreached-when-selected-19rui?file=/src/App.js).

Mohammer5:
Fixes #257 

* Removed the `observer` variable in the `IntersectionDetector` which caused buggy behavior
* Restructured the combination of resize + intersection to not call the `onEndReached` callback twice in the beginning

### DIFF
--- a/packages/core/src/IntersectionDetector/IntersectionDetector.js
+++ b/packages/core/src/IntersectionDetector/IntersectionDetector.js
@@ -13,6 +13,7 @@ export const IntersectionDetector = ({
     onChange,
     children,
     className,
+    dataTest,
     rootRef,
 }) => {
     // Use useRef instead of useState to prevent unnecessary re-render:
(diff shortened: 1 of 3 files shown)


HendrikThePendric (approved):
I can see that this works, but I'm not very keen on the use of a `resizeCounter` and forcing a re-render on the `EndIntersectionDetector` by using this as its `key`....

However, I think addressing this without causing any breaking changes might be challenging and out-of-scope for the current PR. Probably with that in mind we should go with the current implementation now, and I will write an issue with some ideas on where to take this next...

PR #275, opened 2020-09-16
fix: use white background color for textarea instead of transparent

### LINKED ISSUE #274: Enabled textfields don't have a white background
Screenshot 2020-09-14 at 15 00 57

The textfields in this app don't seem to have a white background when enabled. That's confusing to the user because our selects (for example) have a dark background when they're disabled, and our inputs have a white background when enabled. See the screenshot above. The user might (understandably) assume that the textfield is disabled when it's actually enabled.

My suggestion would be to resolve this in UI, and use the same background color for all our enabled fields (so white, it seems). Plus, with a transparent background it might be flat out unusable if the app background is not an appropriate color.

ismay:
Closes #274 

I'm using regular white, just like our Input.

### DIFF
--- a/packages/core/src/TextArea/TextArea.styles.js
+++ b/packages/core/src/TextArea/TextArea.styles.js
@@ -10,7 +10,7 @@ export const styles = css`
         padding: 8px 12px;
 
         color: ${colors.grey900};
-        background-color: transparent;
+        background-color: white;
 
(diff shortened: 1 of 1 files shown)


Mohammer5 (approved):

awgaan (approved):
Looks good.

Issue #273, opened 2020-09-14
Content of disabled buttons is not as legible due to decreased opacity


ismay:
Our disabled buttons look like this:

<img width="842" alt="Screenshot 2020-09-14 at 15 20 51" src="https://user-images.githubusercontent.com/7355199/93091081-f2154500-f69d-11ea-84f0-ed3bd33a25c1.png">

In the style rules you can see that we're reducing the opacity. However, that also reduces the readability of button text and the icons that are rendered in it. Which becomes a problem with for example a disabled button with a loading spinner. All you can see is that the button increases size because of the spinner, but what's going on inside of it is not clear at all.

I suggest we don't make the button transparent, but instead use an opaque color to indicate a disabled state (quick sketch below):

<img width="927" alt="Screenshot 2020-09-14 at 15 23 19" src="https://user-images.githubusercontent.com/7355199/93091490-7667c800-f69e-11ea-9ecc-34bc29b67a6b.png">

For the text this does not make a visible difference, but it makes a big difference for the loading spinner:

before:
<img width="313" alt="Screenshot 2020-09-14 at 15 29 46" src="https://user-images.githubusercontent.com/7355199/93092198-70261b80-f69f-11ea-8e41-cff38078b948.png">

after:
<img width="314" alt="Screenshot 2020-09-14 at 15 30 25" src="https://user-images.githubusercontent.com/7355199/93092217-74eacf80-f69f-11ea-8c6d-048dd12000fb.png">

Not using opacity also means that the button's readability isn't influenced by the app's background color anymore.

cooper-joe:
I agree, this is causing issues. I will investigate and provide new values for disabled button background colors for all types of button.

cooper-joe:
Fixed by: https://github.com/dhis2/ui/pull/278. 

One thing I thought of while adjusting the colors: it would make sense to show a white-colored loading spinner for a primary/destructive button while it was loading. In the above example the standard spinner is shown. I'm not sure if we could set the spinner to inherit the `fill` set on the button though?

ismay:
> Fixed by: #278.
> 
> One thing I thought of while adjusting the colors: it would make sense to show a white-colored loading spinner for a primary/destructive button while it was loading. In the above example the standard spinner is shown. I'm not sure if we could set the spinner to inherit the `fill` set on the button though?

[...]

PR #278, opened 2020-09-16
fix(button): adjust disabled styles

cooper-joe:
Fixes https://github.com/dhis2/ui/issues/273 by adding opaque backgrounds to disabled buttons of type `primary` and `destructive`.

### DIFF
--- a/packages/core/src/Button/Button.stories.js
+++ b/packages/core/src/Button/Button.stories.js
@@ -38,9 +38,43 @@ export const Destructive = () => (
 )
 
 export const Disabled = () => (
-    <Button disabled name="Disabled button" value="default" onClick={logger}>
-        Label me!
-    </Button>
+    <>
+        <Button
+            disabled
+            name="Disabled button"
+            value="default"
+            onClick={logger}
+        >
+            Label me!
+        </Button>
+        <Button
+            primary
+            disabled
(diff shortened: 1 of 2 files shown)


ismay (commented):
Looks good 👍. I noticed an `opacity: 1;` [here](https://github.com/dhis2/ui/blob/870b7994d69b00d9cfda77f55ee4fef0e683694b/packages/core/src/Button/Button.styles.js#L273), I think we can remove that.

And I see some regular hex colors, but I think you mentioned before that you don't necessarily only want to use theme colors in the components, right? Just checking whether you're ok with that, or if we should've been using theme colors.

One other thing that might be nice is for us to add demos to the storybook for all the disabled button types, currently only the standard one is in there (I could add that if you'd like).

Let me know what you think!

cooper-joe:
> I noticed an opacity: 1; here, I think we can remove that.
Yes, nice spot. Removed.

> One other thing that might be nice is for us to add demos to the storybook for all the disabled button types, currently only the standard one is in there.
Good idea - added.

> And I see some regular hex colors, but I think you mentioned before that you don't necessarily only want to use theme colors in the components, right?
Yes, that's right, components have some flexibility where they need custom colors.

ismay (approved):
Looks good! 👍

PR #281, opened 2020-09-17
fix: allow empty string as select option value

### LINKED ISSUE #245: SingleSelect input does not show empty string as value
See here: https://github.com/dhis2/ui/blob/4d48cff42f2dcbc8874929f4a82aec484fb49aa5/packages/core/src/SingleSelect/Input.js#L23

If you have this kind of option:

```jsx
<SingleSelectOption value={''} label="None" />
```

since the empty string is falsey it fails the check and doesn't show the selected value if its value is `''`. I would say it should show that.

ismay:
Closes #245

I wonder whether we can call this a fix. If a user has this kind of select:

```jsx
<SingleSelect selected={""} >
  <SingleSelectOption value="1" label="1" />
</SingleSelect>
```

which was valid code (though not very useful), then this update will break their code. Previously it would skip `""` since it's falsy, but since now `""` will be a valid value, the select will look for that value in the options. When it can't find an option with that value it'll throw an error.

So for this users this would be a breaking change.

### DIFF
--- a/packages/core/src/SingleSelect/Input.js
+++ b/packages/core/src/SingleSelect/Input.js
@@ -20,7 +20,7 @@ const Input = ({
     disabled,
     inputMaxHeight,
 }) => {
-    const hasSelection = selected && typeof selected === 'string'
+    const hasSelection = typeof selected === 'string'
     const onClear = (_, e) => {
         const data = { selected: '' }
 
--- a/packages/core/src/SingleSelect/SingleSelect.js
+++ b/packages/core/src/SingleSelect/SingleSelect.js
@@ -118,7 +118,6 @@ const SingleSelect = ({
 }
 
 SingleSelect.defaultProps = {
-    selected: '',
     dataTest: 'dhis2-uicore-singleselect',
 }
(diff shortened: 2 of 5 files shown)


ismay:
Nice, that broke basically everything.

ismay:
It should be easy to fix this, but I'm having issues with xcode that prevent me from running a successful yarn install and fixing everything locally. Trying to fix xcode atm.

ismay:
So after taking a closer look at this, I think allowing `""` as a value would not only be a potential breaking change, it also goes against the final-form defaults (which use `""` as the default value for inputs, see allowNull [here](https://final-form.org/docs/react-final-form/types/FieldProps)). So that means that we also have to go against the final-form default.

Not fixing this would mean that you have do a bit of manual work if you want `""` as a value in an app. But that code can be fairly straightforward. Whereas in my opinion this change would complicate the situation for all users of the lib (and us as developers). So I'd vote to not implement this as the fix is more complicated than addressing it in an app, which shouldn't occur too often.

PR #284, opened 2020-09-22
feat(ui-icons): add icon source files

cooper-joe:
This pull request adds the source svg files for all the icons ready for ui-icons. 

The icons are organised by into folders by name, different sizes are marked by filename:

```
> svg
>> add
>>> 16.svg
>>> 24.svg
>> arrow-down
>>> 16.svg
>>> 24.svg
```

The <SVG> files have `fill: inherit` and have been compressed with [SVGO](https://github.com/svg/svgo).

I have discussed with @HendrikThePendric how these files could be parsed automatically to build components from the SVG code. I'll let this conversation continue on this PR. I am happy to hand over this PR to someone who can build the components from the source files. Or this PR could be merged and implementation can be handled on a subsequent PR, I do not mind.

A few notes about the source files/components-to-be that may or may not be important:

- it should be possible to set the icon color on components (`color` prop passed to `fill`?)
- not all sizes are present, an icon could theoretically have sizes 12, 16, 24, 32 and these may be added in future,
- the viewbox for icons is **always** its exact size. So an icon is always 24x24, even if it's drawn content is 24x20,
- icon variations are handled by having two individual icons, so there is no need for `-filled` or `-bold` variant props as was previously discussed, these would now be referenced as just `iconName-filled` and`iconName-bold`,
- all `<SVG>` are composed of a single path, not individual `<stroke>` etc.

Edit: setting this PR as draft until we decide whether to merge source files only or continue implementation on this PR.

---

Icons to be added:
image


ismay:
> I have discussed with @HendrikThePendric how these files could be parsed automatically to build components from the SVG code.

Instead of building them every time, I'm thinking if it'd be useful to script something to generate the components once, so that we don't have to deal with maintainting a build script for components that seem like they'd be easiest to deal with as static components.

I don't expect they'll change a lot, so once we have the static components we should basically be good. That would be my initial suggestion. I could draft something if it sounds like a direction we want to explore.

ismay:
I don't know if we have the component api figured out already, but I'd expect something in the direction of:

```
<IconCopy
  size={12}
  type="filled" // I'm assuming filled and bold are mutually exclusive, if not, I'd expect `filled` and `bold` flags
  color={color}
/>
```

For the name (`IconCopy` in the example). I'm thinking just `Copy` without an icon prefix wouldn't convey clearly enough that it's an icon. And I suspect an `Icon` component with a prop for selecting the icon you want, would run into treeshaking problems (since the used and unused icons can't be deduced at build time).

HendrikThePendric:
> Instead of building them every time, I'm thinking if it'd be useful to script something to generate the components _once_ and then commit just the result, so that we don't have to deal with maintainting a build script for components that seem like they'd be easiest to deal with as static components.
> 
> I don't expect they'll change a lot, so once we have the static components we should basically be good. That would be my initial suggestion. I could draft something if it sounds like a direction we want to explore.

Yeah, I haven't given it a great deal of thought beyond the point of: "we shouldn't be converting these SVG  icons to React components by hand, since we'd essentially be doing the same thing over and over again". In my mind the best approach would be to first create a script that traverses the SVG folder and generates a React component from each SVG. Once that works we try to incorporate that into CI so the React and probably put the react components folder into gitignore since this is all auto-generated stuff. During merges to master we can then hopefully build the components and release them. 

My initial thought about only producing the React components for the icons that have changed, is that the benefits (build time) don't weight up to the costs (development time). I'm thinking this because I think there are quite things to consider here, for example: when one SVG changes, we only need to change one component, but when the component template file changes we'd need to change all components. I'm sure there are more things like that....

Anyway, this sort of work is something I am very interested in, but have absolutely no experience with. So I'd like to team up with you if possible.

cooper-joe:
> I don't know if we have the component api figured out already, but I'd expect something in the direction of:
> 
> ```
> <IconCopy
>   size={12}
>   type="filled" // I'm assuming filled and bold are mutually exclusive, if not, I'd expect `filled` and `bold` flags
>   color={color}
> />
> ```

I would prefer not to pass in a `type`, but rather have separate components for the necessary types. This is to allow flexibility in different types. `-filled` is the only type right now, but I can imagine situations where we might want `-simple`, `-rounded`, `-rtl` etc. I would rather just handle all these types as individual components: `IconCopySimple`, `IconCopyFilled`, `IconCopyRTL`, rather than setting up props that need to be updated if a new type emerges (which might only be used once).

HendrikThePendric:
@cooper-joe I've started experimenting a bit with icon component generation. I've noticed that some of my assumptions are currently not correct. For example, I expected that currently _every icon_ would come in a 16 and a 24 variant. However, this is not the case. I've written some code that scans the directory and it seems the following is th case:
```
Icon set appears to be incomplete. The follwoing files were expected but not found:
 - /Users/hendrik/apps/ui/packages/icons/svg/dimension-category-option-groupset/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/dimension-data/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/dimension-data-element-groupset/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/dimension-org-unit/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/dimension-org-unit-groupset/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/export-items/16.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/import-items/16.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-area/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-area-stacked/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-bar/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-bar-stacked/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-column/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-column-multi/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-column-stacked/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-gauge/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-line/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-line-multi/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-pie/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-radar/24.svg
 - /Users/hendrik/apps/ui/packages/icons/svg/visualization-single-value/24.svg
```

I am wondering about the wider implications for this. Can I expect that we end up with a "full set", i.e. currently that every component will come in both sizes? Or will some icons just never be published in some sizes?

cooper-joe:
> For example, I expected that currently _every icon_ would come in a 16 and a 24 variant. However, this is not the case. 
>
> I am wondering about the wider implications for this. Can I expect that we end up with a "full set", i.e. currently that every component will come in both sizes? Or will some icons just never be published in some sizes?

@HendrikThePendric Some icons are very unlikely to be published in some sizes. Icons can be in size: 12, 16, 24 and 32, but I don't think there will be many that have all.

HendrikThePendric:
> @HendrikThePendric Some icons are very unlikely to be published in some sizes. Icons can be in size: 12, 16, 24 and 32, but I don't think there will be many that have all.

🤔  We might want to rethink this... When using an icon library, as a developer I would definitely expect each icon to behave the same, i.e. be available in all sizes....

ismay:
> Once that works we try to incorporate that into CI and probably put the react components folder into gitignore since this is all auto-generated stuff.

What I was thinking is that we don't go with a build script that runs on ci, but that we have a build script that we can initiate manually, after which we commit the result. So that we don't have to generate them each and every time. But just an idea.

> "we shouldn't be converting these SVG icons to React components by hand, since we'd essentially be doing the same thing over and over again"

Yeah. I mean, anything goes as far as I'm concerned. My goals here are to minimize time spent, maximize maintainability of the solution and have it be simple, easy to reproduce and reliable. It's pretty likely that that will involve automation of some sort, but I'm open to all directions.

> Anyway, this sort of work is something I am very interested in, but have absolutely no experience with. So I'd like to team up with you if possible.

Yeah I'd like that 👍

ismay:
I think that first we'll need to settle on our ideal api. And maybe discuss how often we expect icons to need changing, and if those changes can be major or just minor tweaks.

ismay:
Shall we create a separate issue to discuss the preferred api?

cooper-joe:
> > @HendrikThePendric Some icons are very unlikely to be published in some sizes. Icons can be in size: 12, 16, 24 and 32, but I don't think there will be many that have all.
> 
> 🤔 We might want to rethink this... When using an icon library, as a developer I would definitely expect each icon to behave the same, i.e. be available in all sizes....

Yes, this is very true. My comment, "I don't think there will be many that have all sizes", reflects practical time considerations rather than a proposal that this is the best way for the library to work. 

Ideally it would be possible to build the components now and update them as more sizes become available.

> And maybe discuss how often we expect icons to need changing, and if those changes can be major or just minor tweaks.

I would imagine that new icons will be added fairly often, perhaps several times per month. New sizes for existing icons (as discussed above) would likely be added over time, rather than all at once.

Ideally the build process would be automated/simple enough that I could trigger a release by adding some new SVGs. That would be pretty awesome!

HendrikThePendric:
> Yes, this is very true. My comment, "I don't think there will be many that have all sizes", reflects practical time considerations rather than a proposal that this is the best way for the library to work.
> 
> Ideally it would be possible to build the components now and update them as more sizes become available.

I think we should discuss this approach a bit further. I would argue that we should be always adding icons in all sizes. However the approach you are suggesting can probably be accommodated for as well. The code for generating components will become quite bit more complex because of it and we'll end up with a very mixed bag of icons, but at least I'd say it's technically possible.

> I would imagine that new icons will be added fairly often, perhaps several times per month. New sizes for existing icons (as discussed above) would likely be added over time, rather than all at once.
> 
> Ideally the build process would be automated/simple enough that I could trigger a release by adding some new SVGs. That would be pretty awesome!

I think that would be possible, either by manually running some command, or integrated into CI (we need to discuss that a bit more).

ismay:
Api discussion: https://github.com/dhis2/ui/issues/285
Icon size discussion: https://github.com/dhis2/ui/issues/286

varl:
Going over the comments and thoughts it seems to me that the proposal to have one component per size and variant is the preferred approach: e.g. `<ArrowFilled24 />`

The API should be restrictive as is possible to start; no need to add features that we aren't sure we need at this point.

RTL implementation should be internal to the component, so we can add it when we need it.

We can gather the involved parties and document a decision on this.

ismay:
> Going over the comments and thoughts it seems to me that the proposal to have one component per size and variant is the preferred approach: e.g. `<ArrowFilled24 />`
> 
> The API should be restrictive as is possible to start; no need to add features that we aren't sure we need at this point.
> 
> RTL implementation should be internal to the component, so we can add it when we need it.
> 
> We can gather the involved parties and document a decision on this.

I agree, that sounds good 👍

ismay:
Closed from #395

PR #287, opened 2020-09-23
fix(input field ff): forward type property of Field component

Mohammer5:
The type property is only accessible through the `input` prop that's passed to the InputFieldFF component, so it needs to forward that property. The `input.type` property is undefined when no type is provided, so I just forwarded the prop and ignored the missing type case (which will default to `input` in the `Input` component)

### DIFF
--- a/packages/forms/src/InputFieldFF/InputFieldFF.js
+++ b/packages/forms/src/InputFieldFF/InputFieldFF.js
@@ -29,6 +29,7 @@ export const InputFieldFF = ({
     <InputField
         {...rest}
         name={input.name}
+        type={input.type}
         error={hasError(meta, error)}
         valid={isValid(meta, valid, showValidStatus)}
         loading={isLoading(meta, loading, showLoadingStatus)}


ismay:
Final Form's documentation is confusing (me) here. 

I was wondering if we should be specifying type explicitly here. It's not mentioned in the docs explicitly: https://final-form.org/docs/react-final-form/types/FieldRenderProps, so that made me think it's just being passed to the input as an unrecognized property (though FF does interact with it: https://final-form.org/docs/react-final-form/types/FieldProps#type).

But, unrecognized properties get passed as `props.unrecognizedProperty`, not `props.input.unrecognizedProperty`.

So to summarize, FF passes certain props under the input key. And I can't find in their docs which props those are exactly. They do recommend spreading props.input on the input. I wonder if that would work for us. I think as an intermediate fix this could make sense, but eventually I'd like to investigate whether we should be spreading input on all our ui/forms components. Or, specifying all the ones that make sense explicitly.

Anyway, besides the above, two questions:

- I see you labeled it fix, wouldn't it be a feat?
- Do you foresee this breaking any existing functionality (I can't imagine any situations where it would)?

ismay:
Actually, we might have already caught all of them manually with what we're setting on the input explicitly.

Mohammer5:
> I see you labeled it fix, wouldn't it be a feat?

That's a good question, I'd say it's a fix because the actual `InputField` works, but our binding does not, so the binding is broken, but this def. depends on the perspective. I don't care too much about it, wouldn't mind to change it

> Do you foresee this breaking any existing functionality (I can't imagine any situations where it would)?

I don't tbh

ismay:
> > I see you labeled it fix, wouldn't it be a feat?
> 
> That's a good question, I'd say it's a fix because the actual `InputField` works, but our binding does not, so the binding is broken, but this def. depends on the perspective. I don't care too much about it, wouldn't mind to change it

Yeah that makes sense 👍

> > Do you foresee this breaking any existing functionality (I can't imagine any situations where it would)?
> 
> I don't tbh

👍

ismay (approved):

PR #288, opened 2020-09-23
feat(orgunittree): sort org units alphabetically

Mohammer5:
Fixed DHIS2-9440

I've added a `suppressAlphabeticalSorting` in case this behavior is not wanted, sorting now happens by default.
It'll change the order of displayed items in apps that already use this, but as apps don't have any control over this process, this won't break anythings, so in my opinion this does not need to be a breaking change.

### DIFF
--- a/packages/widgets/src/OrganisationUnitTree/useOrgData/helpers.js
+++ b/packages/widgets/src/OrganisationUnitTree/useOrgData/helpers.js
@@ -28,3 +28,18 @@ export const addMissingDisplayNameProps = nodes => {
 
     return fromEntries(nodesWithDisplayName)
 }
+
+export const sortNodeChildrenAlphabetically = node => {
+    if (!node.children) return node
+    const sortedChildren = [...node.children]
+
+    sortedChildren.sort((left, right) => {
+        if (left.displayName === right.displayName) return 0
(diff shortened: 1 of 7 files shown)


packages/widgets/src/OrganisationUnitTree/useOrgData/helpers.js
```diff
@@ -28,3 +28,22 @@ export const addMissingDisplayNameProps = nodes => {
...
     return fromEntries(nodesWithDisplayName)
 }
+
+export const sortNodeChildrenAlphabetically = node => {
+    if (!node.children) return node
+    const sortedChildren = [...node.children]
+
+    sortedChildren.sort((left, right) => {
+        if (left.displayName === right.displayName) {
+            console.log(0)
```
awgaan, line 38: Did you intentionally leave this here?

awgaan (approved):
Looks good.
edit: Or so I thought.
Seems like `children` is `undefined` in `computeChildNodes`.

Mohammer5:
@awgaan I just had to update the testing story, now everything is working fine again. Approval still valid or do you want to take another look?

awgaan (approved):

PR #290, opened 2020-09-24
chore(testing): demo jest cucumber

Mohammer5:
This PR is meant to demo how we could use jest cucumber, it's not meant to be merged into master.

* [x] Add jest-cucumber dep
* [x] Add simple test (checkbox indeterminate prop)
* [ ] Add test of complex feature file

Feel free to check out the branch of this PR and play around. I'll add a couple more tests today/tomorrow to demo more complex examples.

Relates to #289

PR #291, opened 2020-09-24
test: add jest only tests

ismay:
This is to have an equivalent example with jest-only tests for https://github.com/dhis2/ui/pull/290


varl:
This is so simple. :+1:

ismay:
It could be improved by using enzyme's [exists](https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/exists.html) method. I should have used that actually.

ismay:
I'll keep the branch around, but I think the PR can be closed. Was mostly for demonstration purposes.

PR #293, opened 2020-09-25
fix: adjust icon button sizes

cooper-joe:
This PR adjusts the padding on icon-only buttons for regular and dense sizes.

Previously the left/right padding was too large, causing icon-only buttons to be slightly-not-square at 38x36 and 30x28. The fix adjusts the css to make the icon-only buttons correctly sized at 36x36 and 28x28.

### DIFF
--- a/packages/core/src/Button/Button.styles.js
+++ b/packages/core/src/Button/Button.styles.js
@@ -218,7 +218,7 @@ export default css`
     }
 
     .icon-only {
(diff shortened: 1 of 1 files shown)


ismay (approved):
Looks good. This does remind me that we need to refactor the button styles (we have duplicate selectors, descending selector specificity, etc.). I'll create an issue for it.

PR #300, opened 2020-09-28
fix(alertbar): skip init on component update

### LINKED ISSUE #299: 100% cpu usage with certain use of the alertstack/alertbar
See: https://codesandbox.io/s/exciting-sun-pvnzl?file=/src/App.js
Thread is here: https://dhis2.slack.com/archives/CBM8LNEQM/p1600858550004500

Seems like something in the codesandbox above is causing an infinite loop. We should see if this is due to anything in the alertbar/alertstack.

ismay:
Fixes an infinite loop caused by an unconditional setstate on component updates. It doesn't seem necessary to re-run init when the props are updated anyway. The core problem was the setState in the `show()` method, which was called by `init()`, which was run on every update (or more precisely, when state was the same as the previous state).

* Since we're not reusing init, I've just moved what we were doing there to didMount
* Since show was called on didMount and just changing state, I'm now starting with that state on mount. That meant show could also be removed. This meant that the animation logic had to be changed, which I've done as well.

I've tried to stick to the existing code as much as possible. Starting out with the `.visible` class could maybe be refactored to starting out without a class, and then hiding when a `.hiding` class is added instead for example. But I figured that those kinds of more elaborate changes were out of scope for this PR.

Closes #299

### DIFF
--- a/packages/core/src/AlertBar/AlertBar.js
+++ b/packages/core/src/AlertBar/AlertBar.js
@@ -24,34 +24,19 @@ class AlertBar extends Component {
     // visible is used to control the show/hide animation
     // hidden is used to stop rendering entirely after hide animation
     state = {
-        visible: false,
+        visible: true,
         hidden: false,
     }
 
     componentDidMount() {
-        this.init()
-    }
-
-    componentDidUpdate(_prevProps, prevState) {
-        // Only re-init when props change, ignore state changes
-        if (
-            prevState.visible === this.state.visible &&
-            prevState.hidden === this.state.hidden
-        ) {
-            this.init()
-        }
+        this.startTime = Date.now()
+        this.timeRemaining = this.props.duration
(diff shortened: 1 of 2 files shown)


ismay:
Maybe I've changed too much, let me know if I have @HendrikThePendric.

ismay:
> I need to check if that initial change from false to true was meaningful though, it didn't look like it to me initially

Ok, it is meaningful. Let me see what I can do.

ismay:
Doing the transition with an animation makes sense to me: https://jsfiddle.net/40wrL8py. That way we can animate on mount without having to use js. It means we don't have to use the `show()` method to toggle the transition.

HendrikThePendric (approved):
Great work!

ismay:
> Great work!

Thanks! Thanks for the help debugging!

PR #304, opened 2020-09-29
feat(icons): generate React icon components from SVG files

HendrikThePendric:
I've had a stab at generating these icon components from the SVGs. This works but quite likely we are going to end up with a very different implementation. I've mainly worked on this in my spare time to get some more experience with nodejs and working with the file system. Our final solution is probably going just rely on an existing tool like this one: https://github.com/gregberge/svgr and we'll also have to discuss how we want to deal with missing sizes, props API etc in more detail.

In any case, what I have here works OK and we can, at the very least, use it to visually inspect the icons. There seem to be a few issues there, which are probably relevant for @cooper-joe:
1. Some icons aren't rendered at all
2. Other icons don't respond to a color prop

To have a look, follow these steps:
1. Checkout this branch
2. Run `yarn install`
3. Run `yarn start`
4. Open the browser and go to http://localhost:5000
5. Find the "Generated Icons" story

You'll see a great big table full of icons. If an icon is not available in a particular size I am rendering the string `N/A`. So whenever you see an empty cell, this is because the SVG is not rendering properly. When you go to inspect the element you will find that the actual SVG is in fact there, so the SVG itself must have some issues. For the icons that don't respond to the color prop, it's easiest to scan the third column where the icons are supposed to be shown in red, but some aren't.

Suppose you'd want to fix these SVGs and check if the revised version is working properly, you could do so by following these steps:
1. Replace the svg file in `packages/icons/src/svg`
2. Run the icon generation script `yarn generate:icons`
3. Run the storybook again `yarn start`
4. Go back to the "Generated Icons" story and inspect the icon you've changed


HendrikThePendric:
Closing this one in favour of the SVGR based approach as in #313

Issue #305, opened 2020-09-30
What to do with the current icons in the icons package


HendrikThePendric:
Whilst writing a comment in the icon component API discussion (issue #285) I started thinking about what to do with the icons already present. My thoughts can be summarised like so:
- I thought that the icon library was an internal thing and only used by components in `core` and `widgets`, but unfortunately it's not. We can find it [on NPM](https://www.npmjs.com/package/@dhis2/ui-icons).
- This makes removing the current icons a breaking change
- I think we need to press ahead and remove them anyway, because it just doesn't seem like these icons really belong in an icon library: they are closely tied to the component and or are too opinionated.
- I think we should remove these icons and then in `core` and `widgets` go through all the instances where we import an icon and choose one of these options:
    1. If the icon doesn't have any additional behaviour and is included/has an equivalent in the svg-icon-set, we should use that instead.
    2. If the icon has additional behaviour it should probably be colocated with the component that uses it, or be implemented as a standalone component if multiple components rely on it.
    3. If the icon doesn't have any additional behaviour and doesn't have an equivalent in the svg-icon-set, then we need to discuss with @cooper-joe what to do. Most likely course of action would be that the icon gets added to the svg-icon-set.

Does this make sense, or should we do things differently?

ismay:
I think we discussed that a while back, but I'm not sure what the outcome was. I think we discussed it when moving to the monorepo setup.

cooper-joe:
I agree with the three options you outline @HendrikThePendric. Looking through the list of current icons, I think the vast majority are covered by there. There will be a few that will require discussion about whether it should live in `icons` or with the component, but I don't foresee any big issues.

PR #313, opened 2020-10-08
ci(icons): svgr icon build proof of concept

ismay:
I think that this now covers most of the things we want. It's still very bare-bones, failing tests, etc. But they're unrelated to the central concept this POC demonstrates. I've kept it this way so that we can first discuss the approach, see that it works, without me investing too much time up front. If we all agree I can then proceed to address things more in detail. Most important things are:

* ~~Old icons: we need a strategy for these. Do we remove them and do a major version bump? Something else?~~
* The build: icons are kept in `/src` as `.svg`'s, they are built with svgr to `/build`, and an `index.js` is generated by svgr as well. The transformation is controlled by the templates in `svgr-templates`, and the icons are optimized with svgo (turn multiple paths into a single path, etc.)
* Storybook: I've created a single story in `/stories` (see below for a screenshot). For now this is just to demonstrate the principle. I'm imagining we eventually display all the icons here in a clear, searchable layout (say a ui table), with all the variations logically grouped. In this story I'm using icons that were built to `/storybook` by svgr. We could also import from `/build`, but I thought it would be cleaner to separate those from eachother since the purpose is slightly different.

There are some details to take care of, like the config files (like the d2 config) which will need to be configured properly, and building the icons before building storybook, etc. But I think that most major points have been covered.

Below a screenshot of the storybook. Not the final product, just a sketch. It'll need proper styling, and I imagine there's more grouping and information we'd want. Just to show the basic principle of rendering all icons in the lib:

Screenshot 2020-10-12 at 15 12 35


cooper-joe:
Great to see this coming together.

> Old icons: we need a strategy for these. Do we remove them and do a major version bump? Something else?

[Ongoing discussion about this](https://github.com/dhis2/ui/issues/305).

> I'm imagining we eventually display all the icons here in a clear, searchable layout (say a ui table), with all the variations logically grouped.

Agreed, we should make the icon library both easy to explore and to find exactly what you're looking for. I can mock up some design ideas for the interface. It would also be nice to include things like a 'copy name' button that puts the component `IconDownload24` on the clipboard.

ismay:
> Great to see this coming together.
> 
> > Old icons: we need a strategy for these. Do we remove them and do a major version bump? Something else?
> 
> [Ongoing discussion about this](https://github.com/dhis2/ui/issues/305).

Ah thanks, I had lost track of that one. Your suggestion there sounds good to me 👍.

> > I'm imagining we eventually display all the icons here in a clear, searchable layout (say a ui table), with all the variations logically grouped.
> 
> Agreed, we should make the icon library both easy to explore and to find exactly what you're looking for. I can mock up some design ideas for the interface. It would also be nice to include things like a 'copy name' button that puts the component `IconDownload24` on the clipboard.

Sounds good 👍!

ismay:
Opened an issue with svgr for the componentName/named export mismatch: https://github.com/gregberge/svgr/issues/501

varl:
A different proposal to: https://github.com/dhis2/ui/pull/304

Mohammer5:
Is this an alternative to the already merged PR?

ismay:
> Is this an alternative to the already merged PR?

It should have been closed, it's the predecessor. I did reference it in the PR but guess sth went wrong.

PR #317, opened 2020-10-13
chore(dependencies): remove unused package

ismay:
Closes #312 

I don't see this package being imported anywhere. The build and storybook also still work after removing. Yarn why shows this:

Screenshot 2020-10-13 at 15 15 55

I think it's safe to remove it. Seems like it's bundled in the app-shell.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,8 +74,7 @@
         "react-dev-utils": "^10.2.1",
         "react-docgen": "^5.3.0",
         "react-dom": "16.8",
-        "styled-jsx": "^3.3.0",
-        "typeface-roboto": "^0.0.75"
+        "styled-jsx": "^3.3.0"
     },
     "d2": {
         "docsite": {


amcgee:
@ismay I think it's needed here (for storybook, as @Mohammer5 said) - https://github.com/dhis2/ui/blob/master/storybook/.storybook/preview.js#L1 - so I don't think we can remove it

amcgee (changes requested):
Pretty sure we have to keep this...

ismay:
Ah. `rg` must have ignored the storybook dir. So but we don't need it in the root package then do we?

ismay:
I think the root one is just a leftover from what Viktor mentioned: https://dhis2.slack.com/archives/CBM8LNEQM/p1602598734035200

amcgee (approved):
Yep, think so 👍

PR #323, opened 2020-10-15
style(eslint): add eslint import sort rule

ismay:
A proposal for automatically sorting our imports. This rule is fixable, and means that we never have to think about having a sensible order for imports for this repo ever again. Plus it ensures it'll always be a consistent style.

See: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md


varl:
I think this should be in our cli-style eslint config globally. :thinking: :scientist:

ismay:
> I think this should be in our cli-style eslint config globally. 🤔 🧑‍🔬

I agree, that would be even better in my opinion. I've modified the rule slightly btw (no newlines in between imports, could also choose a different setting there if necessary).

Shall I create a PR in cli-style then?

varl:
> > I think this should be in our cli-style eslint config globally. thinking scientist
> 
> I agree, that would be even better in my opinion. I've modified the rule slightly btw (no newlines in between imports, could also choose a different setting there if necessary).
> 
> Shall I create a PR in cli-style then?

Yes, please do!

Could you also add an example snippet how the imports are sorted in the description?

Personally I like having newlines between the different import groups, but if it's automatic I don't _really_ care.

ismay:
> Yes, please do!

Ok, will create it in a bit 👍

> Could you also add an example snippet how the imports are sorted in the description?
> Personally I like having newlines between the different import groups, but if it's automatic I don't _really_ care.

There are different options: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#newlines-between-ignorealwaysalways-and-inside-groupsnever. We could also set it to `always` for example, in which case there would always be a newline between import groups (local, node modules, builtins, etc.). That _could_ result in a lot of newlines though (especially if you have just one import of each type for example), which is why I went with never initially.

HendrikThePendric:
I've had a quick look at the configuration options and at first glance I think I'd like something like this:
```javascript
"import/order": [
    "error", 
    {
        // I'd like to just have two groups, basically "local" vs "other"
        // perhaps "internal" should be moved to the first group
        "groups": [["external", "builtin"],["index", "sibling", "parent", "internal", "object"]],
        // Between these main groups a newline
        "newlines-between": "always",
        // And I'd like case-insensitive sorting
        "alphabetize": {
            "order": "asc",
            "caseInsensitive": true,
        }
    }
]
```
> I can put some examples in the PR.

Could you perhaps incorporate one with the config above 😊  ?

varl:
@HendrikThePendric this is the config that is used in the cli-style alpha version:

https://github.com/dhis2/cli-style/commit/122c0469ba2393835549f8d0b5d7943066f08086

```js
        'import/order': [
            SEVERITY,
            {
                'newlines-between': 'never',
                alphabetize: {
                    order: 'asc',
                    caseInsensitive: true,
                },
            },
        ],
```

Which means that it uses the default groups: `["builtin", "external", "parent", "sibling", "index"]`

And it uses no `newlines-between` because it may be inconsistent, since it allows multiple newlines between groups, and we only want one in that case: https://github.com/dhis2/cli-style/pull/297#issuecomment-709362877

varl:
Updated with `@dhis2/cli-style@alpha`.

HendrikThePendric:
I'm closing this PR since it seems to be superseded by https://github.com/dhis2/cli-style/pull/297

PR #326, opened 2020-10-19
ci: run 10x parallelization for Cypress tests

amcgee:


ismay:
Nice. Does really cut down on the execution time (from 12:44 [here](https://github.com/dhis2/ui/pull/327)) to 3:58 in this PR.

amcgee:
> Nice. Does really cut down on the execution time (from 12:44 here) to 3:58 in this PR.

Yep, I think it's an improvement - @varl has some more holistic changes in mind which he's mocked up as well.

The speed up isn't QUITE as drastic as the 12:44->3:58 implies, since this PR has a yarn cache hit while the dependabot PR has a cache miss, but I think we can definitely speed things up by increasing concurrency!

amcgee:
Actually maybe that excludes the yarn install... this PR has an e2e test runtime of 6m while the 3x version with cache miss ran for 18m

amcgee:
@varl do you think we should merge this or wait for the single-workflow change?

varl (approved):

varl:
We might as well get the benefits of this immediately while i finish up the single workflow change.

PR #358, opened 2020-11-12
fix(headerbar): allow keyboard focus on apps drawer trigger

cooper-joe:
This PR will allow keyboard focus on the apps drawer trigger in the headerbar. It also changes the focus styles of headerbar items for contrast accessbility.

The current apps drawer trigger is a `<a>` without `href` and therefore cannot be focussed. This (naive) fix uses a `<button>` instead to allow focus, and styles the button to be invisible. I am unaware if there was a reason to use an `<a>` initially that would make `<button>` the wrong choice, but I found no issues when testing.

Fixes https://jira.dhis2.org/browse/LIBS-50

### DIFF
--- a/packages/widgets/src/HeaderBar/Apps.js
+++ b/packages/widgets/src/HeaderBar/Apps.js
@@ -243,16 +243,22 @@ export default class Apps extends React.Component {
         const apps = this.props.apps
         return (
             <div ref={c => (this.elContainer = c)} data-test="headerbar-apps">
-                <a onClick={this.onToggle} data-test="headerbar-apps-icon">
+                <button onClick={this.onToggle} data-test="headerbar-apps-icon">
                     <AppsIcon className={appIcon.className} />
-                </a>
+                </button>
 
                 {this.state.show && this.AppMenu(apps)}
 
                 {appIcon.styles}
                 <style jsx>{`
-                    a {
+                    button {
                         display: block;
+                        background: transparent;
+                        padding: 0;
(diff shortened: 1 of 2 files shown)


Mohammer5 (commented):
I'm wondering why this change is necessary. The other icons (message & notifications) are focus-able but they're still `a` elements and you can cycle through all icons with tab.

cooper-joe:
> I'm wondering why this change is necessary. The other icons (message & notifications) are focus-able but they're still `a` elements and you can cycle through all icons with tab.

The other `a` elements have a `href`. `a` without a `href` is [technically not focusable](https://allyjs.io/data-tables/focusable.html#navigation-elements). Although looking closely at this documentation, it looks like we could just add a blank `href` attribute if it only needs to be tabbable, not focusable...

Mohammer5 (changes requested):
Thanks for the clarification :ok_hand: 
I wouldn't use empty `href`s as that can behave differently from browser to browser. It's valid according to the specifications, but not a good idea imo

varl (approved):
Semantics aside, I think this is a reasonable way to solve this without thinking about keyboard navigation in it's entirely.

cooper-joe:
Thanks @varl 👍. @Mohammer5 your review requests changes, but I don't see any? (I did not add empty `hrefs`, jus tested it elsewhere).

Mohammer5 (approved):
Ah sorry, must've hit the wrong radio button

PR #361, opened 2020-11-17
fix: use helpLink from system settings

varl:
- fix: use helpLink from settings end-point
- refactor: drop abstraction for profile menu


varl:
https://jira.dhis2.org/browse/DHIS2-9202

Mohammer5 (commented):
LGTM. I guess you also have to add the new response-mock to the cypress test stories as well.
:+1: for the refactored profile menu

---

Eventually we should either
* use the declarative data fetching stuff in the components that need the data instead of passing it down

or

* create a headerbar component that does not connect to the api but expects props and create a headerbar wrapper that does the api stuff and passes it to the controlled headerbar

varl:
@Mohammer5 thanks for the quick review.

And yeah, I do. I need to get Docker running on this machine to get the backend running. Will publish updated fixtures when I do.

I agree, we need to choose either A or B of those options, but if this is going in before tomorrow I want to defer that decision for now and go with what we have now.

Mohammer5:
> I want to defer that decision for now and go with what we have now.

Yeah, I meant this more as an outlook into the future

PR #362, opened 2020-11-17
fix(tooltip): leverage layer context to ensure correct stacking of the tooltip

HendrikThePendric:
This fixes bug [TECH-473](https://jira.dhis2.org/browse/TECH-473) which @turban reported. I'll provide some info:
- Our layering mechanism can be leveraged in two ways:
    1. Adding new layers via the `Layer` component. This will result in a blocking UI because it actually renders a div that covers the underlying content. For a `Tooltip` this isn't good.
    2. Using the `useLayerContext` hook to get the closest parent layer root node. You can pass this node as the second arg in a call to `React.createPortal` so that the new portal becomes a direct descendant of the parent layer node and the normal CSS z-index stacking context Just Works ™️ . I hadn't used this approach before but the Tooltip seemed to be the right candidate for it.
- Since the Tooltip was initially a class component it could not use the `useLayerContext` hook. So I refactored it as a function component.
- When adding tests I ran into a cypress bug reg. visibility assertions and `pointer-events: none;`. I filed a bug report https://github.com/cypress-io/cypress/issues/9227 and have implemented a small hack in the e2e story to make the test run as expected.

### DIFF
--- a/packages/core/src/Tooltip/Tooltip.stories.e2e.js
+++ b/packages/core/src/Tooltip/Tooltip.stories.e2e.js
@@ -2,6 +2,7 @@ import React from 'react'
 import css from 'styled-jsx/css'
 
 import { Tooltip } from './Tooltip.js'
+import { Modal } from '../Modal/Modal.js'
 
 const noPaddingStyles = css`
     :global(#root) {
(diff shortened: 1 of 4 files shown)


packages/core/src/Tooltip/Tooltip.stories.e2e.js
```diff
@@ -73,3 +74,21 @@ export const ShiftingPosition = () => (
...
+
+export const ModalWithTooltip = () => (
+    <Modal>
+        <p>
+            I am a <Tooltip content="Short content">paragraph</Tooltip> that
+            contains a tooltip.
+        </p>
+        {/* 
+            Added pointer-events: all to the popper element to circumvent
+            a cypress bug: https://github.com/cypress-io/cypress/issues/9227
```
varl, line 86: :tada:

varl (approved):

PR #400, opened 2020-12-07
chore(release): fix entrypoint configuration

ismay:
This fixes our d2-app-scripts entrypoint configuration. We were using `entryPoint`, when the docs reference `entryPoints`: https://platform.dhis2.nu/#/config/d2-config-js-reference?id=supported-properties.

It worked because we were using the default entryPoint (`src/index.js`), which d2-app-scripts falls back to.

### DIFF
--- a/packages/constants/d2.config.js
+++ b/packages/constants/d2.config.js
@@ -1,6 +1,6 @@
 module.exports = {
     type: 'lib',
-    entryPoint: {
+    entryPoints: {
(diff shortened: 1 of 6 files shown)


varl (approved):

varl:
Good catch. Why is this a draft though?

ismay:
> Good catch. Why is this a draft though?

Oh yeah sorry, it shouldn't be. It's ready.

PR #401, opened 2020-12-08
chore(yarn): lock yarn version to 1.19.1

ismay:
This locks the yarn version that will be used when installing dependencies for this monorepo to 1.19.1.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 # See https://help.github.com/ignore-files/ for more about ignoring files.
 
-.yarn
+# We use the newer format, .yarnrc.yml
 .yarnrc
 
 # Editor settings
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,0 +1,1 @@
+yarnPath: .yarn/releases/yarn-1.19.1.js
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,3 +1,4 @@
 [build]
   command = "yarn install --frozen-lockfile && yarn build && yarn docs && yarn demo"
   publish = "dist/"
+  environment = { YARN_VERSION = "1.19.1" }


amcgee:
Nice!  I think we should probably use the new, yml structured format [`.yarnrc.yml`](https://yarnpkg.com/configuration/yarnrc) - this works with both yarn 1 and yarn 2 and is a bit cleaner

The contents of the file can just be:

```yml
yarnPath: .yarn/releases/yarn-1.19.1.cjs
```

amcgee (approved):
👍

ismay:
> Nice! I think we should probably use the new, yml structured format [`.yarnrc.yml`](https://yarnpkg.com/configuration/yarnrc) - this works with both yarn 1 and yarn 2 and is a bit cleaner
> 
> The contents of the file can just be:
> 
> ```yaml
> yarnPath: .yarn/releases/yarn-1.19.1.cjs
> ```

I've changed it, but it seems like yarn doesn't pick up on .yarnrc.yml by default. The .yarnrc change was picked up by the netlify build automatically, but now that I've changed it to use a .yarnrc.yml it just installs with 1.17.0 again.

I read that we can use an env variable to point yarn to the right config file, but having to set that seems like it kind of defeats the purpose of everyone automatically using the same yarn version for installs, regardless of the environment.

Shall I revert to the .yarnrc? I don't know of a way to resolve the above so that it would automatically be picked up.

amcgee:
That is strange and quite obnoxious... seems like a netlify bug?

> 11:14:09 AM: Found yarn version (1.19.1) that doesn't match expected (1.17.0)
up to date in 0.035s

amcgee:
```
ui % yarn --version 
1.19.1
ui % rm .yarnrc.yml 
ui % yarn --version
1.22.10
```

amcgee:
For netlify we could set `YARN_VERSION` in the `netlify.yaml` file (though agree that's obnoxious...)

amcgee:
Github Actions correctly picks up the 1.19.1 version in the latest build https://github.com/dhis2/ui/pull/401/checks?check_run_id=1516999933

amcgee:
.yarnrc.yml support was added in `1.18.0`... https://github.com/yarnpkg/yarn/blob/master/CHANGELOG.md#1180

I'm not sure here, I think I'd prefer to use the newer format and deal with the real exceptions (someone running < 1.18) rather than the other way around

amcgee:
To be super safe we could include both... (sorry, done rambling now :-) )

ismay:
Ah I didn't notice that it did use the proper yarn version for the GH actions. What I do like about the yarnrc solution is that we can rely on it to work in most scenarios. On the other hand the yarnrc.yml does have nicer formatting. I can imagine including both would maybe be a bit confusing (maybe configuration could end up in both, etc.).

I'll leave the .yarnrc.yml then and add a YARN_VERSION to netlify. Then eventually we can omit the netlify config if they move to a newer version.

varl (approved):

PR #404, opened 2020-12-09
chore(yarn): add yarnrc for older yarn versions

ismay:
This adds a .yarnrc, for older versions of yarn that don't yet support the .yarnrc.yml format.


varl (approved):
This allows for both yarn1 and yarn2 so it's a good addition.

PR #406, opened 2020-12-14
fix: adjust import paths to avoid potential dependency issues

ismay:


ismay:
Closing this since this was mostly us researching, but keeping the branch around for reference, until the issue has been resolved.

PR #413, opened 2020-12-15
chore(forms): stop importing from index and import named form components

HendrikThePendric:


ismay (approved):
Looks good! One thing that'd probably be good is to add the forms scope to the fix commit.

PR #414, opened 2020-12-15
fix: prevent access-before-initialisation storybook errors

HendrikThePendric:
https://jira.dhis2.org/browse/LIBS-89


ismay:
Ok, so we've narrowed it down to either the imports in ShareBlock.js or SharingList.js. I've removed the code in those components, and just left a dummy component and all the imports. I've been going through the imports and enabling them one by one, and then rerunning the cypress tests to verify.

So far, the imports enabled in this PR are all ok. So it's one of the commented imports that's causing the problems. Will continue tomorrow.

ismay (approved):

PR #420, opened 2020-12-16
build(eslint): add more import restrictions

ismay:
- The import restriction that prevented imports of `@dhis2/ui` in the root eslint has been moved to the eslint config of the individual packages, so that it doesn't get overwritten.
- Let me know if you think the restrictions are too strict anywhere. Currently this is the hierarchy:

  - constants and icons can't import from any other ui packages
  - core can only import from icons and constants
  - widgets can only import from core, icons and constants
  - forms can import from widgets, core, icons and constants

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,23 +6,16 @@ const config = {
         cy: 'readonly',
         Cypress: 'readonly',
     },
-    rules: {
-        'no-restricted-imports': [
-            'error',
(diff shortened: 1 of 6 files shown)


varl (approved):
The hierarchy makes sense right now. Everything seems to pass. Being strict here at first and gradually loosening up if we realize we've made a mistake is OK to me.

PR #421, opened 2020-12-16
ci(eslint): only run circular dependency check on CI

ismay:
This runs our circular dependency checks only on CI, to not slow down editors, as the checks can be very cpu intensive.

See the GH docs for the `CI` env var I'm using: [GH actions docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables). Besides GH actions, `CI` is a pretty common env var for indicating that code is running on CI so it should even work with most other providers. See also here: https://github.com/watson/ci-info/blob/master/index.js#L52.

### Follow up

Also, as a follow up and if this works out, we can move this and related rules to cli-style and make it a rule we check for all our codebases: https://jira.dhis2.org/browse/CLI-19

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,14 +1,12 @@
-const { config } = require('@dhis2/cli-style')
+const cliStyle = require('@dhis2/cli-style')
 
-module.exports = {
-    extends: [config.eslintReact],
+const config = {
+    extends: [cliStyle.config.eslintReact],
     globals: {
         cy: 'readonly',
         Cypress: 'readonly',
     },
     rules: {
-        'import/no-cycle': 'error',
-        'import/no-self-import': 'error',
         'no-restricted-imports': [
             'error',
             {
@@ -18,3 +16,13 @@ module.exports = {
         ],
     },
 }
+
+// Run cpu intensive checks only on CI
+const isCI = !!process.env.CI
+
+if (isCI) {
+    config.rules['import/no-cycle'] = 'error'
+    config.rules['import/no-self-import'] = 'error'
+}
+
+module.exports = config


HendrikThePendric (commented):
Don't we need to set the env variable somewhere, like in a GitHub workflow?

Sorry, I assumed that env var wasn't there because I didn't see any changes for it. But I see we have it already:
https://github.com/dhis2/ui/blob/4aaac3858c9ea1ff2cc545e88782a3a21b3e7702/.github%2Fworkflows%2Fdhis2-verify-lib.yml#L14

So, these changes look good to me, but I'll let @varl approve (or request changes).

ismay:
> Don't we need to set the env variable somewhere, like in a GitHub workflow?

GH actions will set this the env var for you, so that you can use it to detect if your code is running on CI.

ismay:
> Don't we need to set the env variable somewhere, like in a GitHub workflow?
> 
> Sorry, I assumed that env var wasn't there because I didn't see any changes for it. But I see we have it already:
> 
> https://github.com/dhis2/ui/blob/4aaac3858c9ea1ff2cc545e88782a3a21b3e7702/.github%2Fworkflows%2Fdhis2-verify-lib.yml#L14

That shouldn't even be necessary

HendrikThePendric:
> That shouldn't even be necessary, the GH actions docs have it listed as always set to true under their default env vars for runner environments.

Thanks for pointing that out.

ismay:
> > That shouldn't even be necessary, the GH actions docs have it listed as always set to true under their default env vars for runner environments.
> 
> Thanks for pointing that out.

N.p. 👍. That's how I interpret the docs at least. But maybe I'm missing sth. Or the defaults have changed. I'll wait for Viktor's feedback.

varl (approved):
Looks alright to my eyes.

PR #422, opened 2020-12-17
fix(npm): make package information and documentation consistent and usable from npm

ismay:
https://jira.dhis2.org/browse/LIBS-91

I've gone for the same approach as the react monorepo. Concise readmes for sub-packages, and then point to the main readme and source in package.json.


varl (approved):
Thanks, looks good to me.

PR #432, opened 2021-01-04
feat(ui-icons): add visualization type icons in 24px

cooper-joe:
This PR adds 24px sizes for existing visualization type icons.


ismay (approved):
Looks good 👍

PR #433, opened 2021-01-04
chore(yarn): update yarn.lock

ismay:


amcgee (approved):
LGTM until we update the transitive deps

PR #444, opened 2021-01-11
fix(icons): add missing prop-types dependency

ismay:
The prop-types library is used in the generated react components: https://github.com/dhis2/ui/blob/d7949dad5397c1fe5d1568892c715ded233ec29c/packages/icons/templates/icon-template.js#L16

Also added react and react-dom peer dependencies.

### DIFF
--- a/packages/icons/package.json
+++ b/packages/icons/package.json
@@ -26,6 +26,10 @@
     "files": [
         "build"
     ],
+    "peerDependencies": {
+        "react": "^16.8",
+        "react-dom": "^16.8"
+    },
     "devDependencies": {
         "@svgr/cli": "^5.5.0",
         "execa": "^4.1.0",
--- a/packages/icons/package.json
+++ b/packages/icons/package.json
@@ -30,5 +30,8 @@
         "@svgr/cli": "^5.5.0",
         "execa": "^4.1.0",
         "rimraf": "^3.0.2"
+    },
+    "dependencies": {
+        "@dhis2/prop-types": "^1.6.4"
     }
 }


amcgee:
Nice @ismay thanks for this!  I think this looks good - these are the "patched dependencies" I had to specify when using yarn2 which is more strict, do we still need `ui-constants`?

https://github.com/dhis2/app-platform/pull/491/files#diff-f28e4a8116e6aa62c7ac4cdf57b0040c11bad39d45e24024f0da20ef45972715

ismay:
> Nice @ismay thanks for this! I think this looks good - these are the "patched dependencies" I had to specify when using yarn2 which is more strict, do we still need `ui-constants`?
> 
> https://github.com/dhis2/app-platform/pull/491/files#diff-f28e4a8116e6aa62c7ac4cdf57b0040c11bad39d45e24024f0da20ef45972715

Ah, no ui-constants and styled-jsx shouldn't be necessary anymore. All we're using currently is react and @dhis2/prop-types. I included react-dom as we do that in the other packages as well. I assume that's for compatibility reasons.

ismay:
As an example, this is what one of the generated components looks like:

Screenshot 2021-01-11 at 17 01 01

(the react import comes from the dynamic part of the template, the prop-types lib is hardcoded in our template)

yarn.lock
```diff
@@ -1732,7 +1732,7 @@
     i18next "^10.3"
     moment "^2.24.0"
 
-"@dhis2/prop-types@^1.6.4":
+"@dhis2/prop-types@1.6.4", "@dhis2/prop-types@^1.6.4":
```
amcgee, line 1735: Side note: where is this "pinned" @dhis2/prop-types 1.6.4 coming from?

amcgee (approved):
OK cool - I think the overrides were from before the "new" `ui-icons`

yarn.lock
ismay: Ah, I probably need to regenerate the lockfile. I'll check.

ismay: Fixed, was a remnant of the install I did in ui-icons where I specified the exact version.

PR #460, opened 2021-01-13
fix(headerbar): use <div> instead of <p> for user avatar initials

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-49

### DIFF
--- a/packages/widgets/src/HeaderBar/TextIcon.js
+++ b/packages/widgets/src/HeaderBar/TextIcon.js
@@ -1,16 +1,17 @@
 import propTypes from '@dhis2/prop-types'
 import React from 'react'
 
+const getInitials = name =>
+    name
+        .split(' ')
+        .slice(0, 2)
+        .map(n => n[0])
+        .join('')
+
 export const TextIcon = ({ name, onClick, dataTestId }) => {
-    let title = name[0]
-
-    if (name.indexOf(' ') > 0) {
-        title += name.split(' ')[1][0]
-    }
-
     return (
         <div onClick={onClick} data-test={dataTestId}>
-            <span>{title}</span>
+            <span>{getInitials(name)}</span>
 
             <style jsx>{`
                 div {
--- a/packages/widgets/src/HeaderBar/TextIcon.js
+++ b/packages/widgets/src/HeaderBar/TextIcon.js
@@ -10,7 +10,7 @@ export const TextIcon = ({ name, onClick, dataTestId }) => {
 
     return (
         <div onClick={onClick} data-test={dataTestId}>
-            <p>{title}</p>
+            <span>{title}</span>
 
             <style jsx>{`
                 div {
@@ -26,7 +26,7 @@ export const TextIcon = ({ name, onClick, dataTestId }) => {
                     cursor: pointer;
                 }
 
-                p {
+                span {
                     font-size: 14px;
                     font-weight: 500;
                     letter-spacing: 1px;
--- a/packages/widgets/src/HeaderBar/TextIcon.js
+++ b/packages/widgets/src/HeaderBar/TextIcon.js
@@ -1,7 +1,7 @@
 import propTypes from '@dhis2/prop-types'
 import React from 'react'
 
-const getInitials = name =>
+export const getInitials = name =>
     name
         .split(' ')
         .slice(0, 2)
--- a/packages/widgets/src/HeaderBar/TextIcon.js
+++ b/packages/widgets/src/HeaderBar/TextIcon.js
@@ -10,11 +10,11 @@ const getInitials = name =>
 
 export const TextIcon = ({ name, onClick, dataTestId }) => {
     return (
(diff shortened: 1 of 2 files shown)


packages/widgets/src/HeaderBar/TextIcon.js
```diff
@@ -1,16 +1,17 @@
...
+        .split(' ')
+        .slice(0, 2)
+        .map(n => n[0])
+        .join('')
 
+export const TextIcon = ({ name, onClick, dataTestId }) => {
     return (
         <div onClick={onClick} data-test={dataTestId}>
-            <p>{title}</p>
+            <span>{getInitials(name)}</span>
```
ismay, line 14: I haven't checked the layout, but are we intentionally going with a `span` as opposed to a `div`? `p` is a block level element, so if we want a semantically neutral element a `div` (also block level) seems more appropriate to me.

packages/widgets/src/HeaderBar/TextIcon.js
```diff
@@ -1,16 +1,17 @@
...
-
-    if (name.indexOf(' ') > 0) {
-        title += name.split(' ')[1][0]
-    }
+const getInitials = name =>
+    name
+        .split(' ')
+        .slice(0, 2)
+        .map(n => n[0])
+        .join('')
```
ismay, line 9: It would be nice to extract this and add a unit test for it with jest.

ismay:
One thing I forgot to mention: in the commit messages it's good to specify a topic like so:

fix(topic-here): description of work

That way changes will be grouped by topic in the changelog, and it's easier for a user to parse what happened.

packages/widgets/src/HeaderBar/TextIcon.js
mediremi: I just went with `span` as that's what was suggested in the bug report but you're right that a `div` would make more sense due to being block level.

ismay: Yeah. Maybe there's something we're not aware of. @amcgee, do you agree with the above, or is there something we're overlooking?

ismay (approved):
Looks good to me 👍! I'll leave it to Austin to give feedback on whether there's anything we've overlooked with the elements, but as far as I'm concerned this looks good to squash merge. Nice work!

packages/widgets/src/HeaderBar/TextIcon.js
```diff
@@ -1,19 +1,20 @@
...
 
+export const TextIcon = ({ name, onClick, dataTestId }) => {
     return (
-        <div onClick={onClick} data-test={dataTestId}>
-            <p>{title}</p>
+        <div className={'text-icon'} onClick={onClick} data-test={dataTestId}>
```
amcgee, line 13: Is there a reason to use `{'text-icon'}` instead of just `"text-icon"` here?

packages/widgets/src/HeaderBar/TextIcon.js
```diff
@@ -1,19 +1,20 @@
...
+            <div className={'text-icon-initials'}>{getInitials(name)}</div>
```
amcgee, line 14: Also here

packages/widgets/src/HeaderBar/TextIcon.test.js
```diff
@@ -0,0 +1,21 @@
...
+        const actual = getInitials('Name')
+        const expected = 'N'
+        expect(actual).toBe(expected)
+    })
+
+    it('should only return the first two initials', () => {
+        const actual = getInitials('A Very Long Name')
+        const expected = 'AV'
+        expect(actual).toBe(expected)
+    })
```
amcgee, line 20: Could we add a test here for names that aren't Latin characters?

It should work with utf-8 multi byte characters, but worth adding a test case

amcgee (approved):
`<div>` seems appropriate to me - nice catch.  I added a few comments and suggested adding a test, overall looks good!

packages/widgets/src/HeaderBar/TextIcon.test.js
amcgee: Here's an example name with multibyte characters `'€150 한한한'` - it's an open question whether initials like this even make sense in non-latin languages, though, and we'll want to make sure RTL is correctly handled, but that's out of scope for this change.  Let's make a Jira issue to track that future work!

packages/widgets/src/HeaderBar/TextIcon.js
mediremi: Nope. Good catch :+1:

packages/widgets/src/HeaderBar/TextIcon.test.js
mediremi: I agree that this won't make sense in many non-Latin languages and I've created the following Jira issue: https://jira.dhis2.org/browse/LIBS-99

PR #461, opened 2021-01-15
fix(transfer): move from decorator to Component Story Format for Transfer stories

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-100

It seems like using React hooks in Storybook components causes some kind of infinite loop somewhere in the rendering process and so I've had to remove the `InfiniteLoading`, `CustomFilteringWithFilterInput` and `CustomFilteringWithoutFilterInput` stories. I'm not sure why the `useState` in the `withState` HOC doesn't cause any issues though :confused:

### DIFF
--- a/packages/widgets/src/Transfer/Transfer.stories.js
+++ b/packages/widgets/src/Transfer/Transfer.stories.js
@@ -1,6 +1,5 @@
-import { SingleSelectOption, Tab, TabBar } from '@dhis2/ui-core'
 import React, { useState, useCallback } from 'react'
-import { SingleSelectField, Transfer, TransferOption } from '../index.js'
+import { Transfer, TransferOption } from '../index.js'
 
 const options = [
     {
@@ -95,6 +94,7 @@ const options = [
     },
 ]
 
+/* eslint-disable */
 const withState = (Component, initialState = []) => () => {
     const [selected, setSelected] = useState(initialState)
     const handleChange = useCallback(payload => {
@@ -103,6 +103,7 @@ const withState = (Component, initialState = []) => () => {
 
     return <Component selected={selected} onChange={handleChange} />
 }
+/* eslint-enable */
 
 export default { title: 'Transfer' }
 
@@ -114,7 +115,6 @@ export const SingleSelection = withState(({ selected, onChange }) => (
         options={options}
     />
 ))
-
 
 export const Multiple = withState(({ selected, onChange }) => (
     <Transfer
@@ -166,17 +166,20 @@ export const Filtered = withState(({ selected, onChange }) => (
     />
 ))
 
-export const FilteredPicked = withState(({ selected, onChange }) => (
-    <Transfer
-        filterablePicked
-        onChange={onChange}
-        selected={selected}
-        initialSearchTermPicked="ANC"
-        leftHeader={<h3>Header on the left side</h3>}
-        rightHeader={<h4>Header on the right side</h4>}
-        options={options}
-    />
-), options.map(({ value }) => value))
+export const FilteredPicked = withState(
+    ({ selected, onChange }) => (
+        <Transfer
+            filterablePicked
+            onChange={onChange}
+            selected={selected}
+            initialSearchTermPicked="ANC"
+            leftHeader={<h3>Header on the left side</h3>}
+            rightHeader={<h4>Header on the right side</h4>}
+            options={options}
+        />
+    ),
+    options.map(({ value }) => value)
+)
 
 export const FilterPlaceholder = withState(({ selected, onChange }) => (
     <Transfer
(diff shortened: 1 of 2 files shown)


packages/widgets/src/Transfer/Transfer.stories.js
```diff
@@ -118,36 +94,47 @@ const options = [
     },
 ]
 
-export default { title: 'Transfer', decorators: [statefulDecorator()] }
+/* eslint-disable */
```
HendrikThePendric, line 97: It'd be nice to not have to disable ESLint here. If there are valid reasons to disable the linter I would expect just a specific rule to be disabled, and probably a comment to explain why it wasn't possible to produce this code in a manner that is compliant with our ESLint rules.

HendrikThePendric (changes requested):
In general, this higher order function seems to be a good solution. However, there is a problem with this:
> so I've had to remove the `InfiniteLoading`, `CustomFilteringWithFilterInput` and `CustomFilteringWithoutFilterInput` stories

I don't think we should just drop 3 stories. They were there to showcase some of the Transfer's functionality. If we find that the new solution with the HOC prevents us from implementing these 3 stories, then we need to look for a different solution (but I can't imagine this is actually the case).

HendrikThePendric:
Also, we tend to put the issue nr in the branch name. It's not critical, just a pointer for next time.

mediremi:
By replacing all uses of hooks with class components I have managed to restore the `InfiniteLoading`, `CustomFilteringWithFilterInput` and `CustomFilteringWithoutFilterInput` stories. 

The only issue remaining is figuring out why some part of Storybook is attempting to call the `WithState` component as a function (see stack trace below), which [forces me to wrap it in a functional component](https://github.com/dhis2/ui/pull/461/files#diff-28827d1dd7aa4f6c78cb7dec55aa0f9679d7d212ff17ba5f51b09d624c2aad34R121). Unfortunately, the stack trace emitted is not very helpful as it originates from compiled Webpack code :confused:.

image

HendrikThePendric:
@ismay came up with a different approach based on the storybook v6 API. I've produced a small snippet to illustrate:
```javascript
import React, { useState } from 'react'
import { SingleSelectField, Transfer, TransferOption } from '../index.js'

export default { title: 'Transfer', component: Transfer }

const options = [
    {
        label: 'ANC 1st visit',
        value: 'anc_1st_visit',
    },
    {
        label: 'ANC 2nd visit',
        value: 'anc_2nd_visit',
    },
    {
        label: 'ANC 3rd visit',
        value: 'anc_3rd_visit',
    },
    {
        label: 'ANC 4th or more visits',
        value: 'anc_4th_or_more_visits',
    },
    {
        label: 'ARI treated with antibiotics (pneumonia) follow-up',
        value: 'ari_treated_with_antibiotics_(pneumonia)_follow-up',
    },
    {
        label: 'ARI treated with antibiotics (pneumonia) new',
        value: 'ari_treated_with_antibiotics_(pneumonia)_new',
    },
    {
        label: 'ARI treated with antibiotics (pneumonia) referrals',
        value: 'ari_treated_with_antibiotics_(pneumonia)_referrals',
    },
    {
        label: 'ARI treated without antibiotics (cough) follow-up',
        value: 'ari_treated_without_antibiotics_(cough)_follow-up',
    },
    {
        label: 'ARI treated without antibiotics (cough) new',
        value: 'ari_treated_without_antibiotics_(cough)_new',
    },
    {
        label: 'ARI treated without antibiotics (cough) referrals',
        value: 'ari_treated_without_antibiotics_(cough)_referrals',
    },
    {
        label: 'ART No clients who stopped TRT due to TRT failure',
        value: 'art_no_clients_who_stopped_trt_due_to_trt_failure',
    },
    {
        label:
            'ART No clients who stopped TRT due to adverse clinical status/event',
        value:
            'art_no_clients_who_stopped_trt_due_to_adverse_clinical_status/event',
    },
    {
        label: 'ART No clients with change of regimen due to drug toxicity',
        value: 'art_no_clients_with_change_of_regimen_due_to_drug_toxicity',
    },
    {
        label: 'ART No clients with new adverse drug reaction',
        value: 'art_no_clients_with_new_adverse_drug_reaction',
    },
    {
        label: 'ART No started Opportunist Infection prophylaxis',
        value: 'art_no_started_opportunist_infection_prophylaxis',
    },
    {
        label: 'ART clients with new adverse clinical event',
        value: 'art_clients_with_new_adverse_clinical_event',
    },
    {
        label: 'ART defaulters',
        value: 'art_defaulters',
    },
    {
        label: 'ART enrollment stage 1',
        value: 'art_enrollment_stage_1',
    },
    {
        label: 'ART enrollment stage 2',
        value: 'art_enrollment_stage_2',
    },
    {
        label: 'ART enrollment stage 3',
        value: 'art_enrollment_stage_3',
    },
    {
        label: 'ART enrollment stage 4',
        value: 'art_enrollment_stage_4',
    },
    {
        label: 'ART entry point: No PMTCT',
        value: 'art_entry_point:_no_pmtct',
    },
]

const StatefulTemplate = ({ initiallySelected, ...args }) => {
    const [selected, setSelected] = useState(initiallySelected)
    const onChange = payload => setSelected(payload.selected)

    return <Transfer {...args} selected={selected} onChange={onChange} />
}

StatefulTemplate.defaultProps = {
    initiallySelected: propTypes.array,
}

StatefulTemplate.propTypes = {
    initiallySelected: propTypes.array,
}

export const SingleSelection = StatefulTemplate.bind({})
SingleSelection.args = {
    maxSelections: 1,
    options,
}

export const FilteredPicked = StatefulTemplate.bind({})
FilteredPicked.args = {
    filterablePicked: true,
    initialSearchTermPicked: 'ANC',
    leftHeader: <h3>Header on the left side</h3>,
    rightHeader: <h4>Header on the right side</h4>,
    options,
    initiallySelected: options.map(({ value }) => value),
}
```

mediremi:
@KaiVandivier our migration to Storybook v6's [Component Story Format](https://storybook.js.org/docs/react/api/csf) in this PR might be useful for you.

HendrikThePendric (approved):
LGTM 🎉

PR #463, opened 2021-01-20
feat: introduce (data) table and related components

HendrikThePendric:
This PR was used to experiment a bit with package structures but in the end we are simply introducing a new set of widget components to build a data-table.


.storybook/preview.js
```diff
@@ -41,7 +41,11 @@ addParameters({
...
         storySort: (a, b) =>
             a[1].kind === b[1].kind
-            ? 0
-            : a[1].id.localeCompare(b[1].id, undefined, { numeric: true }),
+                ? 0
+                : a[1].id.localeCompare(b[1].id, undefined, { numeric: true }),
+    },
+    jsx: {
+        filterProps: val => val !== undefined,
+        showDefaultProps: false,
```
HendrikThePendric, line 49: These changes are not directly related to the table, but they do clean up the JSX storybook tab quite nicely:
- Omit props if their value equals `undefined`
- Omit default props

This was also [discussed in Slack](https://dhis2.slack.com/archives/CBM8LNEQM/p1612943303197900) and approved there.

If others prefer to introduce these changes in a separate PR, I would be OK with that.

packages/core/src/Box/Box.js
```diff
@@ -45,14 +45,17 @@ Box.defaultProps = {
...
+ * @prop {string} [marginTop]
  * @prop {string} [maxHeight]
- * @prop {string} [width]
- * @prop {string} [minWidth]
  * @prop {string} [maxWidth]
- * @prop {string} [dataTest]
+ * @prop {string} [minHeight]
+ * @prop {string} [minWidth]
+ * @prop {string} [overflow]
+ * @prop {string} [width]
```
HendrikThePendric, line 58: Also unrelated: Just added some jsDocs because I realised some of the props were missing.

packages/core/src/Table/Table.js
```diff
@@ -1,56 +1,59 @@
...
-    suppressZebraStriping,
-}) => (
-    <Provider value={{ suppressZebraStriping }}>
-        <table className={className} data-test={dataTest}>
-            {children}
-
-            <style jsx>{tableStyles}</style>
-        </table>
-    </Provider>
+export const Table = forwardRef(
```
HendrikThePendric, line 13: Most of the table related components can accept a ref. This is mainly with drag and drop utilities in mind. These usually need to be able to access the underlying DOM element.

packages/core/src/Table/Table.js
```diff
@@ -1,56 +1,59 @@
...
  * @prop {TableHead|TableBody|TableFoot|Array.<TableHead|TableBody|TableFoot>} [children]
  * @prop {string} [className]
- * @prop {string} [dataTest]
- * @prop {bool} [suppressZebraStriping]
+ * @prop {string} [dataTest=dhis2-uicore-table]
+ * @prop {(auto|fixed)} [layout=auto] Sets the `table-layout` property. Switching to `fixed` can prevent
+ *  style issues when dealing with a table with multiple frozen columns or when dealing with filter
+ *  elements in the table headers.
+ * @prop {string} [width=100%] Sets the `width` property. Providing an explicit width can prevent style
+ *  issues when dealing with horizontally scrolling tables with a fixed layout.
```
HendrikThePendric, line 51: Automatic table-layout is quite clever in terms of determining column width etc. However, when you are doing things like trying to apply a specific column width, or conditionally rendering elements in a cell which impacts column width, then automatic table-layout can start working against you....

To combat this, I added the `layout` prop which can be used to switch the table-layout to fixed. But this table-layout comes with a new set of issues of its own, specifically it will allow columns to shrink beyond the width required by its content. This is a problem when you implement a horizontally scrolling table and this problem can be avoided by explicitly setting a table-width (which isn't such a crazy thing to do anyway when you want to implements horizontal scrolling).

So yeah, there you have it: two pretty basic props which simply control some CSS. But the bigger picture is much more complex. I'm not sure how we would want to document this.

Also, another note about `table-layout: fixed`: I am considering implementing this as a `fixedLayout` prop of type `boolean` instead. What would have your preference?

packages/core/src/Table/TableCell.js
```diff
@@ -1,69 +1,129 @@
...
+ * @prop {boolean} [bordered]
+ * @prop {boolean} [fixed]
+ * @prop {string} [width=auto] Required when fixed
+ * @prop {string} [left=auto] Required when fixed
+ * @prop {left|center|right} [align=left]
+ * @prop {boolean} [error] Mutually exclusive
+ * @prop {boolean} [valid] Mutually exclusive
+ * @prop {boolean} [muted] Mutually exclusive
  */
 TableCell.propTypes = {
```
HendrikThePendric, line 110: The `bordered` and `large` are technically props that apply to a cell because they impact the `td` style. 

However, the design system dictates that all cells in a table should be homogenous regarding these styles and as such you could argue these props should have been part of the `Table`. 

So there is some tension there. In our previous implementation we used context to solve this puzzle (we had `suppressZebraStriping` then instead of `large` and `bordered`). The `Table` used to accept a  `suppressZebraStriping` prop and exposed that props via a context provider to all its children. The actual styles were still defined in the `TableCell`, so as whole this seemed to make quite a lot of sense....

However, we've since realised that using context in a library unwillingly turns that library into a singleton, so we decided to move away from using context.

Another way to keep these props on the `Table` and still make sure the `TableCell` knows about these props is to use `React.cloneElement()`. We've done so extensively in the past and we aren't too happy about this solution. Furthermore, in this case we'd actually be passing elements down quite a few levels in the component hierarchy: `table > tbody > tr > td`. That would make this solution even less nice in this case.

Yet another way to keep these props on the `Table`, would be to declare `td` styles in the `Table` component. But this would break the style encapsulation principle.

I discussed this with @ismay at some point, and as he pointed out: normally in an app we will be mapping over data when producing a table anyway, so setting a few props on a cell isn't going to be that inconvenient. I think he's correct and that's why in the end i decided to keep these props on the `TableCell` and `TableCellHead` instead of on the `Table`.

packages/core/src/Table/TableCell/TableCell.styles.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+    :global(tbody) > :global(tr:active) > td {
+        background-color: ${colors.blue100};
+    }
+    :global(tbody) > :global(tr.selected) > td {
+        background-color: ${colors.teal050};
+    }
+    :global(tbody) > :global(tr.selected:hover) > td {
+        background-color: ${colors.teal100};
+    }
```
HendrikThePendric, line 76: It's not recommended/ would be going against our design-system to place a `TableCell` in a `TableHead`, but it won't produce invalid HTML. So by using these global selectors I have disabled these row-hover effects in a `thead` or `tfoot`.

Interesting side-note: I had some styles here with a global `table` selector as, i.e.: 
```css
:global(table) > :global(tr:hover) > td {}
```
I had included those because I thought a `tr` as a direct descendant of a `table` was valid HTML. This turned out to be incorrect: React will throw an "Invalid DOM Nesting" error. So I removed these `table > tr` styles.

packages/core/src/Table/TableCellHead.js
```diff
@@ -1,69 +1,170 @@
...
- * @prop {string} [dataTest]
+ * @prop {string} [scope]
+ * @prop {boolean} [showFilter]
+ * @prop {'asc'|'desc'|null} [sortDirection]
+ * @prop {string} [top] Left or top required when fixed
+ * @prop {string} [width]
+ * @prop {function} [onFilterIconClick]
+ * @prop {function} [onSortIconClick] Sort icon click callback with `nextSortDirection` and `name` in payload
  */
 TableCellHead.propTypes = {
```
HendrikThePendric, line 133: I decided to keep this fully props driven because this ultimately is offers much more flexibility than having an internal state. But it does mean quite a few props:
- We need to be able to show the current sort direction `sortDirection` and attach a click handler to the sort-icon `onSortIconClick`
- For the filter we need to a) specify what the filter is, b) specify if we should show that filter right now, and c) attach a callback to the click on the filter-icon

In total that's 5 props to control 2 functionalities, which seems quite a lot, but I do think the gains in flexibility make it worth it. It's quite likely that at some point in time an app will need to control which filters are showing programatically.

packages/core/src/Table/TableRow.js
```diff
@@ -1,45 +1,75 @@
...
+            children,
+            className,
+            dataTest,
+            expandableContent,
+            selected,
+            draggable,
+        },
+        ref
+    ) => {
+        const [isExpanded, setIsExpanded] = useState(false)
```
HendrikThePendric, line 28: This relates to the comment I placed here https://github.com/dhis2/ui/pull/463/files#r576220806. Here I've actually taken the opposite decision and I'm very unsure about this part.

The trade-off I considered was as follows:
- To control the expandable content behaviour via props we'd need:   
    1. The current `expandableContent` prop which determines what the content actually is
    2. A new `showExpandableContent` (boolean) prop to toggle visibility
    3. A new `onExpandHandleClick` (function) to handle the icon click (and probably toggle the `showExpandableContent` prop from the parent component)
- I then considered how much of a practical use case this would be, and couldn't really think of any.

So considering the above, I concluded a bit of internal state wasn't too bad since it would enable us to have a smaller component API.

packages/core/src/Table/TableRow/ExpandedRowContent.styles.js
```diff
@@ -0,0 +1,22 @@
...
+    :global(tr.isExpanded:active) + tr > td {
+        background-color: ${colors.blue100};
+    }
+    tr.selected > td {
+        background-color: ${colors.teal050};
+    }
+    :global(tr.selected.isExpanded:hover) + tr > td,
+    tr.selected:hover > td {
+        background-color: ${colors.teal100};
+    }
```
HendrikThePendric, line 21: Just some styles that make the expanded content row and the previous row look like they are one.

packages/core/src/Table/Table.js
```diff
@@ -1,56 +1,59 @@
...
-    </Provider>
+export const Table = forwardRef(
+    ({ children, className, dataTest, layout, width }, ref) => {
+        return (
+            <table className={className} data-test={dataTest} ref={ref}>
+                {children}
+                <style jsx>{`
+                    table {
+                        table-layout: ${layout};
+                        border-collapse: separate;
```
HendrikThePendric, line 21: Styling a table with collapsed borders is much easier. We've had to stick with separate borders because sticky cells don't work very well with collapsed borders

packages/core/src/Table/TableRow/TableRow.styles.js
```diff
@@ -0,0 +1,36 @@
...
+        border-bottom-color: ${colors.grey300};
+    }
+    :global(tbody) > tr.draggable {
+        cursor: move;
+        user-select: none;
+    }
+    :global(tbody) > tr > :global(td.active) {
+        background-color: ${colors.blue100} !important;
+    }
+`
```
HendrikThePendric, line 36: This part also includes some styles for `td` elements. It could be that some of these can be removed. But one of the ideas behind this is also that even if a dev would place plain `td` elements in our `TableRow` they would get some functionality out of the box. Upon reflection I think I'd better check which of these styles we actually need and keep these td styles to a bare minimum.

packages/core/src/Table/TableScrollBox.js
```diff
@@ -0,0 +1,88 @@
...
+
+const { className: tableClassName, styles } = css.resolve`
+    div {
+        border: 1px solid ${colors.grey300};
+        border-top: none;
+        box-sizing: border-box;
+    }
+    div > :global(table) {
+        border: none;
+    }
```
HendrikThePendric, line 16: When using a scroll box, the box should take care of the border, not the table. So again, this is violating purely scoped styles.

packages/core/src/Table/TableCell/TableCell.styles.js
HendrikThePendric: *Interesting side-note:*
I had some styles here with a global `table` selector as, i.e.: 
```css
:global(table) > :global(tr:hover) > td {}
```
I had included those because I think a `tr` as a direct descendant of a `table` is valid HTML (I still do). However, React does not agree with me: it will throw an "Invalid DOM Nesting" error. I don't think this is right, but it was very convenient for me because it allowed me to I removed these `table > tr` styles.

packages/core/src/Table/ExpandableTableRow.js
```diff
@@ -0,0 +1,63 @@
...
+import { TableRow } from './TableRow.js'
+
+/**
+ * @module
+ * @param {TableRow.PropTypes} props
+ * @returns {React.Component}
+ * @example import { TableRow } from '@dhis2/ui'
+ * @see Live demo: {@link /demo/?path=/story/table--expandable-table-rows|Storybook}
+ */
+export const ExpandableTableRow = ({
```
HendrikThePendric, line 13: @cooper-joe this component provides the default "Click to show more" functionality. If you want to open show expanded content programatically and/or need any other custom stuff, just build your own using the `TableRow`

Mohammer5:
Very minor: The search button for inline filtering needs space between the label and the button and the search input

image

packages/core/src/Table/Table.js
Mohammer5: > What would have your preference?

@ismay, you, me and some others talked about boolean props vs string values quite a while ago and, **if I remember correctly**, we agreed that boolean props are nicer for boolean values. I'm not entirely sure though, but I think we should try to keep this consistent across our UI library

packages/core/src/Table/TableRow/TableRow.styles.js
Mohammer5: > keep these td styles to a bare minimum.

I agree with this. I wouldn't cover rare edge cases like putting a plain `td` inside the table.

packages/core/src/Table/TableScrollBox.js
Mohammer5: Wouldn't this work the other way as well? (putting `:global(.table-scroll-box) > table` in the table styles?
Or Instead let this one render the `Table` component and accept the `Table`'s props here as well?

Mohammer5 (commented):
### expandable row semantically incorrect
I know that solving this would require a lot of changes, so I think we could address it in a different PR (especially as it shouldn't be a breaking change).
The contents are rendered in an additional row, below the row that visually displays the content. The extra content should be part of the same `tr` though, so I think the following:

```js
<ExpandableTableRow expandableContent={<div>More info about this row!</div>}>
  <TableCell>
    Onyekachukwu
  </TableCell>
  <TableCell>
    Kariuki
  </TableCell>
  <TableCell>
     02/06/2007
  </TableCell>
  <TableCell>
    05/25/1972
  </TableCell>
</ExpandableTableRow>
```

Should result in sth like this:

```js
<tr>
  <td>
    <table>
      <tr> {/* the 4 cells here */}</tr/
      <tr><td>{expandableContent}</td></tr>
    </table/
  </td>
</tr>
```

### colgroup

We have a component for all table components, even if they're just wrappers for the dom elements (e. g. tbody), so we could/should do the same for the colgroup.
Alternatively we could add a prop to the table component and render them based on the prop values?

packages/core/src/Table/ExpandableTableRow.js
```diff
@@ -0,0 +1,63 @@
...
+    children,
+    className,
+    dataTest,
+    draggable,
+    expandableContent,
+    selected,
+}) => {
+    const [isExpanded, setIsExpanded] = useState(false)
+    const toggleExpandableContent = () => setIsExpanded(!isExpanded)
```
Mohammer5, line 22: If we want to make this a "proper" building block, the state and callback should come from the outside, I think (also, in case the app needs to know this state, it would have direct access if it also provides the value)

packages/core/src/Table/TableCell.js
```diff
@@ -1,69 +1,129 @@
...
+            rowSpan,
+            scope,
+            tag,
+            valid,
+            width,
+            onClick,
+        },
+        ref
+    ) => {
+        const CellTag = tag || fixed ? 'th' : 'td'
```
Mohammer5, line 37: I'm not sure how this is interpreted, but it might need:
```js
tag || (fixed ? 'th' : 'td')
```

packages/core/src/Table/TableRow.js
```diff
@@ -1,45 +1,63 @@
...
+            expandableContent,
+            selected,
+            draggable,
+        },
+        ref
+    ) => {
+        const [
+            isHoveringExpandedContent,
+            setIsHoveringExpandedContent,
+        ] = useState(false)
```
Mohammer5, line 30: I think we could get rid of this if we implemented what I'm suggesting in the main comment

packages/core/src/Table/ExpandableTableRow.js
HendrikThePendric: I'm not sure I fully agree with this, also see my comment here https://github.com/dhis2/ui/pull/463#discussion_r576308319. We already have the `TableRow` which is a proper building block and is able to show `expandableContent`. We could make this one fully props-driven as well, which would require us to expose three props i.e. `expandableContent`, `showExpandableContent`, and `onExpandHandleClick`. At this point I'd argue it doesn't there is very little added value for a component like this and I'd rather just remove it.

(I do see your point that this is not a true building block because it has internal state though)

Mohammer5: Fair enough, creating a custom one should be fairly trivial.

packages/core/src/Table/TableRow.js
HendrikThePendric: We tried this approach and discussed on Slack. We discovered that for tables using an automatic table-layout, this approach causes issues with column width. Specifically, the columns of the nested table will completely ignore the column layout in the parent table. As such we concluded that the semantically imperfect solution is what we'd go with for now.

packages/core/src/Table/TableScrollBox.js
HendrikThePendric: > Or Instead let this one render the `Table` component and accept the `Table`'s props here as well?

This is a good idea, I like having a scrolling table. I'm not quite sure whether we should have a separate component, or simply integrate it into the existing `Table`. The current `Table `component implementation is extremely simple, so I'm not worried about this becoming too complex. I guess we can implement this by simply adding a `scrollHeight` and `scrollWidth` prop to the table.

packages/core/src/Table/Table.js
HendrikThePendric: Yeah we've talked about this general topic quite often. We have chosen to implement some properties as boolean even though they aren't really booleans, such as `error`, `warning` and `valid`. We chose to stick with them and even have the custom prop-type called `mutuallyExclusive`.

I had a quick search in the codebase to see where we are using `propTypes.oneOf()` to scan comparable scenarios but couldn't really find one that is super close to this.

In general I think boolean props work for discrete choices, which I think is a slightly broader definition than just saying boolean props are suitable for boolean values. Suppose there were only two values for the CSS `table-layout` property, one called `auto` and the other called `fixed`, where `auto` was the default: in this case I would say this is a 100% fit with a boolean prop called `fixed` or `fixedLayout`.

This nearly true for `table-layout` but not quite because there is also `inherit` and `initial`, which aren't values you'd use very often.

To me both of the following options would be valid solutions:
1. Keep `layout` as a string prop but allow all valid CSS values `auto` `fixed` `initial` and `inherit`.
2. Replace with a boolean prop called `fixedLayout` / `layoutFixed`.

I guess the underlying choice is whether or not we expect we ever need to set a table-layout to anything else but `auto` or `fixed`. 

After re-reading my own reasoning above I'm inclined to go with option 1.

packages/core/src/Table/TableCell.js
HendrikThePendric: The expression on the right hand side of the `||` will not be evaluated if the part on the left is truthy, so we're OK ([reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR#short-circuit_evaluation)).

HendrikThePendric:
> ### colgroup
> We have a component for all table components, even if they're just wrappers for the dom elements (e. g. tbody), so we could/should do the same for the colgroup.

This would be my preferred approach. However, I'm not fully sure we will ever need them. Could we add them "on-demand" (i.e. not now, but when the use-case presents itself?).

> Alternatively we could add a prop to the table component and render them based on the prop values?

I think it'll be hard to represent the `colgroup` and `col` tags and their respective APIs as a prop on the `Table`:
- A `colgroup` can contain `col`
- Each `col` could have a `className` and a `span`
- etc

packages/core/src/Table/Table.js
Mohammer5: > After re-reading my own reasoning above I'm inclined to go with option 1.

:+1:

packages/core/src/Table/TableScrollBox.js
Mohammer5: I think I'm fine with either way as long as we keep styles isolated / together with the component

Mohammer5:
> Could we add them "on-demand"

:+1:

packages/core/src/Table/TableCellHead/Sorter.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export const Sorter = ({ name, sortDirection, onSortIconClick }) => {
+    const nextSortDirection = getNextSortDirection(sortDirection)
+    const handleSortIconClick = event => {
+        onSortIconClick({ name, direction: nextSortDirection }, event)
+    }
+
+    return (
+        <Action onClick={handleSortIconClick}>
```
mediremi, line 27: Maybe we should have a `title` attribute with value `Sort values` or some text that is hidden visually but visible to screen readers?

packages/core/src/Table/TableCellHead/Sorter.js
```diff
@@ -0,0 +1,64 @@
...
+import { Action } from './Action'
+
+export const ASC = 'asc'
+export const DESC = 'desc'
+export const DEFAULT = 'default'
+export const SORT_DIRECTIONS = [DEFAULT, ASC, DESC]
+
+export const getNextSortDirection = currentDirection => {
+    const currentIndex = SORT_DIRECTIONS.indexOf(currentDirection)
+    const endIndex = SORT_DIRECTIONS.length - 1
```
mediremi, line 14: `endIndex` and the ternary operator can be replaced with modulus:
```suggestion
    const nextIndex = (currentIndex + 1) % SORT_DIRECTIONS.length
```

packages/core/src/Table/TableCellHead/Action.js
```diff
@@ -0,0 +1,29 @@
+import { colors } from '@dhis2/ui-constants'
+import PropTypes from 'prop-types'
+import React from 'react'
+
+export const Action = ({ onClick, children }) => (
+    <span onClick={onClick}>
```
mediremi, line 6: Using a `button` instead of a `span` would make `Action` keyboard accessible - users would be able to use the enter key to trigger `onClick`.

mediremi: Diff showing style changes needed:

```diff
diff --git a/packages/core/src/Table/TableCellHead/Action.js b/packages/core/src/Table/TableCellHead/Action.js
index deeec87..a881417 100644
--- a/packages/core/src/Table/TableCellHead/Action.js
+++ b/packages/core/src/Table/TableCellHead/Action.js
@@ -3,10 +3,12 @@ import PropTypes from 'prop-types'
 import React from 'react'
 
 export const Action = ({ onClick, children }) => (
-    <span onClick={onClick}>
+    <button onClick={onClick}>
         {children}
         <style jsx>{`
-            span {
+            button {
+                border: none;
+                padding: 0;
                 background: transparent;
                 width: 24px;
                 height: 36px;
@@ -15,12 +17,13 @@ export const Action = ({ onClick, children }) => (
                 justify-content: center;
                 flex-shrink: 0;
             }
-            span:hover {
+            button:hover, button:focus {
+                outline: none;
                 background: ${colors.grey400};
                 cursor: pointer;
             }
         `}</style>
-    </span>
+    </button>
 )
 
 Action.propTypes = {
```

packages/core/src/Table/Table.stories.js
```diff
@@ -585,12 +889,379 @@ export const CustomAlternatingBgColor = () => (
...
+        lastName: '',
+        incidentDate: '',
+        lastUpdated: '',
+    })
+    const onFilterIconClick = ({ name, active }) => {
+        setVisibility({
+            ...visibility,
+            [name]: active,
+        })
+    }
```
mediremi, line 1172:
```suggestion
    const onFilterIconClick = ({ name, show }) => {
        setVisibility({
            ...visibility,
            [name]: show,
        })
    }
```

cooper-joe:
We may have discussed this at an earlier point, but I did not find any documentation of the discussion, so I'll ask here. I also remember there being some difficulties with this when I worked on the spec:

Would it be possible for the columns used for checkbox, expand-arrow and draggable controls use a smaller `width`? Right now, they are expanding to fill the width like the other cells (as expected). Ideally, these cells will not take up any more room than they need.
For example, the checkbox column looks like this:
image

But ideally, it should look more like this:
image

This would apply to the selection (checkbox), drag-handle and expand-arrow cells/columns.

packages/core/src/Table/Table.stories.js
```diff
@@ -585,12 +889,379 @@ export const CustomAlternatingBgColor = () => (
...
+    const onFilterInputChange = ({ name, value }) => {
+        setValues({
+            ...values,
+            [name]: value,
+        })
+    }
+    return (
```
mediremi, line 1179:
```suggestion
        const rows = [
        {
            firstName: 'Onyekachukwu',
            lastName: 'Kariuki',
            incidentDate: '02/06/2007',
            lastUpdated: '05/25/1972',
        },
        {
            firstName: 'Kwasi',
            lastName: 'Okafor',
            incidentDate: '08/11/2010',
            lastUpdated: '02/26/1991',
        },
        {
            firstName: 'Siyabonga',
            lastName: 'Abiodun',
            incidentDate: '07/21/1981',
            lastUpdated: '02/06/2007',
        },
    ]
    const filteredRows = rows.filter(row =>
        Object.entries(values).every(
            ([prop, value]) => !value || row[prop].toLowerCase().includes(value)
        )
    )
    return (
```

packages/core/src/Table/Table.stories.js
```diff
@@ -585,12 +889,379 @@ export const CustomAlternatingBgColor = () => (
...
+                    <TableCell>Okafor</TableCell>
+                    <TableCell>08/11/2010</TableCell>
+                    <TableCell>02/26/1991</TableCell>
+                </TableRow>
+                <TableRow>
+                    <TableCell>Siyabonga</TableCell>
+                    <TableCell>Abiodun</TableCell>
+                    <TableCell>07/21/1981</TableCell>
+                    <TableCell>02/06/2007</TableCell>
+                </TableRow>
```
mediremi, line 1263:
```suggestion
                {filteredRows.map(row => (
                    <TableRow>
                        <TableCell>{row.firstName}</TableCell>
                        <TableCell>{row.lastName}</TableCell>
                        <TableCell>{row.incidentDate}</TableCell>
                        <TableCell>{row.lastUpdated}</TableCell>
                    </TableRow>
                ))}
```

cooper-joe:
As the stories are written, they are currently replacing the previous "Table" examples. The [Design System](https://github.com/dhis2/design-system) defines two table types, the regular "table" and "data table". Does this component overwrite the previous `Table`? 

I'm not 100% stuck on having two types, we _could_ have a single table component, but I think then there are some other things we'd need to support, like:
- non-interactive rows, so the ability to potentially turn off the :hover or :active states
- row striping on/off

ismay (commented):
I've read up a little on the linked discussion, what I thought is that we'd decided on going with this approach:

> On the other hand, by producing only smaller, composable components we remain flexible, have a smaller API surface, and keep the maintenance burden down. The obvious downside is that we will start seeing various apps implementing very similar code to glue these components together.

The way I understood that is that we'd go with publishing a recipe that composes low level table components to create a data-table. The components in this PR incorporate logic that goes beyond the native table elements. The way I understand the long term roadmap of ui-core is that we're going to go with these kinds of components:

```jsx
const Tr = (children, ...props) => {
  // So just basic styles, no logic
  return (
    <tr {...props}>
      {children}
      <style>some styled-jsx styles here</style>
    </tr>
  )
}
```

That would mean that we'd have to revert some of the logic in these components eventually (unless I'm misunderstanding our long-term goals for ui-core).

I can see how the props are for the users' convenience, but I think that they're a little too high-level for the long-term goals of ui-core . For example, the ConditionalScrollBox, the tag swapping in TableCell, the sorting logic in TableCellHead, the ExpandedRowContent in TableRow, etc. I think for composability's sake, that that level of complexity would be more at home in the app, or at the very least on a higher level in ui. That keeps the table components generic and doesn't tie them down too much to a specific usecase.

edit: Apologies for commenting so late btw. I kind of missed this during the whole scheduler release and everything.

mediremi:
> I can see how the props are for the users' convenience, but I think that they're a little too high-level for the long-term goals of ui-core . For example, the ConditionalScrollBox, the tag swapping in TableCell, the sorting logic in TableCellHead, the ExpandedRowContent in TableRow, etc. I think for composability's sake, that that level of complexity would be more at home in the app, or at the very least on a higher level in ui. That keeps the table components generic and doesn't tie them down too much to a specific usecase. Otherwise it could be hard later on to appropriate these table components for another usecase, if one arises.

Yep it'd be nice to e.g. have the data table track filter visibility and value state rather than making users do that every time they want a filterable table. A potential solution would be to have a higher level `DataTable` widget (similar to the old `d2-ui` component) that uses the `Table` related components. That way most consumers could have sorting and filtering with just `<DataTable columns=["A", "B"] rows=[...] sortable filterable />` while those who need more flexibility would use `Table`.

ismay:
> > I can see how the props are for the users' convenience, but I think that they're a little too high-level for the long-term goals of ui-core . For example, the ConditionalScrollBox, the tag swapping in TableCell, the sorting logic in TableCellHead, the ExpandedRowContent in TableRow, etc. I think for composability's sake, that that level of complexity would be more at home in the app, or at the very least on a higher level in ui. That keeps the table components generic and doesn't tie them down too much to a specific usecase. Otherwise it could be hard later on to appropriate these table components for another usecase, if one arises.
> 
> Yep it'd be nice to e.g. have the data table track filter visibility and value state rather than making users do that every time they want a filterable table. A potential solution would be to have a higher level `DataTable` widget (similar to the old `d2-ui` component) that uses the `Table` related components. That way most consumers could have sorting and filtering with just `<DataTable columns=["A", "B"] rows=[...] sortable filterable />` while those who need more flexibility would use `Table`.

Yes that was also my understanding of the abstractions we were planning to move to. More elaborate logic and specific usecases in widgets. Low-level, unopinionated, style-only components in ui-core.

HendrikThePendric:
OK it seems that despite of the [notes discussion](https://github.com/dhis2/notes/issues/157) and the detailed list of components and their APIs in the PR description (a list which has been available since the creation of this PR), I managed to produce components which vastly differ from people's expectations. I'm going to let all of this sink in a bit, think about where to take this PR and how to find consensus about this.

mediremi:
I think the API can be made to fit our expectations better with two changes:
1. Change the name of each of these components from `Table<X>` to `DataTable<X>` so that the existing `Table` components are not affected and users do not have to deal with breaking changes
2. Provide a component under the `ui-widgets` package that provides a data table with just a few core features like filtering and sorting out-of-the-box (i.e. no need for the consumer to keep track of state themselves). That way consumers with a specific use case can rely on the lower level `DataTable<X>` components while most use cases can be fulfilled with a simpler API

ismay:
> OK it seems that despite of the [notes discussion](https://github.com/dhis2/notes/issues/157) and the detailed list of components and their APIs in the PR description (a list which has been available since the creation of this PR), I managed to produce components which vastly differ from people's expectations. I'm going to let all of this sink in a bit, think about where to take this PR and how to find consensus about this.

Sorry about that, I feel that that's a little bit of a thing I've triggered just now. Didn't mean to throw a spanner in the works.

> I think the API can be made to fit our expectations better with two changes:
> 
>     1. Change the name of each of these components from `Table<X>` to `DataTable<X>` so that the existing `Table` components are not affected and users do not have to deal with breaking changes
> 
>     2. Provide a component under the `ui-widgets` package that provides a data table with just a few core features like filtering and sorting out-of-the-box (i.e. no need for the consumer to keep track of state themselves). That way consumers with a specific use case can rely on the lower level `DataTable<X>` components while most use cases can be fulfilled with a simpler API

I feel having a data-table in ui-core alongside the table still has the same issues with the data-table being too specific of a usecase for ui-core. I think I'd suggest a slight alteration of your suggestions: export a data-table from ui-widgets that's been built with the table components from ui-core. That leaves the basic table components as is, and provides the more specific component from ui-widgets.

I think the tricky thing is also that my suggestions are based on an anticipated change in ui-core. It's a bit of a hypothetical discussion at the moment. But then again, you also wouldn't want to have to make changes to ui-core components right after release because we've switched paradigms. That's why I thought I'd mention it.

Mohammer5:
> > OK it seems that despite of the notes discussion and the detailed list of components and their APIs in the PR description (a list which has been available since the creation of this PR), I managed to produce components which vastly differ from people's expectations. I'm going to let all of this sink in a bit, think about where to take this PR and how to find consensus about this.
>
> Sorry about that, I feel that that's a little bit of a thing I've triggered just now. Didn't mean to throw a spanner in the works.

I think we (team platform) just talked about this the last couple days, which is why this comes as a surprise.

---

I like the idea to move this one into `widgets` and then exposing different building blocks from there.
We can still think about what core components the data table could use when we start tackling the (code-)design of the core & widget components

HendrikThePendric:
> Sorry about that, I feel that that's a little bit of a thing I've triggered just now. Didn't mean to throw a spanner in the works.

Yeah, I there was definitely some frustration there. I had put quite a lot of effort into discussing things thoroughly and providing info at an early stage, to avoid have to rework the implementation too much later on. As such I was disappointed that my efforts hadn't paid off. But still, this was just about the process, the content of the review comments was excellent as usual and I am sure we will end up with a good set of components in the end. My frustrations are gone now.

I've had a talk with Joe and another talk with Viktor on where to take this and I will summarise below:
- From a UI perspective the DataTable and Table components should look almost exactly the same. The only difference is that the Table is for static data display, so doesn't need hover styles etc.
- As @ismay suggested, we do also want to start using slightly different approach, where we have have some extremely simple components that only contain style, and some other components with more logic. How we publish these different types of components is not completely clear yet, and @varl is going to organise a meeting about that topic.

So, even though we're not 100% sure yet on how everything should get published, what has become clear is that I need to extract some even lower-level building blocks, and I will proceed to do that after I've addressed the more specific PR comments above.

I expect this is also going to mean that this PR will remain open for quite some time.

packages/core/src/Table/TableCellHead/Action.js
```diff
@@ -15,12 +17,16 @@ export const Action = ({ onClick, children }) => (
                 justify-content: center;
                 flex-shrink: 0;
             }
-            span:hover {
+            button:hover {
```
mediremi, line 20: As the focused button outline has been removed, we need to highlight the fact that a button is focused with the keyboard using the background (otherwise keyboard users won't be able to tell where focus is):

```suggestion
            button:hover, button:focus-visible {
```

HendrikThePendric:
> Would it be possible for the columns used for checkbox, expand-arrow and draggable controls use a smaller `width`? Right now, they are expanding to fill the width like the other cells (as expected). Ideally, these cells will not take up any more room than they need.

@joe I addressed the checkbox width and the width for the chevrons in 3faf963, but note the following:
- The column width for the checkboxes was updated in the story only, because the checkboxes are not part of the `TableCell`, `TableCellHead` or `TableRow` component.
- The `TableRowExpandable` component is a little more high-level, so for that I did update the component itself.

packages/core/src/Table/Table.stories.js
```diff
@@ -17,190 +22,458 @@ Should be used with multiple Table-related child components - see the table and
...
+        </Table>
+        <TableToolbar position="bottom">
+            <p>Content (bottom)</p>
+        </TableToolbar>
+    </>
+)
+export const Toolbars = ToolbarsTemplate.bind({})
+Toolbars.args = {}
+
+const ExpandableContentTemplate = args => {
```
mediremi, line 280: I think this story can be removed as the more interactive `Expandable Table Row` story below showcases the feature better.

ismay:
Ok, I went through the entire discussion again to reacquaint myself with where we were. I haven't gone into the details and specific code yet. There's been a lot of discussion in this thread, it reminds me a little of how things went with the Select. On a more meta level, this is what stands out to me:

- Some of the components strike me as too specific for core. The TableDataCell, TableHeaderCellAction, TableScrollBox and TableToolbar come to mind. I think that, given how differently we all seem to think about what belongs in core/widgets (or elements/composed), it might just be too soon to create abstractions. I think we should probably just hold off on those until we all agree. It's a bit of a moving target at the moment.
- I still wonder if this isn't more at home as a recipe. The components/process remind me a little of how things went with the select/transfer. It seems fairly complex and I think it'd be a sizeable maintenance burden. In hindsight I wish we hadn't created the select and transfer the way we did, and I'm wondering if we shouldn't call that early for the datatable as well.

I hope you don't mind me being open with these concerns @HendrikThePendric, it's not on you and I know how frustrating these long PRs can get. As I said, reminds me a lot of how things went with the select. Anyway, let me know what you think!

HendrikThePendric:
@mediremi @varl I've looked into prop-name-consistency for the scrolling data-table and I _think_ what I did makes sense:
- We have the `TableScrollBox` with `maxHeight` and `maxWidth`. This is explicit because it actually results in a `max-height` and `max-width` CSS attribute being applied to the div. It is also fully consistent with other components such as the `Modal` and the `Select`.
- Then we have the `DataTable`, which has a `scrollHeight` and `scrollWidth` property. These names are not consistent with the other components, but I think that's justifiable since they do something slightly different as well. These props don't set the `max-height` or `max-width` of the actual HTML table element, instead the limit the dimensions of the scroll-box, so only that part of the table is visible.

So I'm keeping things as they are for now.

packages/widgets/src/DataTable/DataTableColumnHeader.js
```diff
@@ -0,0 +1,154 @@
+import { requiredIf } from '@dhis2/prop-types'
+import cx from 'classnames'
+import PropTypes from 'prop-types'
+import React, { forwardRef } from 'react'
+import styles from './DataTableColumnHeader/DataTableColumnHeader.styles.js'
+import { FilterHandle } from './DataTableColumnHeader/FilterHandle.js'
+import { Sorter } from './DataTableColumnHeader/Sorter.js'
+import { TableHeaderCell } from './TableElements/index.js'
+
+const SORT_DIRECTIONS = ['default', 'asc', 'desc']
```
mediremi, line 10: I think this should be imported from `./DataTableColumnHeader/Sorter.js` instead of being defined again.

mediremi:
LGTM :heavy_check_mark: :rocket:

mediremi (approved):

varl (approved):

PR #464, opened 2021-01-20
fix(headerbar): specify HeaderBar /api/me query fields required

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-73

### DIFF
--- a/packages/widgets/src/HeaderBar/Profile.js
+++ b/packages/widgets/src/HeaderBar/Profile.js
@@ -5,7 +5,8 @@ import { joinPath } from './joinPath.js'
 import { ProfileMenu } from './Profile/ProfileMenu.js'
 import { TextIcon } from './TextIcon.js'
 
-const avatarUrl = (avatar, baseUrl) => joinPath(baseUrl, 'api/fileResources', avatar.id, 'data')
+const avatarUrl = (avatar, baseUrl) =>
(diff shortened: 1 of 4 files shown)


packages/widgets/src/HeaderBar/Profile.js
```diff
@@ -49,34 +42,35 @@ export default class Profile extends React.Component {
...
             >
-                {this.userIcon(user, baseUrl)}
+                {this.renderUserIcon({ avatar, name, baseUrl })}
 
                 {this.state.show ? (
                     <ProfileMenu
-                        avatar={avatarPath(user.avatar, baseUrl)}
-                        name={user.name}
-                        email={user.email}
+                        avatarUrl={avatar ? avatarUrl(avatar, baseUrl) : null}
```
HendrikThePendric, line 65: I think I'd prefer to keep the complexity in the function instead of pulling it out.

HendrikThePendric (changes requested):
This popped up in my lists of "review requested" list, but I haven't looked at this in great detail yet because there are 53 failing tests. The code is looking good. I did one minor suggestion. I'm happy to look at the failing tests with you tomorrow.

mediremi:
Sorry about the failing tests I didn't realise cypress had network request fixtures. I've updated the URLs of `/api/me` requests in the cypress tests and now all tests pass.

HendrikThePendric (approved):
Excellent!

PR #465, opened 2021-01-20
fix(header bar): enforce size of 36px in header bar and 48px in profile menu for ImageIcon and TextIcon

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-101


HendrikThePendric (changes requested):
This works of course and the code is looking good. But I do wonder if a slightly different approach wouldn't be more suitable:
- The `ImageIcon` doesn't render the wrapper `div` but only the `img` 
- In the `Profile` and `ProfileHeader` component nest the `ImageIcon` in a `Box` with a `width` and `height` specified.

I would prefer that over adding a new prop + dynamic styles to the `ImageIcon`. Lemme know if you agree.

mediremi:
I agree that sounds much simpler :100:

HendrikThePendric (changes requested):
Thanks for making these changes. I've left one remark in the code for further improvement. However, I realise there are some more things wrong with the `ImageIcon`:

## Stories and cypress tests
As I note in the code-comment, there are no demo or e2e stories about the image icon functionality. These should be added, and I'd argue that's still within the scope of the current JIRA issue. Usually when we find a bug, we first make an e2e story in which we replicate the bug, and then we fix things using that story. This makes it easier for the reviewer and also it guards against regressions. 

Unfortunately, in this case it seems that we don't have any feature files yet that describe the user-icon functionality, so we will first need to introduce that. Probably in the fixture file we need to cover these scenarios:
- Renders a text icon with the user's initials if no profile-picture is present
- Renders an image icon if a profile picture is present
- The image icon in the headerbar is 36px
- The image icon in the profile menu is 48px
 
So you'll need to create an e2e story for these scenarios (I think you just need 2 stories, not 2). Also I think it'd be good to add one demo that showcases the image icon.

## Style not compliant with design-system
The [design-system specs](https://github.com/dhis2/design-system/blob/master/organisms/header-bar.md) for the `HeaderBar` are very minimal but what we do see is that the image should be circular. So probably that would mean a completely different implementation of the `ImageIcon`, likely a bit more [like this](https://stackoverflow.com/a/25685393/1143502). Technically this is out of scope, so it's up to you if you want to fix it now, or otherwise create a new issue for this. It seems like a fairly small fix to me though.

packages/widgets/src/HeaderBar/ImageIcon.js
```diff
@@ -2,21 +2,21 @@ import propTypes from '@dhis2/prop-types'
...
             }
-
-            div {
-                width: 48px;
-                height: 48px;
-            }
         `}</style>
-    </div>
+    </>
 )
```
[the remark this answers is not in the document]
mediremi: Ah I didn't realise you could nest styled jsx like that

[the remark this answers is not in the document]
mediremi: React doesn't seem to like this:
> img is a void element tag and must neither have `children` nor use `dangerouslySetInnerHTML`.

:disappointed:

mediremi:
Apologies for the great increase in diff size :sweat_smile: 

- While adding the `ImageIcon` demo and the icon E2E stories for the HeaderBar I realised that I needed to refactor to allow passing an avatar URL directly
- It turns out that the icon components are the only ones in the header bar that are not keyboard accessible (they cannot be triggered using the `Enter` key) so I had to wrap them in a `<button>`
- The size of the `TextIcon` was fixed at 36px in both the header bar and profile menu, so I had to make the same changes that I did to `ImageIcon` (namely setting `height` and `width` to `100%`) so that it would grow to 48px in the profile menu

packages/widgets/src/HeaderBar/HeaderBar.js
```diff
@@ -76,8 +79,10 @@ export const HeaderBar = ({ appName, className }) => {
                     <Profile
                         name={data.user.name}
                         email={data.user.email}
-                        avatar={data.user.avatar}
-                        baseUrl={baseUrl}
+                        avatarUrl={
+                            testAvatarUrl ||
+                            avatarUrl(data.user.avatar, baseUrl)
+                        }
```
HendrikThePendric, line 85: See first comment. This should not be required

packages/widgets/src/HeaderBar/HeaderBar.stories.e2e.js
```diff
@@ -3,13 +3,27 @@ import { storiesOf } from '@storybook/react'
...
-        config={{
-            baseUrl: 'https://domain.tld/',
-            apiVersion: '',
-        }}
-    >
-        <HeaderBar appName="Example!" />
-    </Provider>
-))
+// TODO: Replace with DHIS logo
+const avatarUrl = `https://placekitten.com/200/300`
```
HendrikThePendric, line 7: IMO using some placeholder image site is fine for this type of thing.

packages/widgets/src/HeaderBar/HeaderBar.js
```diff
@@ -105,4 +110,5 @@ export const HeaderBar = ({ appName, className }) => {
 HeaderBar.propTypes = {
     appName: propTypes.string,
     className: propTypes.string,
+    testAvatarUrl: propTypes.string,
```
HendrikThePendric, line 113: remove

packages/widgets/src/HeaderBar/ImageIcon.js
```diff
@@ -1,30 +1,22 @@
 import propTypes from '@dhis2/prop-types'
 import React from 'react'
 
-export const ImageIcon = ({ src, onClick, dataTestId }) => (
-    <div onClick={onClick} data-test={dataTestId}>
-        <img src={src} alt="user avatar" />
+export const ImageIcon = ({ src, dataTestId }) => (
+    <>
+        <img src={src} alt="user avatar" data-test={dataTestId} />
```
HendrikThePendric, line 6: We use `dataTest`, not `dataTestId` for this type of thing.

HendrikThePendric (changes requested):
I've left some more comments in the code, but lso have a general recommendation:
- We should have one `UserIcon` component 
- It should be rendered in a `Box` that sets its dimensions

Currently it seems like we are doing very similar things twice  in the header-bar and the profile-menu.

amcgee:
Closes DHIS2-9167

HendrikThePendric (approved):
Great work, well done.

PR #473, opened 2021-01-25
fix(headerbar): translate search bar placeholder text

mediremi:
The original intent of this PR was to fix https://jira.dhis2.org/browse/DHIS2-8875, but after discussing with @amcgee we decided that it would make more for the backend to be responsible for translating application names - which it would do by translating the `displayName` property of the data returned by the `getModules` API.

Moving translation to the backend reduces the number of network requests and reduces the amount of effort needed across frontend apps.

However, the translation of the placeholder text of the search bar ('Search apps') is the responsibility of the frontend and this PR fixes a regression in its translation.

### DIFF
--- a/packages/widgets/src/HeaderBar/Apps.js
+++ b/packages/widgets/src/HeaderBar/Apps.js
@@ -1,5 +1,5 @@
 import { useConfig } from '@dhis2/app-runtime'
-import i18n from '@dhis2/d2-i18n'
+import i18n from '../locales'
 import propTypes from '@dhis2/prop-types'
 import { colors, theme } from '@dhis2/ui-constants'
 import { Card } from '@dhis2/ui-core'
--- a/packages/widgets/src/HeaderBar/Apps.js
+++ b/packages/widgets/src/HeaderBar/Apps.js
@@ -1,5 +1,4 @@
 import { useConfig } from '@dhis2/app-runtime'
-import i18n from '../locales'
 import propTypes from '@dhis2/prop-types'
 import { colors, theme } from '@dhis2/ui-constants'
 import { Card } from '@dhis2/ui-core'
@@ -7,6 +6,7 @@ import React, { useState, useEffect, useCallback, useRef } from 'react'
 import css from 'styled-jsx/css'
 import { Settings, Apps as AppsIcon } from '../Icons/index.js'
 import { InputField } from '../InputField/InputField.js'
+import i18n from '../locales'
 import { joinPath } from './joinPath.js'
 
 const appIcon = css.resolve`
--- a/packages/widgets/src/HeaderBar/Apps.js
+++ b/packages/widgets/src/HeaderBar/Apps.js
@@ -1,4 +1,4 @@
-import { useDataMutation, useConfig } from '@dhis2/app-runtime'
+import { useConfig } from '@dhis2/app-runtime'
 import i18n from '@dhis2/d2-i18n'
 import propTypes from '@dhis2/prop-types'
 import { colors, theme } from '@dhis2/ui-constants'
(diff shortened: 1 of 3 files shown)


amcgee:
Hi @mediremi - can you describe the changes made and reasoning for them in the PR description here?  Thanks!

packages/widgets/src/HeaderBar/Apps.js
```diff
@@ -84,194 +83,218 @@ Search.propTypes = {
...
+}
 
-    componentWillUnmount() {
-        document.removeEventListener('click', this.onDocClick)
-    }
+const mutation = {
+    resource: 'i18n',
+    type: 'create',
+    data: ({ items }) => items,
+}
```
amcgee, line 223: Hmmm... logically this is not a mutation, it's a query, the backend just uses a POST endpoint to support a long list of apps.  Ideally we should use a `Query` here and transparently translate it into a POST request in the data engine - the consumer shouldn't have to care about how the data is actually fetched.

I think it's OK to "break the glass" here and pretend it's a mutation, but we'll want to fix it properly soon so if we don't do that now let's create an issue to track it.

packages/widgets/src/HeaderBar/Apps.js
```diff
@@ -84,194 +83,218 @@ Search.propTypes = {
...
+    const [translations, setTranslations] = useState({})
+    const [show, setShow] = useState(false)
+    const [filter, setFilter] = useState('')
+
+    const handleVisibilityToggle = useCallback(() => setShow(!show), [show])
+    const handleFilterChange = useCallback(({ value }) => setFilter(value), [])
+
+    const [mutate] = useDataMutation(mutation, {
+        onComplete: translations => setTranslations(translations),
+    })
```
amcgee, line 235: You shouldn't need a separate `translations` state variable here - the mutation returns data, which can be accessed just like you would for a query.  If this is changed to a query, that becomes even more straightforward.

```js
const [mutate, { loading, error, data: translations }] = useDataMutation(mutation)
```

packages/widgets/src/HeaderBar/Apps.js
```diff
@@ -84,194 +83,218 @@ Search.propTypes = {
...
+            {show ? (
+                <AppMenu
+                    apps={apps.map(app => ({
+                        ...app,
+                        displayName: translations[app.name] || app.displayName,
+                    }))}
+                    filter={filter}
+                    onFilterChange={handleFilterChange}
+                />
+            ) : null}
```
amcgee, line 271: We should probably show a loading state here or within <AppMenu>... this will currently flash the English name before the translations load, which isn't what we want.

packages/widgets/src/HeaderBar/Apps.js
```diff
@@ -84,194 +83,218 @@ Search.propTypes = {
...
+const AppMenu = ({ apps, filter, onFilterChange }) => (
+    <div data-test="headerbar-apps-menu">
+        <Card>
+            <Search value={filter} onChange={onFilterChange} />
+            <List apps={apps} filter={filter} />
+        </Card>
+
+        <style jsx>{`
+            div {
+                z-index: 10000;
```
amcgee, line 203: Why do we need an absolute z-index here?

amcgee: (I see this code was just moved, we should look into this later...)

amcgee (changes requested):
Generally looks good, left some recommendations / comments about the use of a mutation and handling a loading state

amcgee:
> Hi @mediremi - can you describe the changes made and reasoning for them in the PR description here? Thanks!

Pinging this again, as now it's all just a strikethrough paragraph ;-)

If this doesn't fix the linked issue, we need to remove it from the description and update the title.

Remember that these PRs are referenced by the squashed commits and in changelogs, so it's important to make them self-explanatory rather than relying on context or familiarity with the changed code.

mediremi:
@amcgee sorry about that I've now adjusted the PR description to summarise our previous conversation on this PR and its current objectives.

mediremi:
Closing in favour of https://github.com/dhis2/ui/pull/557

PR #474, opened 2021-01-31
refactor: remove context and z-index from Layer

varl:
Introduces a new Portal component that can be used as a standard React
component, either with a default or custom mount point.

BREAKING CHANGE: Removes the Layer.level prop and manual control of
z-index based rendering. The Layer component has a zIndex prop that can
be used to manually control the z-index if needed.

BREAKING CHANGE: Removes LayerContext and hooks for it, e.g.
useLayerContext.

BREAKING CHANGE: Nested Modals in JSX is now an unsupported technique.
Modals can be siblings, but not placed within each other.

If a modal needs to create another modal, then a wrapper component that
controls which of the sibling modals should open and when it should be
used. Modals placed as siblings will render on top of one another based
on the order in the DOM, so last sibling renders on top of the
preceeding ones.

Jira-ticket: LIBS-109


amcgee:
Does this solution support nested `FlyoutMenu` components?  (Haven't looked at the implementation, just thought of another important use-case)

varl:
> Does this solution support nested `FlyoutMenu` components? (Haven't looked at the implementation, just thought of another important use-case)

Yes, I "expanded" the Modal story to contain a few components that need to work, FlyoutMenu among them: https://deploy-preview-474--dhis2-ui.netlify.app/demo/?path=/story/modal--large-modal-with-more-nested-modals

2021-02-01-095214_973x743_scrot

amcgee:
Awesome, thanks - would be great to remove the need for a ui singleton, if possible

ismay:
@varl explained the PR to me in a call. I'll try to summarize what we talked about, as well as some thoughts.

#### Change in approach

The concept of this PR seems slightly different than our current system. If I understand correctly the current system allows z-index values to work as if they're in one global space, without being scoped. If there's a chance of scoping, our current layering will prevent that and ensure the z value still applies as if in a global space. This means that z-index can be controlled with a single hierarchy of z-index values.

Whereas this PR relies on the dom order, without z-index. That's in line with what uber does in their component library: https://baseweb.design/components/layer. It seems simple, and from the tests in this PR seems to work well. What's probably good to take a look at is what they mention as caveats:

>  If you don't wrap the application in a BaseProvider all layers will be added directly to the body element and may create conflicts with other libraries that may add children to body.

I believe in this PR we add children to the body, maybe we should see if that could be a problem for us. They also mention:

> There are many other layer related issues that include focus management, hover management, keyboard navigation, click events and hotkeys usage.

Also good to see whether that could be an issue for us with this approach and whether it all works as well want to. I think their approach is close enough to ours that we could probably learn from how they've approached things.

#### Related changes

So we could see this change as an initial step to a couple of changes:

1. The stacking order of the visual components will become the responsibility of the app, our lib is (as good as) agnostic to stacking (we should also take a look at the useAlert hook in the app platform)
2. The Layer can become just a PageCover (related to the ComponentCover). The PageCover (or AppCover?) can be combined with a Portal to create an overlapping layer in the ui.
3. Move away from context for layering as it seems that's a bit tricky to manage since you must then have a single version of ui in your deps
4. Update the design system with instructions that show how to manage z-index and stacking order in an app. Incorporate what we've learned about it and make sure our recommendations work with our components.

I think we could manage to tackle 1,2 and 3 in this PR. 4 would then be a task that we tackle later on. The challenge seems that as we finish the first three tasks, that we don't completely know what will pop up during 4. Maybe we can keep things minimal for the initial trial, and then hopefully that'll allow us to incorporate anything encountered during 4 and still allow us to move ahead with this PR without having to wait for 4.

 I'm thinking that in 4 we might be able to incorporate the approach I've often seen mentioned in z-index management recommendations, with a local stacking context within components, and potentially using a portal for a globally usable alert layer for example. If that is a useful system, it would be good to not block that approach.

 #### Testing
 
 What @amcgee mentions also made me think: we should make sure that the usecases we want to support are all in the feature files.

varl:
Thanks, @ismay.

> > If you don't wrap the application in a BaseProvider all layers will be added directly to the body element and may create conflicts with other libraries that may add children to body.
> 
> I believe in this PR we add children to the body, maybe we should see if that could be a problem for us.

I have thought about this a bit, and instead of adding children directly to `document.body`, we can create a container `div` and add that the `document.body`. My initial version used that approach from react-portal[1], but they use a hack to make sure it's the same `div`:

```js
const defaultNode = document.createElement('div')

export const Portal = ({ children, node, disable }) => {
    const [mountNode, setMountNode] = useState(null)

    useEffect(() => {
        if (!node && defaultNode) {
            document.body.appendChild(defaultNode)
        }
        setMountNode(node || defaultNode)
    }, [node])

    if (disable) {
        return children
    }

    return mountNode ? createPortal(children, mountNode) : mountNode
}
```

If we use the `defaultNode` as the mount node for the `Portal` component, that puts all _our_ portal:ed components a bit further away from what other libraries does, and is similar in concept to what the `LayerManager` does in baseweb -- without the context.

[1] https://github.com/tajo/react-portal/

HendrikThePendric (commented):
Thanks @ismay for summarising the impact of these changes correctly. It's good to see everyone is aware of the implications of these changes. I talked with Viktor about this as well, and think just relying on DOM order is probably good enough in 99.9% of cases.

Some points I am thinking about:
- To support the other 0.1% of cases where DOM order does not provide the correct stacking order, we could we also expose a `zIndex` prop on the "popper-powered" components so library consumers can brute force a certain stacking effect. Probably we can just keep this out-of-scope for now and simply wait whether we actually need this.
- I think we definitely need a dedicated DOM node to attach our layers-portals too. I don't have a preference on how we should implement this, either via "a hack" or otherwise via some component that we need to wrap the app in, is both fine with me.
- I had some reservations about wether we actually need the `Portal` component. Thinking about it some more, it might be a good place to getOrCreate the portal root node.

So what I think still needs to be done still in this PR is implement something that makes sure we have a dedicated root node.

varl:
Thanks, I'll take a pass at a "special" default node.

HendrikThePendric:
I realise we want to remove context for a reason: we keep running into problems when multiple instance of `@dhis2/ui` are being used, so this seems to be a valid enough reason to want to remove it.

Currently the `Table` and the `Layer` were the only places we were using context and we are removing the use of context in both. We are effectively saying we should just refrain from using context full-stop, because it makes our library a singleton and that's bad.

However, context can be useful to solve particular problem, for example a `Table` exposing some properties to a `TableCell` a few levels down the DOM tree. By restricting ourselves from using it we will, in some cases, be making life more difficult for ourselves.

I wonder why other mainstream UI libraries don't feel the need to restrict themselves in such a way. When doing a [code search in GitHub](https://github.com/search?l=&p=4&q=createContext+repo%3Amui-org%2Fmaterial-ui+repo%3Auber%2Fbaseweb+repo%3Achakra-ui%2Fchakra-ui+repo%3Aant-design%2Fant-design&type=Code) covering a Material-UI, Ant-Design, BaseWeb and Chakra-UI, we get a lot of hits on `createContext`.

My suggestion is that perhaps we are trying to solve this problem in the wrong place. Effectively I have only seen the "multiple-versions-bugs" happening in the dataviz-app, and I believe this is happening because it relies internally on the analytics package. Both packages rely on `@dhis2/ui`. So wouldn't it make more sense to fix the problem there instead?

Having said all of that, I do realise I am not seeing the full picture here. I know that we also need to support "plugins", but I am not familiar with how these work. I suspect the need for our UI-lib to continue working when there are multiple versions installed could be related to that.....

HendrikThePendric:
> I have thought about this a bit, and instead of adding children directly to `document.body`, we can create a container `div` and add that the `document.body`.

@varl I happened to stumble upon [this article](https://atfzl.com/don-t-attach-tooltips-to-document-body) in which the author makes a case for not appending directly to the document body, but a dedicated portal root node. The performance benefits mentioned in the article are pretty substantial, so I'd argue the new layering system should also append to a dedicated portal root node instead of the document body.

varl:
> The performance benefits mentioned in the article are pretty substantial, so I'd argue the new layering system should also append to a dedicated portal root node instead of the document body.

Nice find.

PR #482, opened 2021-02-02
fix: add outline to tab component when focused

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-77

### DIFF
--- a/packages/core/src/Tab/Tab.js
+++ b/packages/core/src/Tab/Tab.js
@@ -82,6 +82,11 @@ const Tab = ({
                 overflow: hidden;
                 text-overflow: ellipsis;
                 transition: fill 150ms ease-in-out;
+                padding: 0px 2px;
+            }
+
+            button:focus span {
+                outline: 1px dotted ${colors.grey900};
             }
 
             button > :global(svg) {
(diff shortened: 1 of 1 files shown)


packages/core/src/Tab/Tab.js
```diff
@@ -82,6 +82,11 @@ const Tab = ({
                 overflow: hidden;
                 text-overflow: ellipsis;
                 transition: fill 150ms ease-in-out;
+                padding: 0px 2px;
```
cooper-joe, line 85: What's the purpose of this padding? Removing it doesn't seem to affect the focus styles, and I'd rather not make aesthetic changes unless we need to.

mediremi: I thought the extra padding was being suggested by the description of https://jira.dhis2.org/browse/LIBS-77 (maybe to make the outline less tight around the text?). I'll remove it :+1:

cooper-joe: Oh yeah, sorry! It was actually me who initially suggested it 😅 I think let's keep it out to avoid the style change 👍

cooper-joe (approved):

PR #483, opened 2021-02-03
docs(transfer): add recipe for handling all loaded options are selected

Mohammer5:


ismay:
I've tagged @deboragaleano. Not sure if this fits with content that you'd like to offer feedback on Debora, but if it does it would be good to hear what you think!

deboragaleano:
Just in terms of organization, I think that maybe this recipe could be moved under the `troubleshooting` section of the sidebar - instead of creating a new section for it. That way we can keep adding new recipes in one place. This organization would be implemented later as well so no need to worry about it now ! 👍 

> I've tagged @deboragaleano. Not sure if this fits with content that you'd like to offer feedback on Debora, but if it does it would be good to hear what you think!

martinkrulltott (approved):

PR #484, opened 2021-02-03
fix: hide tooltips when reference is hidden

KaiVandivier:
This addresses the issue in [this thread](https://github.com/dhis2/notes/discussions/205)

The 'hide' modifier from popper is used to hide tooltips when their reference component is obscured

Some questions:
1. Should the popper hide when the reference component is partially covered (as in the demo above), or only when fully covered? Check out the demo at the bottom of the popper docs page to see the difference between 'escaped' and 'reference-hidden'
2. I think this behavior should be the default for all tooltips, but should it be configurable with props? E.g. set the tooltip to hide on partial cover/full cover/not at all

### DIFF
--- a/packages/core/src/Tooltip/Tooltip.js
+++ b/packages/core/src/Tooltip/Tooltip.js
@@ -11,6 +11,11 @@ const TOOLTIP_OFFSET = 4
 const popperStyle = resolve`
     z-index: ${layers.applicationTop};
     pointer-events: none;
+
+    // Hide popper when reference component is obscured (https://popper.js.org/docs/v2/modifiers/hide/)
+    div[data-popper-escaped="true"] {
+        visibility: hidden;
+    }
 `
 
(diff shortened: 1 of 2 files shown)


varl:
> Should the popper hide when the reference component is partially covered (as in the demo above), or only when fully covered?

Partially covered is fine for now I think? We can change that in a patch release since it's an internal implementation detail.

> I think this behavior should be the default for all tooltips, but should it be configurable with props?

It's easy to add a new `prop`later, but hard to remove so the general strategy is to not include it unless we are sure it is needed.

KaiVandivier:
Added a 'flip' before hiding with the intention of improving behavior when reference element is half-exposed - [see demo here](https://github.com/dhis2/notes/discussions/205#discussioncomment-338802)

HendrikThePendric (approved):
Great work! 🎉 👍

PR #500, opened 2021-02-10
chore: add missing dependencies to ui-icons package

amcgee:
The generated output of the `ui-icons` build requires `react` and `@dhis2/prop-types`, but those packages are missing from package.json which could lead to issues when icons are used.  This PR adds `react` as a peer dependency and `@dhis2/prop-types` as a runtime dependency.


amcgee:
Oops, looks like someone beat me to it 🤷

PR #511, opened 2021-02-22
fix(tooltip): configure open and close delays via prop (LIBS-126)

KaiVandivier:
Allows configuring tooltip open and close delays via `openDelay` and `closeDelay` props to cater to a wider variety of use cases.

Relates to [this question in notes](https://github.com/dhis2/notes/discussions/230)

### DIFF
--- a/packages/core/src/Tooltip/Tooltip.js
+++ b/packages/core/src/Tooltip/Tooltip.js
@@ -1,5 +1,6 @@
 import propTypes from '@dhis2/prop-types'
 import { colors, layers } from '@dhis2/ui-constants'
+import PropTypes from 'prop-types'
 import React, { useRef, useState } from 'react'
 import { createPortal } from 'react-dom'
 import { resolve } from 'styled-jsx/css'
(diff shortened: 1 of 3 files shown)


cooper-joe (approved):
Visual/Design approved 👍 (this is not a code review 😉) 

I'll make a note to add documentation for the delay durations to the design system guidelines.

packages/core/src/Tooltip/Tooltip.js
```diff
@@ -1,5 +1,6 @@
 import propTypes from '@dhis2/prop-types'
 import { colors, layers } from '@dhis2/ui-constants'
+import PropTypes from 'prop-types'
```
HendrikThePendric, line 3: `import propTypes from '@dhis2/prop-types'`
should be enough

HendrikThePendric (approved):
I think this is good to go in as it is, but we could perhaps add some tests to verify the tooltip actually shows and hides at the specified times. I think Jest would be OK for that.

packages/core/src/Tooltip/Tooltip.js
KaiVandivier: I used the react library so I wouldn't have to revisit this once the storybook changes are merged 😅 is it okay to leave it this way?

KaiVandivier:
> I think this is good to go in as it is, but we could perhaps add some tests to verify the tooltip actually shows and hides at the specified times. I think Jest would be OK for that.

Sure, I'll add that

PR #520, opened 2021-02-24
feat(input): add min/max/step attr for inputs (LIBS-133)

KaiVandivier:
Adds support for adding native `min`, `max` and `step` attributes to inputs in `Input` and `InputField`

Relates to this discussion: https://github.com/dhis2/notes/discussions/240

### DIFF
--- a/packages/core/src/Input/Input.js
+++ b/packages/core/src/Input/Input.js
@@ -129,6 +129,9 @@ export class Input extends Component {
             loading,
             value,
             tabIndex,
+            max,
+            min,
+            step,
             dataTest,
         } = this.props
 
@@ -141,6 +144,9 @@ export class Input extends Component {
                     ref={this.inputRef}
                     type={type}
                     value={value}
+                    max={max}
+                    min={min}
+                    step={step}
                     disabled={disabled}
                     readOnly={readOnly}
                     tabIndex={tabIndex}
@@ -196,6 +202,10 @@ Input.defaultProps = {
  * @prop {string} [value]
  * @prop {string} [tabIndex]
  *
+ * @prop {string} [max] The native `max` attribute
+ * @prop {string} [min] The native `min` attribute
+ * @prop {string} [step] The native `step` attribute
+ *
  * @prop {boolean} [disabled]
  * @prop {boolean} [readOnly]
  * @prop {boolean} [dense] - Compact mode
@@ -216,9 +226,15 @@ Input.propTypes = {
     error: sharedPropTypes.statusPropType,
     initialFocus: propTypes.bool,
     loading: propTypes.bool,
+    /** The [native `max` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefmax), for use when `type` is `'number'` */
+    max: propTypes.string,
+    /** The [native `min` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefmin), for use when `type` is `'number'` */
+    min: propTypes.string,
     name: propTypes.string,
     placeholder: propTypes.string,
     readOnly: propTypes.bool,
+    /** The [native `step` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefstep), for use when `type` is `'number'` */
+    step: propTypes.string,
     tabIndex: propTypes.string,
     type: propTypes.oneOf([
         'text',
--- a/packages/core/src/Input/__tests__/Input.test.js
+++ b/packages/core/src/Input/__tests__/Input.test.js
@@ -1,0 +1,13 @@
+import { shallow } from 'enzyme'
+import React from 'react'
+import { Input } from '../Input.js'
+
+describe('<Input>', () => {
+    it('passes min, max, and step props as attributes to the native <input> element', () => {
+        const testProps = { min: '0', max: '10', step: '0.5' }
+        const wrapper = shallow(<Input type="number" {...testProps} />)
+
+        const inputEl = wrapper.find('input')
+        expect(inputEl.props()).toMatchObject(testProps)
+    })
+})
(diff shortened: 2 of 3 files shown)


packages/core/src/Input/Input.js
```diff
@@ -196,6 +202,10 @@ Input.defaultProps = {
  * @prop {string} [value]
  * @prop {string} [tabIndex]
  *
+ * @prop {string} [max] The native `max` attribute
+ * @prop {string} [min] The native `min` attribute
+ * @prop {string} [step] The native `step` attribute
```
ismay, line 207: I see that here we're specifying the type as a string, but the accepted prop-type is a number. I took a look, and it seems that react will convert numbers passed to native html elements to strings (which is what native html elements expect). So both will work, but we should probably follow a single convention (string, number, or string || number).

I'm not sure what our convention is though.

ismay (commented):
Changes look good to me, only thing that I'm not sure about is the accepted type (see comment). Once that's resolved this should be good to merge 👍

ismay:
Ah, and this: I think I'd personally classify this update as a `feat` update, unless I'm missing something it seems like we didn't support this before and it's a new feature. But if I'm misinterpreting that and fix is actually better that's fine with me as well.

One other thing: it's good to specify a scope for all commits. So instead of say `fix: ...`, do `fix(input): ...`. That way the changelog changes will be grouped by topic and easier to read. Also, I'd get rid of the merge commit by rebasing on master, that way you'll have a cleaner history before merging (not sure if that'll have an effect for semantic release, but it seems cleaner to me).

packages/core/src/Input/Input.js
ismay: I talked to Jan-Gerke and he mentioned that we don't have a convention for this yet. Seems like we could just start conservatively and only accepts strings. And then if there's a reason for accepting numbers as well we can always accept that later on.

KaiVandivier: Woops! this was a mistake - I meant to make them all `number` after copying and pasting from definitions above. Thanks for catching it 😉

KaiVandivier:
> Ah, and this: I think I'd personally classify this update as a `feat` update, unless I'm missing something it seems like we didn't support this before and it's a new feature. But if I'm misinterpreting that and fix is actually better that's fine with me as well.

'feat' is fine with me 👍 

> One other thing: it's good to specify a scope for all commits. So instead of say `fix: ...`, do `fix(input): ...`. That way the changelog changes will be grouped by topic and easier to read. 

Ah, that sounds good - I'll do that from now on. I didn't do it for that commit above because it includes Input and InputField, but maybe it would have been better to split them up for scoping purposes.

>Also, I'd get rid of the merge commit by rebasing on master, that way you'll have a cleaner history before merging (not sure if that'll have an effect for semantic release, but it seems cleaner to me).

I'm planning to squash merge because I don't think the individual commits here are important to see on master, does that seem fine?

Thanks @ismay!

packages/core/src/Input/Input.js
KaiVandivier: Oh weird - your second comment there wasn't visible when I posted my comment above, but its timestamp is before 🤔 

I will switch the prop types to strings then

HendrikThePendric: I agree with @ismay and @Mohammer5  that we should (for now) only accept strings. As @ismay says, React will accept numbers but will cast them to strings because that's what DOM element attributes are. That means that if we accept a number on a prop we use to control a DOM element's attribute, and later have to read the value of the value of that attribute, we get a type mismatch. For specifying `min`, `max` and `step` in particular, I don't really expect any problems. However, for something like `value` it can be problematic, for example:
```jsx
const NumberInput = (value, onChange) => <input type="number" value={value} onChange={onChange} />
NumberInput.propTypes = {
    value: PropTypes.number,
    onChange: PropTypes.func,
}
const UncontrolledNumberInput = () => {
    const [value, setValue] = useState(1)
    return <NumberInput value={value} onChange={event => setValue(event.target.value)} />}
}
```
In the example above `value` will first be a `number` with value `2`. However, once the users changes that value, `event.target.value` will be of type `string` and so will the state. We'll now also see a props error on the `UncontrolledNumberInput`. We could of course fix this by relaxing the prop-types a bit, but I really don't like props changing value type across render cycles.

The above example is probably pretty specific to `value`, but since `value` is a DOM attribute, just like `min`, `max`, and `step`, I think we should treat them all the same and only accepting a `string` for DOM attributes seems like the best solution to avoid ambiguity.

Mohammer5: The reason why I argued for strings only, even for number-type `value` values (:sweat_smile:) is:
Our current strategy is to us RFF, which provides the parse/format functionality, so it's something that's not really an issue to dhis2 right now.

3rd party dhis2 apps and other apps would face the same problem, regardless of the prop type definition anyways, as the dom element converts the value to a string anyway, so they'll have to figure out a solution regardless of what we do.

ismay: > The above example is probably pretty specific to value, but since value is a DOM attribute, just like min, max, and step, I think we should treat them all the same and only accepting a string for DOM attributes seems like the best solution to avoid ambiguity.

That sounds like a good reason 👍. Plus we can always add to it later, whereas removing an accepted proptype will be more difficult.

KaiVandivier:
Cleaned up history to two commits with correct scopes ^

packages/core/src/Input/Input.js
KaiVandivier: Strings it is 👍

packages/core/src/Input/__tests__/Input.test.js
```diff
@@ -0,0 +1,13 @@
+import { shallow } from 'enzyme'
+import React from 'react'
+import { Input } from '../Input.js'
+
+describe('<Input>', () => {
+    it('passes min, max, and step props as attributes to the native <input> element', () => {
+        const testProps = { min: 0, max: 10, step: 0.5 }
```
ismay, line 7: I see these are still numbers (they're causing a prop warning here: https://github.com/dhis2/ui/pull/520/checks?check_run_id=1978276580#step:7:1037)

ismay (approved):
Looks good, nice work! I'd recommend a regular merge over a squash merge in this case, since you have two updates in the PR. Would be a shame to let the rebasing go to waste 😋

PR #521, opened 2021-02-25
fix(checkbox): make checkbox accessible

ismay:
The disabling of pointer events on the input interfered with the accessibility of our checkbox. I've followed the recommendations in this article https://www.sarasoueidan.com/blog/inclusively-hiding-and-styling-checkboxes-and-radio-buttons. The strategy is basically to leverage the behaviour of the native checkbox as much as possible, by positioning it on top of the svg, and just using the svg for display purposes. Instead of depending on the label to capture the clicks, we're now allowing the input to behave as it normally does, and just using the svg to replace the look of the checkbox.

It could be improved a lot more btw, but for now I've just fixed the most immediate accessibility issue. Later on we should apply the recommendations from the article in full for both our checkboxes and our radio buttons in my opinion.

Other than that I've:

- Moved the defaults to the defaultprops
- ~~Removed `cursor: pointer` on the label as that's not the browser default (pointer indicates a link according to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor))~~
- I've removed `user-select: none` on the label as well as that's also not a browser default (see [jsfiddle](https://jsfiddle.net/m650y3vd/))

The above three changes could be seen as extraneous to this PR, though I think they're useful. So let me know if you think they should be omitted.

Closes https://jira.dhis2.org/browse/LIBS-136

### DIFF
--- a/packages/core/src/Checkbox/Checkbox.js
+++ b/packages/core/src/Checkbox/Checkbox.js
@@ -65,8 +65,8 @@ class Checkbox extends Component {
 
     render() {
         const {
-            checked = false,
-            indeterminate = false,
+            checked,
+            indeterminate,
             className,
             disabled,
             error,
@@ -128,8 +128,6 @@ class Checkbox extends Component {
                         align-items: center;
                         justify-content: flex-start;
                         cursor: pointer;
-                        pointer-events: all;
-                        user-select: none;
                         color: ${colors.grey900};
                         font-size: 16px;
                         line-height: 20px;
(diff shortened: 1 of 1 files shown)


cooper-joe (approved):
Great to see these fixes, nice work @ismay. Design/UX working as expected 👍

KaiVandivier:
Just to chime in with my two cents, I think it's best to leave the cursor as `cursor: pointer`: 👆 it may not be the browser default, but in the current state of web pages, pointers are used for pretty much everything interactive -- see all the buttons and tabs on this page with a pointer, selectable cursor demos at the top of the MDN page, and the label/checkbox demos on that accessibility blog

ismay:
> Just to chime in with my two cents, I think it's best to leave the cursor as `cursor: pointer`: 👆 it may not be the browser default, but in the current state of web pages, pointers are used for pretty much everything interactive -- see all the buttons and tabs on this page with a pointer, selectable cursor demos at the top of the MDN page, and the label/checkbox demos on that accessibility blog

So you could say that in those examples, the pointer is used as a hint that the element can trigger 'something', instead of indicating just that the element hovered over is a link.

I guess that in the end this is a design consideration. What do you think @cooper-joe?

varl (approved):
Code looks good to me. :+1:

ismay:
I'll remove the cursor change as that's not unanimous, and I'll create a separate issue for it so we can talk about that separately from this PR.

cooper-joe:
> I'll remove the cursor change as that's not unanimous, and I'll create a separate issue for it so we can talk about that separately from this PR (and when we have more time).

Good idea - that's fine with me and my approval still applies.

PR #522, opened 2021-02-25
fix(modal): close modal by clicking above (LIBS-127)

KaiVandivier:
Fixes an issue where modals don't register onClose events when clicked above (or below, for bottom-aligned modals), as seen in this discussion: https://github.com/dhis2/notes/discussions/232

### Diagnosis
CenteredContent attempts to enable 'clicking through' the container component that applies the positioning rules by disabling pointer events on that container, then re-enabling pointer events on the `centered-inner-content` div that will hold the children.  The issue is that any margins or other space around children in that div will have pointer events enabled, and thus will not click through to content behind the CenteredContent as desired (as seen in the modal example in the discussion above)

### Proposed solution

~~Adding an `innerContentStyles` prop to CenteredContent that adds styles to `div.centered-inner-content`, which preserves CenteredContent functionality everywhere it’s currently used.~~

~~I moved margins from the `<aside>` in Modal to the `innerContentStyles` to fix the issue. Note that for the modal, passing ‘pointer-events: none’ to `innerContentStyles` and then using ‘pointer-events: auto’ on the `aside`'s styles would also fix this issue, but I chose to use margins because it’s more intuitive.~~

~~I chose to use a styles object over styled-jsx for `innerContentStyles` because I figured it would be a bit more straightforward for a consumer using a CenteredContent in their project to configure the styles this way, but I'm open to feedback about that choice.~~

### Better solution

Use styled-jsx `resolve` to add padding to the CenteredContent, as suggested below

### DIFF
--- a/packages/core/src/Modal/Modal.js
+++ b/packages/core/src/Modal/Modal.js
@@ -18,6 +18,15 @@ const scrollBoxCard = resolve`
     }
 `
 
+const centeredContent = resolve`
+    .top {
+        padding-top: ${spacers.dp64};
+    }
+    .bottom {
+        padding-bottom: ${spacers.dp64};
+    }
+`
+
 /**
  * @module
  * @param {Modal.PropTypes} props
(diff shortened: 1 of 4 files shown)


varl (changes requested):
Thanks for a good diagnosis, @KaiVandivier.

We don't want to open the door for inline styles. That technique is currently not used at all in UI, and once it gets into the code base it will start to spread and leaves us with two ways to style things. We don't want that.

It is possible to use styled-jsx to create scoped styles in the Modal, and pass them along to the CenteredContent, here is a version of your fix that uses that technique: https://github.com/dhis2/ui/commit/81da3b09f6ee23b0ff657ec9b95fc36008d30506

As far as bug fixes go, I would prefer to keep the change:

- isolated to the Modal, 
- not touch the API for the CenteredContent (or at all),
- and not introduce a new way of handling styles.

I do see the need to control how the CenteredContent is positioned in relation to the top/bottom (and maybe left/right) edges on the screen, but we should tackle that in a different PR because that API needs to be considered and will be a new feature instead of a fix.

KaiVandivier:
Improved by using resolved css on the CenteredComponent to manage spacing as suggested 👍

varl (approved):
Good stuff!

HendrikThePendric (approved):

PR #523, opened 2021-02-25
chore(cli app scripts): update to 5.7.2 and adjust related files

Mohammer5:
- [x] Upgraded the `@dhis2/cli-app-scripts` package to 5.7.0
- [x] Changed `build/[es|cjs]/lib.js` to `./build/[es|cjs]/index.js`
- [x] Added the `exports` field to all workspaces' `package.json
- [x] Added locale files to sideeffects in the widgets and forms workspaces
- [x] Adjusted the storybook's babel config file
- [x] Figure out if the changes to get the icons library to work again are legit

When running `yarn build:icons` on master, the following files are being created:
```
packages/icons/build/es/lib.js
packages/icons/build/cjs/lib.js
```

When switching to this branch, this does not happen anymore until I added an `index.js` to the icons' src folder which imports `./react/index.js`

### DIFF
--- a/packages/icons/package.json
+++ b/packages/icons/package.json
@@ -2,8 +2,12 @@
     "name": "@dhis2/ui-icons",
     "version": "6.5.1",
     "description": "Icons used in the UI libs",
-    "main": "build/cjs/lib.js",
-    "module": "build/es/lib.js",
+    "main": "./build/cjs/index.js",
+    "module": "./build/es/index.js",
+    "exports": {
+        "import": "./build/es/index.js",
+        "require": "./build/cjs/index.js"
+    },
     "author": "Viktor Varland <viktor@dhis2.org>",
     "license": "BSD-3-Clause",
     "private": false,
--- a/packages/icons/package.json
+++ b/packages/icons/package.json
@@ -2,11 +2,11 @@
     "name": "@dhis2/ui-icons",
     "version": "6.5.1",
     "description": "Icons used in the UI libs",
-    "main": "./build/cjs/index.js",
-    "module": "./build/es/index.js",
+    "main": "./build/cjs/react/index.js",
+    "module": "./build/es/react/index.js",
     "exports": {
-        "import": "./build/es/index.js",
-        "require": "./build/cjs/index.js"
+        "import": "./build/es/react/index.js",
+        "require": "./build/cjs/react/index.js"
     },
     "author": "Viktor Varland <viktor@dhis2.org>",
     "license": "BSD-3-Clause",
--- a/.storybook/.babelrc.js
+++ b/.storybook/.babelrc.js
@@ -1,6 +1,0 @@
-const makeBabelConfig = require('@dhis2/cli-app-scripts/config/makeBabelConfig.js')
-
-module.exports = makeBabelConfig({
-    moduleType: 'es',
-    mode: 'development'
-})
--- a/.storybook/.babelrc.js
+++ b/.storybook/.babelrc.js
@@ -1,5 +1,6 @@
-const babelPath = require.resolve('@dhis2/cli-app-scripts/config/app.babel.config.js')
+const makeBabelConfig = require('@dhis2/cli-app-scripts/config/makeBabelConfig.js')
 
-module.exports = {
-    extends: babelPath,
-}
+module.exports = makeBabelConfig({
+    moduleType: 'es',
+    mode: 'development'
+})
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -38,6 +38,10 @@ module.exports = {
     addons: [
         '@storybook/preset-create-react-app',
         '@storybook/addon-essentials',
+        {
+            name: '@storybook/addon-storysource',
+            options: { loaderOptions: { injectDecorator: false } },
+        },
         'storybook-addon-jsx',
         '@storybook/addon-a11y',
     ],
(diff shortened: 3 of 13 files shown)


amcgee (approved):
Code change LGTM, obviously tests need to pass and I haven't tried running this locally

packages/icons/package.json
```diff
@@ -2,8 +2,12 @@
...
     "version": "6.5.0",
     "description": "Icons used in the UI libs",
-    "main": "build/cjs/lib.js",
-    "module": "build/es/lib.js",
+    "main": "./build/cjs/index.js",
+    "module": "./build/es/index.js",
+    "exports": {
+        "import": "./build/es/index.js",
+        "require": "./build/cjs/index.js"
+    },
```
amcgee, line 10: @Mohammer5 I think you just want to update these (and the `entryPoints.lib` in `d2.config.js`) to `<base>/react/index.js` right?

Mohammer5: That produces the following output when building:

```
[build:icons] Build parameters: 
[build:icons]  - Mode : production
[build:icons] [WARNING] Invalid "main" field in package.json, expected "./build/cjs/index.js" (got "./build/cjs/react/index.js") 
[build:icons] [WARNING] Invalid "module" field in package.json, expected "./build/es/index.js" (got "./build/es/react/index.js") 
[build:icons] [WARNING] Invalid "pkg.exports.import" field in package.json, expected "./build/es/index.js" (got "./build/es/react/index.js") 
[build:icons] [WARNING] Invalid "pkg.exports.require" field in package.json, expected "./build/cjs/index.js" (got "./build/cjs/react/index.js") 
```

Mohammer5: Which causes the build command to become stale

varl: There is a bug here in the app-scripts: https://github.com/dhis2/app-platform/blob/master/cli/src/lib/validators/validatePackageExports.js#L42

The `entrypointBasename` at this point is `index.js`, and then when the expected exports `expectedESMExport` and `expectedCJSExport` are constructed, it is assumed that the `entrypointBasename` is always under `build/esm/index.js`.

In this case, the constructed path should be `build/esm/react/index.js` so the logic in app-scripts needs to be adjusted.

varl: Or, we say that there must be a `index.js` in the source directory, and don't support nested entry-points under `<source>/<other>/index.js`.

Mohammer5: I guess the `src/index.js` workaround in this PR should be ok for now, I'll create an issue (if you haven't done it already)

varl: https://github.com/dhis2/app-platform/pull/new/fix-validation-exports

varl:
There is a funky test failure in the Transfer's LeftSide component when it imports the background color. I can resolve it manually, so it's definitely _there_, but for some reason loading the CJS fails in the test.

varl:
Ah, I see:

`./core/src/AlertBar/AlertBar.stories.js: Error: styled-jsx/css: if you are getting this error it means that your `css` tagged template literals were not transpiled.`

Mohammer5:
> There is a funky test failure in the Transfer's LeftSide component when it imports the background color. I can resolve it manually, so it's definitely there, but for some reason loading the CJS fails in the test.

I've seen that one as well. But I can't figure out why it's happening. It's never failed before, the imports seem correct. I was able to get the test to succeed again by renaming that file, run the test and then rename the file back to its original name. The second time the error occurred, I wasn't able to do anything about it (even after deleting all node module folders, clearing jest's and yarn's caches and then installing the packages again)

varl:
The new `app-scripts build` command builds and copies everything from `src` to `build`, including all the test files and stories.

So when we run `app-scripts test` from the root, it finds all the test files in `src` and in `build` and runs all of them. The tests fail on the `build/**/*.test.js` files.

varl:
The test is failing because we now copy all the source files to the build directory, and Jest is very eager when it comes to finding and running tests, so it also runs them from the `build/` directory. The tests don't necessarily work in their transpiled variants, so we get problems.

This fixes the `test` command in `app-scripts`: https://github.com/dhis2/app-platform/pull/521

varl:
There is some strange interactions going on between Storybook's Webpack and Bable setup, where some sources are imported through the entry-points in the `build` folder, and some are imported through the `src` folder. The `build` imports are already transpiled so they work properly, but the `src` imports somehow end up not being transpiled by Storybook's Webpack+Bable combo, even after overriding them.

When I swap out e.g. `import { Button } from './Button.js'` to `import { Button } from '@dhis2/ui'` in `Button.stories.js`, Storybook starts working correctly.

Mohammer5:
> When I swap out e.g. import { Button } from './Button.js' to import { Button } from '@dhis2/ui' in Button.stories.js, Storybook starts working correctly.

In my opinion importing from the built library is the actual correct implementation, because it doesn't require that storybook needs to be aware of any build steps required to get the component to work.
On the other side I can see why it's handy to just import the component via the relative path..
Given the current situation, I think I'd prefer changing the imports in all stories, if you agree (importing from `'@dhis2/...'`)

varl:
@Mohammer5 That brings us to the next step.

Nothing in subpackages is allowed to import `@dhis2/ui`. We can either ESLint ignore in every Storybook file, or move the Storybook stuff outside of `core`.

varl:
Alright, I figured out a way to override the Storybook Babel configuration with our necessary customization.

varl (approved):

PR #528, opened 2021-03-04
docs: fix intersection observer stories (LIBS-137)

KaiVandivier:
Intersection observer stories were broken and throwing a 'rootRef is undefined' error; I'm coming back around to fix it after adding the other changes to storybook.

Also added some description to the IntersectionDetector component and its props.


Mohammer5 (approved):
Any idea why the red box is not showing for the first two stories on the docs page?

KaiVandivier:
> Any idea why the red box is not showing for the first two stories on the docs page?

It's visible to me 🤔 but not for you?
Screen Shot 2021-03-04 at 1 59 06 PM

PR #529, opened 2021-03-04
docs: adjust storybook design

cooper-joe:
This PR adjusts the design of some elements of the storybook documentation:
- adjusts document link color to meet WCAG AAA
- adds visibly different `:visited` state for document links
- adjusts sidebar headings for readability

Note: I followed the directions for using the [storybook CSS escape hatches](https://storybook.js.org/docs/react/configure/theming#css-escape-hatches), but this still required using `!important` on all rules.


varl (approved):
A bit annoying that we need `!important` everywhere but accessibility is important, damn it!

PR #531, opened 2021-03-04
fix(profilemenu): fetch helpPageLink from system settings

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-82


packages/widgets/src/HeaderBar/Profile/ProfileMenu.js
```diff
@@ -105,6 +101,11 @@ const ProfileContents = ({ name, email, avatarUrl, helpUrl }) => {
     )
 }
 
+ProfileContents.defaultProps = {
+    helpUrl:
+        'https://docs.dhis2.org/master/en/user/html/dhis2_user_manual_en.html',
+}
+
```
Mohammer5, line 108: I think this should be in the `Profile` components. In this component is should be a required value.

Mohammer5 (changes requested):

packages/widgets/src/HeaderBar/Profile/ProfileMenu.js
Mohammer5: It probably doesn't make a big difference.. Quite a minor issue.
I'm wondering if it makes sense to just not have a default value for this one and just turn it into a required prop in all components using the help link

Mohammer5 (approved):
On second thought, the change request I made is such minor issue, I don't think it matters for this review (and what I'm proposing is quite opinionated)

packages/widgets/src/HeaderBar/Profile/ProfileMenu.js
mediremi: So should `Profile` provide the default to `ProfileContents` if `helpPageLink` is empty (`<ProfileContents helpUrl={data.help.helpPageLink || 'https://docs.dhis2.org/master/en/user/html/dhis2_user_manual_en.html'} />`) or should we assume that the `helpPageLink` system setting is always defined and not have any default value set anywhere in `ui`?

mediremi:
I'm happy to move/remove the default value of `helpUrl` before merging since you're right that `ProfileContents` isnt't the best place to store defaults.

varl:
I would prefer not to hard code a default URL in UI, and not render the `MenuItem` for the help page if the system setting is undefined.

varl (approved):
I wrote a separate comment with an optional change. Either way, looks good to me.

packages/widgets/src/HeaderBar/Profile/ProfileMenu.js
Mohammer5: I agree with what @varl said:

> I would prefer not to hard code a default URL in UI, and not render the MenuItem for the help page if the system setting is undefined.

Basically making the prop required in all components and just render nothing if the help link couldn't be fetched.
Could be a different PR / issue as Viktor mentioned

PR #541, opened 2021-03-11
fix: update LinearLoader story text

mediremi:


KaiVandivier (approved):
Thanks for catching this! :)

PR #542, opened 2021-03-11
fix: enforce min w and h on UserIcon to keep circle shape on small screens [LIBS-151]

jenniferarnesen:
One of the main analytics features for 2.36 is the responsive dashboard app. The goal is that the app should look good and have a subset of functionality on phone portrait orientation.  This PR is to make the UserIcon also look correct on phone portrait, since it is part of the overall impression of the responsive dashboard app.

Before:
image

After: 
image


HendrikThePendric (approved):
👍

PR #543, opened 2021-03-11
docs: add Migrating page

KaiVandivier:
Adds a page to both the docsite and the storybook that links to information about migrating from previous UI versions


amcgee (approved):
🎉 🌮

PR #545, opened 2021-03-16
fix(modal): add ARIA dialog role to modals

mediremi:
Add the ARIA `dialog` role and `aria-modal` attribute (see [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role) and [here](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html#rps_label)) to modals to improve accessibility.

### DIFF
--- a/packages/core/src/Modal/Modal.js
+++ b/packages/core/src/Modal/Modal.js
@@ -75,6 +75,8 @@ export const Modal = ({
             className={centeredContent.className}
         >
             <aside
+                role="dialog"
+                aria-modal="true"
                 data-test={dataTest}
                 className={cx(className, { small, large })}
             >


KaiVandivier (approved):
Nice :)

For discussion's sake, I'll mention that I noticed other PRs adding `role` as a _prop_ for core components ([Table](https://github.com/dhis2/ui/pull/518) and [Input](https://github.com/dhis2/ui/pull/519)), but I think it makes sense to be proactive and prescriptive with `role` in this situation, as you have done.  (imo I think the same could be done for Table, or at least use a default prop value, but that's another discussion)

KaiVandivier:
You may want to commit this as `fix(modal): ...` to scope it for the changelog 👍

mediremi:
> For discussion's sake, I'll mention that I noticed other PRs adding role as a prop for core components (Table and Input), but I think it makes sense to be proactive and prescriptive with role in this situation, as you have done.

Yep for this component it's hard to argue that the role should be optional since `role="dialog"` and `aria-modal="true"` will always be correct for a modal.

> imo I think the same could be done for Table, or at least use a default prop value, but that's another discussion

+1 for default aria attribute values

PR #546, opened 2021-03-17
fix: data table styles

cooper-joe:
This PR adjusts the styles for the incoming data table component.

### DIFF
--- a/packages/core/src/Table/TableCell/TableCell.styles.js
+++ b/packages/core/src/Table/TableCell/TableCell.styles.js
@@ -26,6 +26,7 @@ export default css`
     td.active {
         background-color: ${colors.white};
         outline: 2px solid ${colors.grey600};
+        outline-offset: -2px;
     }
     td.bordered,
     th.bordered {
--- a/packages/core/src/Table/TableCell/TableCell.styles.js
+++ b/packages/core/src/Table/TableCell/TableCell.styles.js
@@ -24,7 +24,8 @@ export default css`
         border-right: 1px solid ${colors.grey300};
     }
     td.active {
-        background-color: ${colors.blue100};
+        background-color: ${colors.white};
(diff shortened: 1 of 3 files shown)


HendrikThePendric (approved):
Thanks @cooper-joe. I like the general hover and active row colours more than what we had before. The active cell style is also good, and because you are using an outline instead of a border we shouldn't run into problems with border-width either.

However, I do wonder about the "inline-editing" use case a little bit here, because I think this happens:
- The outline is not part of the element and takes up 2px *around* the `td`
- Inline editing will be done using a `Popper` or `Popover` which will use the `td` as a positioning reference element
- By default, the `Popper` or `Popover` will position itself exactly adjacent to the reference element, which means it will fully overlap the outline

Perhaps this is acceptable, in which case all is well, but I just wanted to bring it to your attention.

Also, if it isn't acceptable, this isn't a problem we can't solve: `popper.js` has a notion of `offset` so we can definitely get the `Popper` / `Popover` to show 2px away from the cell. With the `Popper` it's already possible using the `modifiers` prop, but that's a bit low-level. We'd probably want to add a dedicated `offset` prop for this, or introduce a dedicated component, i.e. a `TableCellPopover` or something.

However, if we would choose a different style for an active cell which doesn't introduce visual elements outside of the `td`, then there wouldn't be a problem to solve in the first place.

HendrikThePendric (changes requested):
Oh actually, we can also just add this rule:
```css
    td.active {
        background-color: ${colors.white};
        outline: 2px solid ${colors.grey600};
        outline-offset: -2px; /*ADDED*/
    }
  ```
This will keep your outline style but will keep it contained within the `td`. This prevents the complications for the inline-editing use-case and also shouldn't cause problems obscuring the cell content, since the padding is greater than 2px.

HendrikThePendric (approved):
Thanks!

PR #548, opened 2021-03-19
docs: add pages describing final form and helpers to sb & docsite (LIBS-156)

KaiVandivier:
https://jira.dhis2.org/browse/LIBS-156

Adds some documentation pages describing Final Form features in the UI library, validators, and some helper functions.

The changes affect the storybook and the existing doc site


HendrikThePendric (approved):
I expect the forms part might change substantially in the foreseeable future, but this is a very good guide about the current implementation.

KaiVandivier:
Good to know about the changes 🙂  I'll keep an eye out for them!

PR #549, opened 2021-03-22
feat: introduce ui-elements package

varl:
Important: note that this deprecates ui-core but does not remove it. UI
Elements are also not included in the @dhis2/ui package nor in
Storybook.

## `@dhis2/ui-elements`

Worth exploring: https://github.com/dhis2/ui/tree/platonic-components/packages/elements/src

Adds the UI Elements package with an incomplete set of tests that are
intended to ensure that all the Element components comply to the
requirements.

Implements the Button using the UI Elements principles from the README.
This is the proposal of how an UI Element should be structured.

UI Elements will contain more components than exist in the Design
System, as we will need to divide each component into a set of UI
Elements.

We expect to have many UI Elements as we need to ensure that all our
higher level components (and more) can be composed using the UI
Elements.

Supersedes `@dhis2/ui-core`.

## Additional packages

```
src/packages
├── composes
├── constants
├── elements
├── forms
├── hooks
├── icons
├── layouts
└── ui
```

Among these, `composes`, `layouts`, `hooks` are new.

-   `@dhis2/ui-composes` supersedes widgets, as it covers all components that are composed. The antonym package for `elements`. The idea is that these components should be re-composable from another context, meaning that everything that is used to build these components should be available from e.g. `elements`, and `hooks`.
-    `@dhis2/ui-layouts` has been a long time coming (#27), but not completed yet.
-   `@dhis2/ui-hooks` to gather up the hooks we use for our composed components to build behavior.

## `@dhis2/ui` export structure

It is getting crowded so we may want to do something about the way we re-export everything, while still ensuring that consumers has one dependency in their project. The way we have it now, we cannot have multiple components with the same name. A blessing and a curse. This PR does nothing with this point; I figured it was worth writing down.

## Name conventions

To provide better support for automated tools, wider arrange of
developer environments, and be predictable, the name conventions
changes.

PascalCasing and camelCasing are reserved for source code use. The file
system should be kebab-case.

In the future, conventions will be enforced through linting.

These are the conventions applied:

-   directories should always be kebab-case

-   a component is organized into a subfolder with it's name:
    src/{component}

-   component files should be kebab-case with the suffix:
    src/{component}/{component}.component.js

-   component styles should have the suffix:
    src/{component}/{component}.styles.js

-   components stories should have the suffix:
    src/{component}/{component}.stories.js

-   component test files should have the suffix:
    src/{component}/{component}.spec.js

-   in case of multiple specs, they can be in subdirectory for the
    component named:
    src/{component}/tests

    -   test files under tests/ do not need the component as prefix

-   component feature files go in a subdirectory of the component:
    src/{component}/features

## TODO

-   [ ] Get Storybook working again.
-   [ ] Make an example `ui-composes` component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,15 @@ All components can be imported directly from `@dhis2/ui` like so:
 import { Button } from '@dhis2/ui'
 ```
 
-We recommend that you use `@dhis2/ui` as the entrypoint for all imports of our frontend components (as in the example above). That way your imports won't break if any of the underlying packages change.
+We recommend that you use `@dhis2/ui` as the entrypoint for all imports
+of our frontend components (as in the example above). That way your
+imports won't break if any of the underlying packages change.
 
 ## Documentation
 
-`@dhis2/ui` is based on the specifications in our design-system: https://github.com/dhis2/design-system. See the documentation there for more information.
+`@dhis2/ui` is based on the specifications in our design-system:
+https://github.com/dhis2/design-system. See the documentation there for
+more information.
 
 -   Docs: [ui.dhis2.nu](https://ui.dhis2.nu)
 -   Live demo: [ui.dhis2.nu/demo](https://ui.dhis2.nu/demo)
@@ -41,25 +45,53 @@ We recommend that you use `@dhis2/ui` as the entrypoint for all imports of our f
 
 ### `@dhis2/ui-constants`
 
-This package provides access to shared constants, such as colors, spacers and elevation values. They are used across our frontend components and can be used directly in applications as well. Our constants can be imported like so:
+This package provides access to shared constants, such as colors,
+spacers and elevation values. They are used across our frontend
+components and can be used directly in applications as well. Our
+constants can be imported like so:
 
 ```js
 import { colors } from '@dhis2/ui'
 ```
 
-See our [api docs](https://ui.dhis2.nu/#/api) for a full list of the available constants.
+See our [api docs](https://ui.dhis2.nu/#/api) for a full list of the
+available constants.
+
+### `@dhis2/ui-elements`
+
+Think of these components as styled HTML elements. These are the
+elements that are used to compose other UI components.
+
+-   Elements MUST NOT import other elements.
+-   Elements SHOULD NOT provide behavior.
+-   Elements MUST provide relevant styles for the design system.
+-   Elements MUST forward the `ref`.
+-   Elements MAY provide props for specific variants.
+-   Elements SHOULD spread props to the underlying HTML element.
+
+Since UI Elements are useful in other contexts to compose new components
+with specific behavior, they provide a stable base to build on with a
+flexible API.
 
 ### `@dhis2/ui-icons`
 
-This package provides a collection of icons as react components. For tree shaking purposes the icon name, variant and size are all expressed in the component name. Our icons can be imported like so:
+This package provides a collection of icons as react components. For
+tree shaking purposes the icon name, variant and size are all expressed
+in the component name. Our icons can be imported like so:
 
 ```js
 import { IconApps16 } from '@dhis2/ui'
 ```
 
-For a list of all the available icons see [the ui-icons package](https://github.com/dhis2/ui/tree/master/packages/icons/src). Note that during their transformation to React components the svg filenames are PascalCased and prefixed with `Icon`. So `apps-16.svg` becomes `IconApps16` and can then be imported as in the example above.
+For a list of all the available icons see [the ui-icons
+package](https://github.com/dhis2/ui/tree/master/packages/icons/src).
+Note that during their transformation to React components the svg
+filenames are PascalCased and prefixed with `Icon`. So `apps-16.svg`
+becomes `IconApps16` and can then be imported as in the example above.
 
-The default fill of our icons is inherited from `color` with `currentColor`. To set a custom icon color you can use the `color` prop like so:
+The default fill of our icons is inherited from `color` with
+`currentColor`. To set a custom icon color you can use the `color` prop
+like so:
 
 ```jsx
 <IconApps16 color="#DE683D" />
@@ -67,33 +99,55 @@ The default fill of our icons is inherited from `color` with `currentColor`. To 
 
 ### `@dhis2/ui-core`
 
-This package provides the low level building blocks for all our components. We use them as building blocks for our more elaborate components and they can be used in applications for the same purpose. Any of our core components can be imported in the following manner:
+> :warning: DEPRECATED:
+
+This package provides the low level building blocks for all our
+components. We use them as building blocks for our more elaborate
+components and they can be used in applications for the same purpose.
+Any of our core components can be imported in the following manner:
 
 ```js
 import { Button } from '@dhis2/ui'
 ```
 
-See our [api docs](https://ui.dhis2.nu/#/api) or the [live docs](https://ui.dhis2.nu/demo) for a full list of the available core components.
+See our [api docs](https://ui.dhis2.nu/#/api) or the [live
+docs](https://ui.dhis2.nu/demo) for a full list of the available core
+components.
 
 ### `@dhis2/ui-widgets`
 
-This package provides widgets, more elaborate components that are often composed of multiple core components. Components in this package can include translations, interact with an API or address a DHIS 2 specific usecase. An example of a widget import:
+> :warning: DEPRECATED:
+
+This package provides widgets, more elaborate components that are often
+composed of multiple core components. Components in this package can
+include translations, interact with an API or address a DHIS 2 specific
+usecase. An example of a widget import:
 
 ```js
 import { Transfer } from '@dhis2/ui'
 ```
 
-See our [api docs](https://ui.dhis2.nu/#/api) or the [live docs](https://ui.dhis2.nu/demo) for a full list of the available widgets.
+See our [api docs](https://ui.dhis2.nu/#/api) or the [live
+docs](https://ui.dhis2.nu/demo) for a full list of the available
+widgets.
 
 ### `@dhis2/ui-forms`
 
-This package provides several components that allow for easy integration of our form components with [react-final-form](https://github.com/final-form/react-final-form). Besides form components, we also export several validator functions for common usecases. Components from this library can be imported like so:
+This package provides several components that allow for easy integration
+of our form components with
+[react-final-form](https://github.com/final-form/react-final-form).
+Besides form components, we also export several validator functions for
+common usecases. Components from this library can be imported like so:
 
 ```js
 import { TextAreaFieldFF } from '@dhis2/ui'
 ```
 
-The `FF` suffix ensures that these components don't clash with our regular field components from the widgets package and is an abbreviation of `final-form`. See our [api docs](https://ui.dhis2.nu/#/api) or the [live docs](https://ui.dhis2.nu/demo) for a full list of the available components and validators.
+The `FF` suffix ensures that these components don't clash with our
+regular field components from the widgets package and is an abbreviation
+of `final-form`. See our [api docs](https://ui.dhis2.nu/#/api) or the
+[live docs](https://ui.dhis2.nu/demo) for a full list of the available
+components and validators.
 
 ## Reporting an issue or opening a PR
 
--- a/packages/elements/src/button/button.spec.js
+++ b/packages/elements/src/button/button.spec.js
@@ -1,0 +1,27 @@
+import { mount } from 'enzyme'
+import { Button } from './button.component.js'
+
+describe('<Button>', () => {
+    test('has class "toggled" if toggled is true', () => {
+        const wrapper = mount(<Button toggled />)
+
+        const actual = wrapper.find('button')
+        expect(actual.hasClass('toggled')).toBe(true)
+    })
+
+    test('no class "toggled" if toggled is false', () => {
+        const wrapper = mount(<Button />)
+
+        const actual = wrapper.find('button')
+        expect(actual.hasClass('toggled')).toBe(false)
+    })
+
+    test('the built-in classes are in addition to the custom one', () => {
+        const wrapper = mount(<Button primary large className="foobar" />)
+
+        const actual = wrapper.find('button')
+        expect(actual.hasClass('foobar')).toBe(true)
+        expect(actual.hasClass('primary')).toBe(true)
+        expect(actual.hasClass('large')).toBe(true)
+    })
+})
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -41,59 +45,109 @@ We recommend that you use `@dhis2/ui` as the entrypoint for all imports of our f
...
+-   Elements MUST NOT import other elements.
+-   Elements SHOULD NOT provide behavior.
+-   Elements MUST provide relevant styles for the design system.
+-   Elements MUST forward the `ref`.
+-   Elements MAY provide props for specific variants.
+-   Elements SHOULD spread props to the underlying HTML element.
+
+Since UI Elements are useful in other contexts to compose new components
+with specific behavior, they provide a stable base to build on with a
+flexible API.
```
varl, line 74: :point_up: This is an attempt at defining UI elements using keywords as defined by: https://tools.ietf.org/html/rfc2119

We might be able to strengthen the `MAY` to `SHOULD` and the `SHOULD`s to `MUST`s.

ismay (commented):
Looks great. I need to consider the ramifications of the proposal a little more thoroughly, but my initial reaction is that I think this'll be a good direction to take the lib in. More detailed:

- I think the suggestions you've made for the file structure are solid. Since filesystems across operating systems differ in how they treat casing for the filesystem, eliminating casing from our naming strategy seems very good to me. I would love to use this convention for apps as well. One addition I wouldn't mind is limiting depth of the folder tree as well.
- I like the naming (elements and composes). To me the names are very clear and immediately reflect the criterium for what goes where. Simple and easy to understand.
- I think it's good that we export from a single namespace. Though it does limit the names we can use, it allows us to move components internally between packages, which is very useful I think. Plus, identical names in different packages would only be confusing I think.
- I think there are quite a few packages now. Having a timeline on when we want to remove the deprecated libs might be good. Wouldn't want ui-core lingering, or becoming some sort of parallel lib.
- Why not export elements? I would find it super useful to have access to styled primitives in an app. Plus it fits with our philosophy of allowing users to compose components themselves for usecases that are too specific for ui.

Just some initial thoughts. I'll leave a couple comments on specific points and think it over some more.

README.md
```diff
@@ -41,59 +45,109 @@ We recommend that you use `@dhis2/ui` as the entrypoint for all imports of our f
...
+available constants.
+
+### `@dhis2/ui-elements`
+
+Think of these components as styled HTML elements. These are the
+elements that are used to compose other UI components.
+
+-   Elements MUST NOT import other elements.
+-   Elements SHOULD NOT provide behavior.
+-   Elements MUST provide relevant styles for the design system.
```
ismay, line 67: This seems a little unclear to me. I think I'd find something like `Elements MUST provide relevant style primitives for the design system.` more clear. Though that is still open to interpretation: what is relevant? What is the definition of a style primitive (you could also call it an element, or a styled element, but still)?

ismay: The other rules all seem a lot more concrete to me, less open to interpretation. Would be good if we could whittle this rule down to that level as well.

varl:
> * I think there are quite a few packages now. Having a timeline on when we want to remove the deprecated libs might be good. Wouldn't want ui-core lingering, or becoming some sort of parallel lib.

I should have mentioned this. The timeline here is that we would merge this proposal to `alpha`, and work on removing everything from core and widgets and refactoring them to be inline with the requirements.

We merge to `next` once core and widgets have been fully removed, at this point all the docs might not be up to date.

We merge to `master` once we have written codemods to help transition over existing code to the new UI structure and have written migration guides.

> * Why not export elements? I would find it super useful to have access to styled primitives in an app. Plus it fits with our philosophy of allowing users to compose components themselves for usecases that are too specific for ui.

That's a mistake, elements are absolutely meant to be exported !

packages/elements/src/button/button.spec.js
```diff
@@ -0,0 +1,27 @@
+import { mount } from 'enzyme'
+import { Button } from './button.component.js'
+
+describe('<Button>', () => {
```
ismay, line 4: It would be great if we could have visual regression tests for elements, since their main purpose is style. Otherwise it's hard to do anything but test implementation details, which is more labour intensive and a little more fragile.

Maybe until we have those we should just test the hard requirements, like passing a ref, spreading props, and omit testing styles.

packages/elements/src/button/button.styles.js
```diff
@@ -0,0 +1,252 @@
+import { colors, theme, spacers } from '@dhis2/ui-constants'
+import css from 'styled-jsx/css'
+
+export default css`
+    button {
```
ismay, line 5: The button styles are maybe not as indicative of the rest of our lib, but I do think they demonstrate a potential issue. When you lint these styles with a css linter, a couple issues pop up, like the descending specificity in the selectors (which you don't want), and other issues. Overall I also think the styles are too complex and elaborate for a style primitive. Two things come to mind:

- Since elements are going to be mainly style building blocks, and those styles are going to be expressed in css, I think we should enable linting for the css used in the elements. That seems like a critical first line of defense.
- It would be good if we could somehow limit the complexity of the styles for style primitives. I think that's a potential pitfall. We're already limiting the complexity for elements by only allowing styles, but I can see things going off the rails with numerous overly complex and hard to debug styles as well. I can imagine there's going to be some pressure on elements to become too complex stylewise, and I think it'd be good to have something in place that'll help us resist that pressure.

packages/elements/src/button/button.spec.js
mediremi: @ismay :100: since these components exist for styling, visual regression would be the most efficient way of testing them

packages/elements/src/button/button.styles.js
varl: I agree that these styles are pretty heavy for a button. Linting would be a nice heads-up that we are doing something strange in CSS just as it is otherwise.

packages/elements/src/button/button.spec.js
ismay: https://www.chromatic.com/pricing it does say "free for open source" here. It's a testing service (that includes visual tests) created by the storybook team. I do think that that'd be a good fit for ui.

HendrikThePendric (commented):
I think the new structure is a probably an improvement over what we had since the distinction `elements` `composed` and `hooks` is very clear. This should help settle the _what-goes-where_ discussion and prevent the same discussion from resurfacing over and over again.

There are also some things in this PR (and mainly the PR description) I would like to get some more info about and/or have some comments on:

### Extracting behaviour into hooks in `dhis2/hooks` package
At its core, this seems like a logical thing to do and in theory this should make our UI libraries a lot more composable. However, this approach would represent a pretty big shift in how we do things. I personally don't have any experience organising a UI library this way and feel like we're just jumping in at the deep end with this approach. I would probably feel a bit better if we did some more research into this approach first before deciding this is the way forward.

Also, I guess we must ensure that none of our hooks use `useContext()`, because otherwise we'd re-introduce that singleton issue?

### Name conventions
I think the suggestions there make perfect sense, but I am missing one detail: _Where do we put the test stories?_ I think  they should probably be in the root of the `features` dir.

### Export structure
I agree it was worth mentioning this topic, and I think this topic might even become more relevant in our new package-structure, especially for the `elements` and `hooks`. I've been working on a lot of table-related components and have been thinking about how to extract the `elements` from what I have currently produced. I am anticipating to produce a vast number of components, which I will probably put in the `Table` namespace (i.e. use `Table` prefix). This is going result in a whole bunch of components exported from our main entry point, with very long names, and most/all of them only make sense when being used as a child of a table-related component (i.e. `TableHeaderCellSorter`). In some way it'd feel a lot more natural to import from `@dhis2/table` in these type of cases. 

Just mentioning this here, even though it's not really relating to any of the code changes.

### Event callback function signature
In our current codebase we (should) have consistently adopted a custom function signature for callbacks, for example, like this:
```js
const changeHandler = event => {
    props.onChange(
        {
            value: event.target.value,
            name: props.name,
        },
        event
    )
}
```
The decision to do so came about around the time when we were introducing our selects and transfer components. I believe the reason for taking this approach was rooted in the fact that our selects and transfers weren't backed by actual form tags, so there wasn't really a native event.

I believe we are pretty consistent with our `callback(payload, event)` pattern now, and I think this would fall apart once we switch to the `elements` extraction level.

This is a pretty small practical point, but I think we need to consider it before we proceed to implement things a suggested in the PR.

### Default strings with `i18n`
In this new structure, I suppose we would be adding default strings to the `composed` package? Or will we use hooks for those too?

HendrikThePendric:
🤔  🤯  💥 
This idea is probably a completely bonkers, but we could also completely forget about even trying to find any logical distinction between our packages, and instead just organise everything by component:
```js
import {
    SomeComponentWithBehaviour,
    SomeComponentBuildingBlockA,
    SomeComponentBuildingBlockB,
    useSomeComponentHook,
} from '@dhis2/ui/some-component'
```
I'm not even remotely convinced this is even a viable option, but thought it might be good to "think outside the box" for a bit.

HendrikThePendric:
Another crazy idea💡 I started considering:
The `@dhis2/ui-forms` package does very little and is a bit of an odd duckling. When this package was created it was doing a little bit more, but as time went on, logic from these form components was extracted out and moved into widget components (both composition and default strings). At that point we still had some form components like `CheckboxGroup` which implemented some custom logic, but later on we decided to remove these `XGroup` components too and just keep the singular elements (i.e. just Checkbox, not CheckboxGroup).

I think at this point it's fair to say that the components in `ui-forms` do very little. Besides that we have a bunch of useful validator functions. However, these aren't doing anything special, there must be a lot of NPM packages out there containing more validation functions, and have even better coverage.

So what I am thinking now is:
- Remove the `forms` package 
- Offer one or two hooks in our `hooks` package that help binding our form components with React Final Form (or another form library when we decide we want to move to Formik or whatever).
- Recommend a validation library (for example [Yup](https://github.com/jquense/yup) seems quite nice)

packages/elements/src/index.spec.js
```diff
@@ -0,0 +1,50 @@
...
+                expect(button.props().foo).toEqual('bar')
+                expect(button.props().fuz.buz).toEqual('yikes')
+            })
+
+            test('custom class', () => {
+                const wrapper = mount(<Component className="foobar" />)
+                const actual = wrapper.find('button')
+
+                expect(actual.hasClass('foobar')).toBe(true)
+            })
```
KaiVandivier, line 39: This is probably an artefact of copy & paste, but `wrapper.find('button')` won't work for all the elements

mediremi:
For reference/inspiration, here's some other React component libraries

- [Material UI](https://material.io) — [GH](https://github.com/material-components/material-components-web-react/tree/master/packages)
- [Base Web](baseweb.design) — [GH](https://github.com/uber/baseweb/tree/master/src)
- [Chakra UI](https://chakra-ui.com/) — [GH](https://github.com/chakra-ui/chakra-ui/tree/main/packages)
- [Ant Design](https://ant.design/) — [GH](https://github.com/ant-design/ant-design/tree/master/components)

Mohammer5:
> `ui-elements`, `ui-layouts`, `ui-composed`

:heart:

HendrikThePendric:
Thanks for sharing those links @mediremi. Always good to check out the competition 😉 

What particularly stood out to me that none of these libraries attempted to structure their components at all. It's just a bunch of folders with components. Some grouped together because that makes sense, others just exported individually. But no conceptual split according to atomic design or anything.

When I first suggested https://github.com/dhis2/ui/pull/549#issuecomment-804033379, I saw it more as a crazy idea, but after doing some research into what other libs are doing, I am warming to it....

ismay:
> 🤔 🤯 💥
> This idea is probably a completely bonkers, but we could also completely forget about even trying to find any logical distinction between our packages, and instead just organise everything by component:
> 
> ```js
> import {
>     SomeComponentWithBehaviour,
>     SomeComponentBuildingBlockA,
>     SomeComponentBuildingBlockB,
>     useSomeComponentHook,
> } from '@dhis2/ui/some-component'
> ```
> 
> I'm not even remotely convinced this is even a viable option, but thought it might be good to "think outside the box" for a bit.

I think I'd be open to a slight modification of this suggestion. Instead of the deep import you're suggesting here, I wouldn't mind moving everything to a single package (@dhis2/ui). To our users it's already a single package, practically. And for us, internally we could still have whatever kind of organisation makes sense to us.

But maybe that's an unrelated discussion. Basically, I'm suggesting to move away from the monorepo and towards just a single package. To me that seems simpler. Since our libs are already so coupled, I think the added complexity of managing it as a monorepo seems like overhead without a lot of added benefit. But again, probably a separate discussion.

mediremi:
> Since the libs in ui are already so coupled, I think the added complexity of managing it as a monorepo seems like overhead without a lot of added benefit. But again, probably a separate discussion.

I agree :+1: (and it'd be nice to remove the styled-jsx and no i18n restrictions too)

HendrikThePendric:
> [...] moving everything to a single package (@dhis2/ui). [...] But maybe that's an unrelated discussion. [...]

In my head this is related, because this PR describes not only how we should structure our components internally (i.e. having elements that spread props etc.), but also how we split our components into packages.

ismay:
> I think I'd be open to a slight modification of this suggestion. Instead of the deep import you're suggesting here, I wouldn't mind moving everything to a single package (@dhis2/ui). To our users it's already a single package, practically. And for us, internally we could still have whatever kind of organisation makes sense to us.
> 
> But maybe that's an unrelated discussion. Basically, I'm suggesting to move away from the monorepo and towards just a single package. To me that seems simpler. Since the libs in ui are already so coupled, I think the added complexity of managing it as a monorepo seems like overhead without a lot of added benefit. But again, probably a separate discussion.

To be clear: in the comment above I'm not referring to the internal organisation of the lib. Internally I'm still in favor of the organisation proposed in this PR. I was just commenting on how I don't really see the added benefit of exposing that organisation externally.

HendrikThePendric:
Upon reflection, I don't think my last comments were very helpful, but rather a bit disruptive. Having these separate packages (or even just separate folders in a single package as Ismay suggests) has helped us in the past to produce a very composable set of low-level components, so we shouldn't get rid of that. Apologies for my "crisis of faith" @varl 😊 

To summarise:
- I am in favour of dividing our components into `elements` and `composed` instead of `core` and `widgets`
- I don't have an opinion about the monorepo vs just use folders discussion
- The points I raised in my initial comment https://github.com/dhis2/ui/pull/549#pullrequestreview-617360640  are still valid
- In my Table/DataTable PR I'm attempting to apply the split suggested in this PR as an experiment. I will report when I bump into anything noteworthy.

varl:
Thanks for all the input. I think the meeting https://github.com/dhis2/notes/pull/275 marks this as "cannot merge as is", but I'll leave the branch intact and we'll figure out a less aggressive way forward.

PR #554, opened 2021-03-24
chore: merge alpha to master

ismay:
The `merge-alpha-to-master` branch has been branched off my local master, which I merged alpha into (after solving the merge conflicts). Normally I'd just push my master, but that's not possible due to our required status checks on master.

So this seems like the next best thing. I'm not sure if this will mess with history though (for semantic release), so I've requested a review from @varl, to double check if this is our intended way to merge alpha into master.

p.s.: These changes have already been reviewed when they went into alpha (though technically the merge conflict resolutions haven't).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "root",
-    "version": "6.5.7",
+    "version": "6.6.0-alpha.1",
     "repository": {
         "type": "git",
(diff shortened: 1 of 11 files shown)


varl (approved):
While not the smoothest way to merge alpha to master, I do believe that it will work.

ismay:
Noice, yeah seems to have worked. 🎉

PR #555, opened 2021-03-24
fix: don't set the global locale in the headerbar

mediremi:
[The app shell sets the global locale before rendering the application](https://github.com/dhis2/app-platform/blob/98a7a567d0732a421df19a97faa03f40192b678b/adapter/src/components/AuthBoundary.js#L16-L18) and so setting it in the headerbar should be redundant.


varl:
There may be cases where the header bar is used in applications without app-platform, so this should be a breaking change.

mediremi:
@varl I've added a `BREAKING CHANGE:` note in the commit footer.

amcgee:
I'm not sure we should do this yet... I think it will have ramifications we haven't fully considered.

varl:
Alright. Let's track this in https://jira.dhis2.org/browse/LIBS-180 and close this PR for now.

PR #556, opened 2021-03-24
docs: add to maintainers page (LIBS-146)

KaiVandivier:
* Adds a table of contents to navigate this quite-large document
* Adds information about writing stories in MDX, including using a code mod to convert an existing stories file
* Adds information about customizing the source code transformer function that makes source code snippets of stories
* Adds some more information about customizing arg types


docs/for-maintainers.md
```diff
@@ -0,0 +1,423 @@
...
+### Making a reusable story template
+
+To keep stories concise and DRY, you can reuse a story template (that uses the `args` syntax) and set the relevant props on each story using `StoryName.args = { ... }`. That looks like this:
+
+```js
+// The story template
+const Template = args => <Button {...args} />
+
+// Stories to render
+export const Primary = Template.bind({}) // Makes a copy of the template
```
mediremi, line 137: Since arrow functions lexically bind `this`, `someArrowFn.bind({})` is equivalent to `someArrowFn.bind()` and so we don't need to pass an object as `this` when binding a template.

(I should have brought this up earlier when we were porting stories to the new storybook format :sweat_smile:)

KaiVandivier: Ah good to know 😁 I just copied the storybook docs 😅 

After thinking about it a bit, since they're equivalent, I think there's an argument to be made for keeping `.bind({})` how it is so it's the same as the storybook docs (which are linked to a lot in this 'For Maintainers' page) so people don't wonder about the difference as they switch between SB's docs and our docs - does that sound okay to you? 😁  Maybe once everyone adjusts to the new story format we can remove the `{}` library-wide 🙂

mediremi: I've been thinking about why the Storybook team may have used `bind` in this way, and I think that it might be due to typescript: its [type definition for `bind`](https://github.com/microsoft/TypeScript/blob/16031bc429305e5daabf263f208678a6729a161e/lib/lib.es5.d.ts#L293) specifies one required argument ([code example with error](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMOcAUBKAXIgylAJxjAHNEAfRcAEwFNhjbrEBvAKEU8QNqhAKQAiFHEFsAvmzYQEAZyiIAHogC8yVADoARsWqYgA)). Since [there's prior discussion about us moving to typescript for more apps](https://github.com/dhis2/notes/issues/33) it makes sense to keep `.bind({})` even once everyone adjusts to CSF.

So yeah let's stick to how things are now to not only be consistent with the storybook docs and avoid confusion but also because it makes a potential switch to typescript less painful :+1:

mediremi (approved):
This is really useful, especially the edge cases and workarounds section - since we don't often make changes to the stories it's good to have a quick references guide :smile:

docs/for-maintainers.md
KaiVandivier: Ah interesting, it's nice that this supports typescript :) thanks for taking a look!

PR #557, opened 2021-03-24
refactor(headerbar): convert Apps to functional component

mediremi:
This PR is a cherry pick of commit https://github.com/dhis2/ui/pull/473/commits/c9ba64da8c4fd98aa25b380d0525b4f39f08a1c2 from https://github.com/dhis2/ui/pull/473.

The other changes in https://github.com/dhis2/ui/pull/473 are no longer required thanks to some backend fixes.


HendrikThePendric (approved):
There's nothing wrong with these changes, so I'm approving.

Personally I would have probably prioritised other changes over the ones just implemented. The HeaderBar predates quite a few of our ui-core components and as such uses some custom components which could/should be swapped out by ui-core components. I am mainly thinking of the wrapper div which should be implemented using a `Popover` or `Popper` (with `Layer`), and especially the document click listener part should be removed because this has proven to be buggy.

PR #558, opened 2021-03-26
docs: add script to fetch design system pages for storybook

KaiVandivier:
This adds a script to fetch the 'Principles' pages from the Design System and formats for them for use in the storybook so users can see that info side-by-side with component demos.  The fetch script runs before the `build-storybook` command in the `demo` script, and the fetched files are added to a git-ignored-directory in `./docs`.

To avoid having to download the whole design system and all of its assets, any links to other pages or assets are transformed to external Github links. External links and same-page links are left as-is.


mediremi:
Looks really good :+1: 

Maybe we should update the links to the principles in https://github.com/dhis2/ui/blob/master/docs/constants.md to point to the storybook versions?

KaiVandivier:
Yeah that sounds good 🙂

cooper-joe:
Good idea @KaiVandivier. I'm all for making the principles more accessible and easy to find.

Regarding the placement, I am concerned about the storybook sidebar growing and having to present many different types of content. The structure of these pages is not completely clear to me going forward. 

For example, this "[homepage](https://ui.dhis2.nu/#/)" is perhaps a more logical place for reading about `ui`. The storybook is linked as "Live demo", which isn't where I'd expect to find principles and such content. So I think it depends on the status of [the homepage](https://ui.dhis2.nu/#/), though I can't find a documented decision about how we're moving forward with it. Perhaps eventually the storybook will replace that page, then obviously we need to support all kinds of content. If the storybook is intended as "demos only", then I think keeping documentation-only pages somewhere else would make sense.

@KaiVandivier, are you aware of a decision that's been made about these pages?

KaiVandivier:
@cooper-joe That's a valid concern - I don't think the final decision about the documentation future is written down (although it's mentioned in some notes [here](https://github.com/dhis2/ui/pull/506#issuecomment-782001191)), but my understanding of the conclusion we reached during the storybook renovation code review is that the storybook would become the primary documentation source; I'll ping @varl to confirm. Getting Started, Recipes, etc. would live in the storybook, but the existing 'home page' would still present an auto-generated API page like it does now (except that it would ideally use React Docgen to generate the API instead of JSDocs so that the source code only needs one kind of comments to document it).  There was some talk about eschewing the API page because the storybook can provide that information, which would make the storybook encapsulate everything, but I know @varl likes the API page.

I think it's best to have everything in one place, and since the documentation-only pages can go in the storybook but the demos can't go in the doc site, the storybook is the way to go, and it supports docs-only pages pretty well.

How does that sound to you @cooper-joe? Maybe this requires a proposal to kind of 'deprecate' the existing home page/doc site to have a definitive deicision

KaiVandivier:
@mediremi Linking to the storybook pages has turned out to be a bit complicated by the page being present on both the homepage/doc site _and_ the storybook (which will hopefully be temporary) because they handle links differently.  Relative links from markdown on storybook are initiated from an iframe as root (though this is not the case [when using MDX](https://github.com/storybookjs/storybook/issues/8618#issuecomment-632671601)).

One solution is to link to 'https://ui.dhis2.nu/demo/?path=/docs/slug-slug-slug--page', which will work from the doc site and the storybook, but will require a full refresh on the storybook and won't be very useful during development. Maybe those tradeoffs are okay for now, though.

If this were a storybook-only page, linking from one story to another using MDX would not require a full refresh of the page and could use a relative url that works both during development and deployment, so this is more motivation to pursue consolidating UI documentation in the storybook.

cooper-joe (approved):
I can't comment on the implementation details, but I generally approve.

One question I do have: are there consequences to the Design System repo? I assume that I'll just keep adding/updating content there as necessary and anything in the `principles` folder will be pulled in automatically?

KaiVandivier:
Thanks Joe!
That's right about the design system repo 🙂 you don't need to do anything differently, and the script will fetch the current files from GitHub when the storybook is built for deployment.
About the sidebar: do you think it would be better if the files in 'Using UI' and 'Design System Principles' were collapsed under a folder in their respective sections? (Like the 'Recipes' or 'Button' folders)

cooper-joe:
> About the sidebar: do you think it would be better if the files in 'Using UI' and 'Design System Principles' were collapsed under a folder in their respective sections? (Like the 'Recipes' or 'Button' folders)

I think that could potentially be required in the future (maybe splitting the Principles into folders based on theory/practice, or something) - but I don't think it's necessary yet. It's good to keep that option in mind though if/when the amount of pages in those sections grows.

package.json
```diff
@@ -75,6 +75,8 @@
         "concurrently": "^5.3.0",
         "enzyme": "^3.11.0",
         "enzyme-adapter-react-16": "^1.15.5",
+        "lodash": "^4.17.21",
```
HendrikThePendric, line 78: maybe just add `lodash.startcase` instead?
`yarn add lodash.startcase`
and `const startCase = require('lodash.startcase');`

docs/fetch-design-system-pages.js
```diff
@@ -0,0 +1,89 @@
...
+        fileInfo.filter(
+            ({ name, type }) => type === 'file' && name.endsWith('.md')
+        )
+    )
+    // Fetch files and create local ones for storybook
+    .then(downloadAndProcessFiles)
+    .then(() =>
+        console.log('Design system files downloaded & processed for storybook')
+    )
+    .catch(console.error)
```
HendrikThePendric, line 89: Catching the error and logging it to the console implies we only want to notify about this failure. I think for something like this, that does probably make sense: we can still showcase our components without including the design system files. However, this does imply that the storybook demo is still functional with these and I have not verified this.

Also, I think we should probably include some additional information about the context before printing this error to the console, i.e.:
> The following error occurred while downloading and processing design system files. The design system will not be included in the demo storybook.
(and then the error)

HendrikThePendric (changes requested):
I'm probably not the best person to review this PR. The main point I'm unsure about is how you are producing the files via template strings. We could switch to a slightly more robust AST approach. Having said that, I can't fault the template string approach either, there's nothing very complex going on, so perhaps sticking to a simple approach is best...

Besides that I have placed 2 comments in the code and I think we should address/discuss the error handling point before proceeding.

package.json
KaiVandivier: Good idea :) after looking into it, that method will be deprecated, but it seems like there's a [better way](https://lodash.com/per-method-packages):
> if you import or require methods directly, e.g. const throttle = require('lodash/throttle'), only the subset of lodash code your package uses will be bundled in projects that use your package.

docs/fetch-design-system-pages.js
KaiVandivier: That's a good point -- if there's something that could go wrong that would interfere with the rest of the storybook, it would be partially-made story files that could cause run-time errors. In that case, if there's an error, maybe the directory the design system files get copied into should get wiped to delete any malformed files

KaiVandivier:
Thanks @HendrikThePendric!

Nice tip about the AST processing -- I agree that using a template string is probably good enough for this application, BUT there are some things that could be better about the link transformations and other things related to migrating from GitHub to storybook, and the AST approach might help with that. I'll look into it!

I think another improvement for the file outputs is to skip making `.md` files and write the content straight into the `.stories.mdx` files instead of importing it from there, which will save a few steps and make the template a bit simpler.

varl:
Hey @KaiVandivier and @cooper-joe :wave: 

Fun topic and interesting PR, nice work @KaiVandivier and thanks for digging into this.

Over the last 2 years we (me and joe) have talked about merging the design system with the UI repo, and last time I checked that's still the long term goal. Perspectives change over time though, so I think it would be good if we three (@varl, @KaiVandivier, @cooper-joe) have a meeting and discuss the way forward before we merge this.

Let's change it to a draft to be on the safe side.

KaiVandivier:
Sounds good 👍

PR #572, opened 2021-04-13
feat: add native list attribute to Input

mediremi:
This adds the [native `<input>` `list` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdeflist) to the `Input` and `InputField` components.


varl:
The datalist stories are not showing up in the deploy preview: https://deploy-preview-572--dhis2-ui.netlify.app/demo/?path=/story/forms-input-input--with-datalist

PR #579, opened 2021-04-28
refactor: component packages

varl:
## UI Restructure meeting - 26 May, 2021

Outcome:

- DECISION: We will go ahead with this proposal for now to split out components into separate packages.
- DECISION: We deprecate the collections [of components] `ui-core` and `ui-widgets`.
- DECISION: Will merge to `alpha` after cleaning up the `reorg` branch and iron out bugs there.

## Links

- [Pros/Cons document](https://docs.google.com/document/d/1CJxndnkkvA_hYq9u9PbMO_dxr-wD5_baNQRPW11VZzw/edit#)
- [2021-05-19](https://github.com/dhis2/notes/blob/master/minutes/2021/05/19-ui-components-proposal.md)

## To do before `alpha`

- COMPLETE

## To do in `alpha`

- https://jira.dhis2.org/browse/LIBS-184
- https://jira.dhis2.org/browse/LIBS-185
- https://jira.dhis2.org/browse/LIBS-186
- https://jira.dhis2.org/browse/LIBS-187
- https://jira.dhis2.org/browse/LIBS-188

## Done

- [x] Take a pass at the groups in `components/` and see if we can logically group them, preferably close to the Design System classification.
- [x] Cypress needs to be adjusted.
- [x] `yarn build` in the root package needs to be sorted as it would be nice to avoid listing all components as scripts.
- [x] Jest needs to look in `components` as well as `packages`.
- [x] Hot reload for components (HMR / Fast Refresh) both in "full" and "per component" modes.
- [x] `manypkg` linter run for best practices with monorepo.
- [x] `yarn test` needs to pass


mediremi:
> "What goes where" = components go in components/. That's it !

:tada: :tada: :tada: 

> Each components needs a package.json, d2.config.js, and some other boilerplate.

Maybe we should add a script to the top-level `package.json` so that we can do something like `yarn component:new <component name>` and all the boilerplate is setup?

varl:
> > "What goes where" = components go in components/. That's it !
> 
> tada tada tada
> 
> > Each components needs a package.json, d2.config.js, and some other boilerplate.
> 
> Maybe we should add a script to the top-level `package.json` so that we can do something like `yarn component:new <component name>` and all the boilerplate is setup?

We can do that, I have a "boilerplate" component locally that I'm using to make porting a bit easier:

```
tmp/component/
├── d2.config.js
├── package.json
└── src
    └── index.js

1 directory, 3 files
```

Then I copy that component to `components` with the `component-name` and run `sed` to replace `{component}` with `component-name` and `{Component}` with `ComponentName`.

After that it's pretty much a `mv` from `packages/core/src/ComponentName` to `components/component-name/src` and updating the references in the respective packages, and adding the dependencies of course.

Everything is easily scriptable.

Mohammer5:
I might ask some really dumb question because I'm not seeing the bigger picture.. I have a few things I'm a bit confused about:

### Do we lose proper tree shaking when using individual packages?
Or do we have to add a lot of peer instead of normal dependencies to the individual component packages that use other components?

### Does this reorg tackle how apps could re-build their own components which the building blocks?
* Do we want this at all actually? (maybe it's an indicator that components are missing in UI or we should question why it's done in the first place; On the other hand, if an app has a need that's not provided by UI, they can quickly implement their own version, example from the past: controlled dropdown button, menus, etc)

### How to apps manage to keep the version of all components they use in sync?
This could get a little odd when an app needs to update one of the components but leaves the other ones.
If we still expose a single `ui` package that all apps are going to use, why should component be their own package in the first place instead of just folders?

> Per-component dependencies means no complex rules about which components are allowed to import other components.

From what I understand so far, this means: Any component can use any other component, it just needs to specify it as its dependency. What is the difference to having an eslint rule which specified that you can only import from the `index.js` of other modules

#### The killer feature

> neat, i think the killer-feature is to be able to do yarn start to load a single component into storybook. it's a lot faster than loading all of it, and very nice when working on a single component.

This should be possible with `yarn start ./components/button` (which can be forwared to storybook's `main.js` and applied there)

varl:
> This should be possible with `yarn start ./components/button` (which can be forwared to storybook's `main.js` and applied there)

There are ways to solve this in the current structure as well, for sure, like picking up `process.argv` in the `main.js` file and use that when finding relevant stories, as you mentioned. There needs to be more care taken that the path is correct when resolving the stories but other than that it should be functionally equivalent.

I would argue that it is not quite as nice and straightforward as `yarn start` inside of the package itself, or using a workspace command if launched from the project root. It's predictable, and works without special knowledge.

But this isn't the point of the reorg, it was just a "Ahhh yes, _finally this just works_" realization that made me happy.

varl:
> How to apps manage to keep the version of all components they use in sync?

Apps should depend on `@dhis2/ui` just as before.

`@dhis2/ui` explicitly depends on all the components and re-exports them, as it does today.

`@dhis2/ui-core` and `@dhis2/ui-widgets` will not get any _new_ components to re-export, but they will also re-export the ones they do today for consistency, and to keep this as a non-breaking change.

But, for example, App Hub is able to depend on specific components, and can avoid using components that require e.g. the app-runtime on a granular level. In that case, Dependabot will _help_ keep the component versions in sync.

varl:
> Does this reorg tackle how apps could re-build their own components which the building blocks?

No, it doesn't. It intentionally ignores this question.

That said, I think that this reorg opens up other options for us when discussion the composition of UI components. First and foremost, it allows us to separate the topics on:

- how UI is structured (core vs. widgets, elements vs. composes, etc.) and,
- how composition works.

Which has been a massive roadblock for a long time now (a year or more).

varl:
>If we still expose a single `ui` package that all apps are going to use, why should component be their own package in the first place instead of just folders?

Everything is still "just folders", so this is the usual "monorepo or not" discussion. The arguments for monorepos still applies and the trade-offs are well understood and accepted.

Does this proposal comes with new problems that we need to consider ?

varl:
> From what I understand so far, this means: Any component can use any other component, it just needs to specify it as its dependency. What is the difference to having an eslint rule which specified that you can only import from the `index.js` of other modules

While it is possible to lint for those relationships, I rather it is solved explicitly using structure, than to create and maintain rules that jump at me when running a linter.

So the difference is in the way it goes about getting to the same ends: one uses a tool with custom rules on how to enforce the structure, and the other relies on structure provided by the packages ecosystem.

ismay:
> Does this proposal comes with new problems that we need to consider?

I talked to JG a bit about this, and I do agree with him that the main risk with this approach is that we're introducing more complexity. In other organisations I have worked with codebases that split up functionality in separate node modules for everything, and it was always quite a hassle to maintain. In my opinion if modules have to be released together to not break then the modules have been scoped wrong.

I've already had a lot of trouble with our monorepo setup, so I can imagine introducing even more package boundaries could aggravate that.

I know I originally mentioned this approach in the thread with chakra, so it might seem weird for me to have second thoughts about it. But in thinking about it I do agree with JG that it's seems like it introduces even more complexity than we currently have. In thinking about it, I agree that the one undisputable module boundary seems to be @dhis2/ui. So moving that to a single module makes sense to me as well.

We often gravitate towards technology because that's the tool we know and have. But I don't think it's neccesarily the right tool for the problem at hand. I think the core problem is still clear scope and categorization. Maybe we should move to a single module as that seems like an undisputable, simple starting point, and then move ahead from there.

HendrikThePendric:
I think moving in this direction makes a lot of sense. The continued what-goes-where discussion has been extremely counter-productive and finding a good structure that doesn’t force us to classify components on a conceptual level combats this.

In my mind the advantages of solving the what-goes-where discussion outweigh all potential downsides that have been mentioned here.

varl:
>Maybe we should move to a single module as that seems like an undisputable, simple starting point, and then move ahead from there.

This is a enticing option for the same reasons that "rewriting from scratch" is alluring. Let's start again and avoid the mistakes and bad decisions that were made.

UI started out as a single library. The reason for the split between ui-core and ui-widgets was that the UI components that do not depend on the DHIS2 API should be usable in situations that do not involve the DHIS2 API.

A few examples of this is DHIS2 branded software such as: App Hub, Developer Portal, application prototypes, and application mockups, to name a few concrete ones.

varl:
> I know I originally mentioned this approach in the thread with chakra, so it might seem weird for me to have second thoughts about it.

Pointing out an option doesn't imply that you support it, so doesn't seem weird to me. I've rejected this approach before, solely based on it being used in `d2-ui` and there were problems with it. Workspaces and monorepos has come a long way since then, and after @HendrikThePendric and @amcgee both brought up the option, I decided to let go of my bias (for a while) and re-evaluate it on its own merit.

This branch is the result of that exploration, and likewise I ask you to let go of biases against monorepos (for a little while) and explore it without committing to it.

ismay:
> This is a enticing option for the same reasons that "rewriting from scratch" is alluring. Let's start again and avoid the mistakes and bad decisions that were made.

I think I see this differently. To me a single lib does a lot of the same things that this PR is doing, just differently. Both move away from the hierarchy that we were previously enforcing. The thing that this PR allows you to do (keep publishing ui-core and ui-widgets) would not be possible with a single module though. But we'd keep our recommended entrypoint `dhis2/ui` intact.

The reason I'm arguing for a single module is that in thinking about it that makes the most sense to me as a module boundary. It's what our users are installing at the moment. I think it's also hard to really decouple all our components from eachother with just semver as an interface. So that's also why I'm leaning towards a single module.

> likewise I ask you (as in everyone, including myself) to let go of biases and hypotheticals (for a little while) and explore it without committing to it.

Definitely. Completely agree. That's also why I initially suggested it, thought we might as well explore it. But in looking at it, besides the above, I think we're taking a social/planning problem and instead of solving the social issue directly we're tending towards a technical solution. I don't think a technical solution will be able to solve a problem that at its core doesn't seem technical to me.

Way I see it the core problem we're running into with ui is because of unclear scoping (what's the responsibility of the lib, what belongs in an app, etc.) and misalignment between devs in how things should be done in the lib. I think structural changes will still run into the same social issues, just with a different structure.

Maybe we should focus on the social/planning issues first? And then tackle the technical stuff later? Let me know what you think!

Mohammer5:
I think we have to split this discussion into two different topics:
* `elements & composes` vs `components`
* Multiple workspaces and/or packages vs one

The former one has nothing to do with publishing, transpilation, etc. as each component can be a workspace, regardless of its location.

---

I do like the move to go from multiple folders to one components folder, it simplifies things quite a bit :ok_hand: 
In the end, the consumers of the library do not care about the distinction and internally it has caused quite a bit of confusion.
We can still formulate some rules how to make our components "hookable" (like the `renderOption` prop of the `Transfer` component, making sure that every component is controllable and as well that every component does not use the api but instead expects a callback that can be used with parameters to fetch data (e. g. with the OrgUnitTree))

---

> While it is possible to lint for those relationships, I rather it is solved explicitly using structure, than to create and maintain rules that jump at me when running a linter.

It's one rule that doesn't need to be maintained if I'm not mistaken ([this one](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-internal-modules.md) I believe).
Considering that the maintenance burden is quite low when having each component in its own folder & its internals co-located with the `index.js` of that folder (which also helps identifying potentially too-complex components with a lot of files), to me it feels that having different packages is more of a maintenance burden.

> The arguments for monorepos still applies and the trade-offs are well understood and accepted.

I'm still not sure I understand what the explicit benefits of a monorepo would be. I'm trying to do some research about this and my findings indicate that it's a way to group multiple projects (that's a term that's shared across all articles I've read so far), which makes it easier to handle dependencies from one project to another

> With multiple repos, you typically either have one project per repo, or an umbrella of related projects per repo, but that forces you to define what a “project” is for your particular team or company, and it sometimes forces you to split and merge repos for reasons that are pure overhead. For example, having to split a project because it's too big or has too much history for your VCS is not optimal.
>
> With a monorepo, projects can be organized and grouped together in whatever way you find to be most logically consistent, and not just because your version control system forces you to organize things in a particular way. Using a single repo also reduces overhead from managing dependencies.
> ([source](https://danluu.com/monorepo/))

> * Developers only need to check out and update one (or a few) repositories. New projects are created in subdirectories and automatically distributed to the other developers without them having to check out additional repositories.
> * Developers have a consistent status of the overall project at all times – not only in the master but also on branches and old commits. However, with many individual repositories, incompatibilities can occur quickly.
> * Atomic Commits: Related changes to multiple subprojects can be checked in together.
> * Creating and merging branches for changes that involve multiple subprojects is time-consuming if the projects are stored in separate repositories. In a monorepo, you only have to merge once.
> * Handling of merge conflicts that refer to earlier states of the code base is simplified since in a monorepo the complete code base is in a consistent state while handling the conflict. In the case of several repositories, it may be necessary to bring them to a consistent state manually.
> ([source](https://www.happycoders.eu/java/monorepos-advantages-disadvantages/#What_are_the_advantages_of_a_monorepo))

> yarn workspaces constitutes the only technology that comes with native capabilities for Mono-Repos. In contrast to lerna, you do not have to execute a separate step for bootstrapping dependencies of the packages. yarn install does the trick by installing the dependencies of the root folder and then for every package.
> ([source](https://doppelmutzi.github.io/monorepo-lerna-yarn-workspaces/))

To me it seems that none of these apply to us. We do not have multiple projects but simply individual components. We don't need independent dependency management (in fact we're trying to avoid adding dependencies as much as possible). With tree-shaking, bundle size isn't an issue either. Some apps, which are just a few in numbers, will have to install the UI's defined peer dependencies, but it's not really a maintenance burden but rather an edge case.

In essence I agree with what @ismay wrote above and I think these are issues we can't solve with a technological approach:

> I think we're taking a social/planning problem and instead of solving the social issue directly we're tending towards a technical solution. I don't think a technical solution will be able to solve a problem that at its core doesn't seem technical to me.
> 
> Way I see it the core problem we're running into with ui is because of unclear scoping (what's the responsibility of the lib, what belongs in an app, etc.) and misalignment between devs in how things should be done in the lib.

---

If we want to go with a monorepo, maybe we should be looking into lerna, at least when I'm reading this:

> Even if you pick yarn workspaces for dependency management, it is a good idea to use lerna as well. The reason is that lerna provides utility commands to optimize management of multiple packages. For example, with one lerna command you can iterate through all or particular packages, running a series of operations (such as linting, testing, and building) on each package. Thereby, it compliments yarn workspaces that takes over the dependency management process.
>
> Using lerna for testing or linting from within the root folder is faster than invoking all operations manually from every package folder. John Tucker’s blog post deals with testing with lerna in great detail. 
> ([source](https://www.happycoders.eu/java/monorepos-advantages-disadvantages/#What_are_the_advantages_of_a_monorepo))

varl:
We have used Lerna before, and it is not a good fit and was not trivial to migrate away from. Lerna makes things a lot more complex than releasing multiple packages as the same version, and it makes semantic-release impossible to use in a monorepo.

varl:
OK, I need help with answers to these two questions, because I don't see how to do it with the proposal of a unified `@dhis2/ui` package:

- How can we move towards a unified `@dhis2/ui` without a breaking change?
- How can we support using UI components in non-DHIS2 contexts (no API and no App Platform) with a unified `@dhis2/ui` package?

The use case for having access to `@dhis2/ui-core` components is one that we support and that external developers rely on: https://github.com/dhis2/ui/network/dependents?package_id=UGFja2FnZS00NTIwNjMxMDE%3D

This is a requirement that I think we were spot on with from the start, and there are more public examples of `@dhis2/ui-core` being used than `@dhis2/ui` being used. And least of all people are using `@dhis2/ui-widgets`.

The cleanest way to a unified `@dhis2/ui` package would be to put all components that use the DHIS2 API in the App Platform, and all components that do not in `@dhis2/ui`. But this is literally the original separation between `ui-core` and `ui-widgets`, so why not just keep that then?

Mohammer5:
> How can we move towards a unified @dhis2/ui without a breaking change?

79 Projects have a dependency on `@dhis2/ui-widgets`, 135 have a dependency on `@dhis2/ui-core`. Unfortunately we can't see how many of those using `@dhis2/ui-core` also use `@dhis2/ui-widgets`. We can only make assumptions. My intuition is that there are many who use `core` + `widgets` (so they could also switch to `ui`) solely because that was the way before we introduced `@dhis2/ui`.

If we really want to still publish updates to the core- and widgets-library, then we could have multiple entry points and temporarily replace the the `package.json` before publishing a specific package so that names and dependencies are as needed: 
* `src/index.js` => `@dhis2/ui`
* `src/core.js` => `@dhis2/ui-core`
* `src/widgets.js` => `@dhis2/widgets`

Another alternative would be to have a core and widgets workspace that re-export the parts from `@dhis2/ui` that are being exported today by these libraries.

---

> How can we support using UI components in non-DHIS2 contexts (no API and no App Platform) with a unified @dhis2/ui package?

Right now we only have dhis2-only-contextual components: OrganisationUnitTree and HeaderBar. There are many more components in `widgets` though. If we want to separate these out, we could have a second workspace that uses `@dhis2/ui` and exports the dhis2-only components.

My preference would be to change the OrgUnitTree slightly so it accepts a callback that it can use to fetch the next org units instead of using the app-runtime directly, then it'd could be re-used in non-dhis2 contexts. With the `HeaderBar` we could do the same, forcing the app platform to provide the appropriate callbacks (which kind of makes sense to me), which moves the whole app-runtime dependency to `app-platform` and the apps using the org unit tree.

With that out of the way, we could deprecate `ui-core` and recommend to use `ui` if they want to have the latest updates. Tree-shaking should ensure that the bundle size doesn't get bloated.

These two change would be breaking changes, although only the OrgUnitTree change should be of consequence. I think it's a fair change that I've proposed several times already regardless of this topic.

varl:
> If we really want to still publish updates to the core- and widgets-library, then we could have multiple entry points and temporarily replace the the package.json before publishing a specific package so that names and dependencies are as needed

This is more complicated than packaging everything individually and publishing it.

> Another alternative would be to have a core and widgets workspace that re-export the parts from @dhis2/ui that are being exported today by these libraries.

This makes `ui` a dependency of both `ui-core` and `ui-widgets`, and pulls along all the deps from `ui` into `ui-core` and `ui-widgets`. I don't see how this is cleaner even if it is tree-shaken out. The user is still going to see a lot of warnings about peer dependencies.

Both these options seem more convoluted to implement and goes to great lengths to avoid something that is technically simpler and more explicit. I might be missing how this is intended to work, so if you are proposing this as a serious option to explore, please prepare a branch for how that can practically work.

> These two change [moving app-runtime to app-platform, callback-driven data fetch] would be breaking changes, although only the OrgUnitTree change should be of consequence. I think it's a fair change that I've proposed several times already regardless of this topic.

I think both of those proposals make sense, and that they make sense with this proposal. I actually think it would be fantastic for this proposal as it means that a user could pick and choose any component to install and not worry about getting hit with app-runtime requirements.

mediremi:
> My preference would be to change the OrgUnitTree slightly so it accepts a callback that it can use to fetch the next org units instead of using the app-runtime directly, then it'd could be re-used in non-dhis2 contexts. With the HeaderBar we could do the same, forcing the app platform to provide the appropriate callbacks (which kind of makes sense to me), which moves the whole app-runtime dependency to app-platform and the apps using the org unit tree.

I think it'd make more sense to have `OrgUnitTree` and `HeaderBar` as separate packages. The organisation unit tree is very DHIS2-specific and having to provide callbacks in every place we use it would be pretty painful.

Mohammer5:
> if you are proposing this as a serious option

I'd prefer to just deprecate ui-core and ui-widgets, especially if we remove the app-runtime dependency (regardless of how we do it). That seems to be the easiest option to me. Then we wouldn't have to have a monorepo and we wouldn't have to do weird hacks just to publish ui-core and ui-widgets.

I wasn't even aware that we want to do this as we were talking about some serious changes in regards to package structure and internal organisation.

varl:
Over the years, we've carried out multiple, and attempted more, restructures of the UI repo, e.g. moving away from BEM CSS to "PostCSS Modules", to StyledJSX, to split repos (ui => ui-core & ui-widgets), and to a monorepo (ui-core, ui-widgets, ui-forms => ui), to other monorepo structures like `ui-base` and `ui-composes`/`ui-elements` and _this_ "every component is a package" has by far been the simplest to do, reason about, and get working with minimal technical fuss, so I'm a bit surprised by the comments about this being more complex.

I'm having the opposite experience working with it, it is very consistent and pleasant to not think about where a component comes from, or which way it is appropriate to import it with.

Imports such as `@dhis2/ui-button` is a lot nicer than `../../Button.js` or `../index.js` in `ui-core`, and from `@dhis2/ui-core` when working in `ui-widgets`, as is entirely consistent across all packages and situations, even in stories.

Even in a "unified" package structure, imports are a mess. There's a reason devs resort to `resolve` hacks and why packages that use e.g. `~` to indicate the root source directory in imports are popular, and linting rules are used to mitigate the problem.

The main annoyance is that because of how `import`s resolve, the `build/` folders must exist, but that's true of all code that uses a build step.

We'd still need to wrangle Storybook internals since we use non-standard configurations for StyledJSX in Babel, so I don't see a great simplification there either.

Even migrating components out of `@dhis2/ui` (should we want to) becomes (relatively speaking) simpler and can be moved to a completely different repo and deprecated, before way before triggering a breaking change. Today, if we move it we break it. The structure is more flexible and resilient than I initially thought.

In addition, we can introduce this without triggering any breaking changes. We don't even have to front the change at all, it can be done internally to the lib.

This also means that if turns out to be way too complicated and full of problems, we have a breaking change "to spare" where we consolidate everything into a unified package. :wink:

varl:
Thanks for the discussion !

I've heard your comments loud and clear. I am going to continue work on this branch and see if I bump up against any of the concerns voiced here. Even if we end up closing this PR without merging it, it's worth investigating a bit deeper.

ismay:
> OK, I need help with answers to these two questions, because I don't see how to do it with the proposal of a unified @dhis2/ui package:
> 
>     How can we move towards a unified @dhis2/ui without a breaking change?
>     How can we support using UI components in non-DHIS2 contexts (no API and no App Platform) with a unified @dhis2/ui package?

Yeah I think it'll be hard to do so without a breaking change.

> The cleanest way to a unified @dhis2/ui package would be to put all components that use the DHIS2 API in the App Platform, and all components that do not in @dhis2/ui.

Hmm, I like that tbh, seems like a clear distinction 👍

> But this is literally the original separation between ui-core and ui-widgets, so why not just keep that then?

True. I think from my perspective the argument against the monorepo approach within ui is simplicity. Besides that, keeping only those components in ui that really belong in ui seems like a good way to simplify things as well. End goal being less confusion about what goes where and a little more curation before things are put in ui.

varl:
We generate source files into `src` for translations and icons, so on a clean checkout, there are missing source files.

Since we generate sources on build, we have to run the build scripts before starting Storybook.

So the commands required to get this up and running is:

```
git clone git@github.com:dhis2/ui.git --branch reorg --single-branch && cd ui

yarn install

yarn workspace @dhis2/ui-icons build
yarn workspace @dhis2/ui-widgets d2-app-scripts i18n generate
yarn workspace @dhis2/ui-forms d2-app-scripts i18n generate

yarn start
```

This is not unique for the monorepo approach either. In a unified repo we would still need to generate the icons and generate locales since we do not commit those sources. In both cases, we need to `build` before we `start` to generate our required sources.

More drastic measures, like committing locales and/or the interim build icons, fix this for both worlds.

varl:
Build all and start: 23 s for manager and 25 s for preview 
Only build icons, generate locales: 16 s for manager and 19 s for preview

`--no-manager-cache` for both runs.

varl:
No need to lock the Yarn version anymore, so that's been removed in 3856378.

mediremi:
Since we're removing the distiction between core and widgets, will the `dataTest` default prop values be changed from `dhis2-{uicore,uiwidget}-X` to `dhis2-ui-X`?

Mohammer5:
> Since we're removing the distiction between core and widgets, will the dataTest default prop values be changed from dhis2-{uicore,uiwidget}-X to dhis2-ui-X?

That's a very good question. If we're going to use a dedicated components package, then I'd prefer `dhis2-components-xxxx`

varl:
I think that is a breaking change from our perspective, so for now, we'll keep them intact as they are and figure out a migration path down the line.

I think I prefer `dhis2-ui-{component}` as it matches the package name better and is more predictable.

PR #580, opened 2021-05-04
chore(deps): upgrade to new cli-style version to fix editors plugins

HendrikThePendric:
I had been having issues with the VSCode ESLint plugin for quite some time now, but just assumed that was to do with the plugin itself. However, after realising the WebStorm IDE ESLint plugin was breaking in exactly the same way, @varl had a look as well and we realised that upgrading `@dhis2/cli-style` to the latest alpha resolved this problem.

Apart from resolving the issue above, having a separation between code-linting and style-linting seems to be very nice.

So the official purpose of this PR is that we try out the new `@dhis2/cli-style` package to see if it works well.

And the officious purpose is that I get my squiggly lines back.


varl:
@HendrikThePendric please run `yarn d2-style install` and commit the `.hooks` directory.

ismay:
Ah, nice. I was planning on taking a look at the cli-style alpha again. The caching work got in the way, but since that's waiting for feedback I'll take a look at the pending work now.

ismay:
I've created a PR where we can review the merge of cli-style alpha to master here: https://github.com/dhis2/cli-style/pull/389. After that we could upgrade cli-style here.

.eslintrc.js
```diff
@@ -6,6 +6,7 @@ const config = {
         cy: 'readonly',
         Cypress: 'readonly',
     },
+    ignorePatterns: ['.yarn/**/*.js'],
```
ismay, line 9: We could also use an .eslintignore file here. I don't really have a particular preference though.

HendrikThePendric: yeah, good point. That's probably a bit more consistent given the fact we also have a `.prettierignore` now

.gitignore
```diff
@@ -2,6 +2,7 @@
 
 # Editor settings
 /.vscode
+.idea/
```
ismay, line 5: I was wondering about the start vs the trailing slash: https://stackoverflow.com/a/38559600. I think we should probably use a consistent pattern to match directories, `.idea/` is different from `node_modules/` for example. I'd align them.

.eslintrc.js
ismay: Actually I see we're using a .prettierignore as well. For consistency I'd go with an .eslintignore.

.storybook/preview.js
```diff
@@ -64,6 +64,7 @@ export const parameters = {
     },
     docs: {
         // Customize docs page layout (in order to rename 'Stories' section)
+        /* eslint-disable-next-line react/display-name */
```
ismay, line 67: We could also disable this rule globally for all `*.stories.js` and for everything in `.storybook` with an [override](https://eslint.org/docs/user-guide/configuring/configuration-files#how-do-overrides-work).

HendrikThePendric: Shall we perhaps leave that out-of-scope for now?
If we change this later, after discussing, we will get a warning/error about the superfluous comment anyway.

.eslintrc.js
HendrikThePendric: le done

package.json
```diff
@@ -33,13 +33,9 @@
...
         "docs": "d2-utils-docsite build ./docs -o ./dist --jsdoc ./packages/**/src --jsdoc-output-file api.md",
-        "format": "yarn format:js && yarn format:text",
-        "format:staged": "yarn format:js --staged && yarn format:text --staged",
-        "format:js": "d2-style js apply",
-        "format:text": "d2-style text apply",
-        "lint": "concurrently -n w: \"yarn:lint:*\"",
-        "lint:js": "d2-style js check",
-        "lint:text": "d2-style text check",
+        "format": "d2-style apply",
+        "format:staged": "yarn apply --staged",
```
ismay, line 37: I suspect the `yarn` here is supposed to be a `d2-style`.

.eslintignore
```diff
@@ -0,0 +1 @@
+.yarn/**/*.js
```
ismay, line 1: I think you can just ignore this entire dir. btw. Same syntax as gitignore.

package.json
HendrikThePendric: well spotted, fixed now.

ismay:
When running `yarn format` from the latest commit on this branch, this is the output I get:

Screenshot 2021-05-05 at 16 19 35

1. To avoid the warning, we should probably pass a more specific glob to d2-style, to prevent it from linting the file and logging the warning.
2. The output directly below `javascript > prettier` is a little confusing to me, I'm not sure what's happening there. Seems like something's going wrong. Especially since it says that code-style issues have been fixed, but when I run git status, there are no changed files. Also the formatting seems off.

ismay:
The above problem does not occur when I run yarn lint:

Screenshot 2021-05-05 at 16 27 51

ismay:
I thought maybe using a `.` would help resolve point 1. from my earlier comment, but that results in this error:

Screenshot 2021-05-05 at 16 31 53

I'm not sure what glob syntax we use for d2-style. I think we need @varl's help to either craft a bespoke glob here (so that we're not passing ignored files to d2-style), or we need to address it in cli-style.

ismay:
So Viktor and I took a look at what was causing the above:

- The weird formatting was caused (it seems) by prettier cutting off the newline when the terminal isn't wide enough
- I wasn't able to reproduce the `Code style issues fixed` false positve, so what I suspect is that prettier did actually fix files, but that it fixed them to a state which was exactly the same as the git history, so it didn't show up in git as changes.
- Viktor has created a PR for the .yarn dir ignore warning: https://github.com/dhis2/cli-style/pull/392
- The `No matching javascript files for pattern ...` warning on using a `.` is something we can take a look at later.

So from my perspective that takes care of everything that stood out about this PR. Looks good to merge to me 👍

ismay (approved):
I'm not sure if there's anything from @varl's perspective that still needs to be addressed about this PR, but to me it looks like we've addressed everything that stood out 👍

varl (approved):

PR #585, opened 2021-05-11
chore(d2-style): skip ignored files for linting

ismay:
This uses what was introduced in https://github.com/dhis2/cli-style/pull/400 to omit ignored files from the files passed to eslint. This prevents the lint warning you'd otherwise get on running `yarn lint`.

See here for the lint step on CI, which doesn't log any warnings now: https://github.com/dhis2/ui/pull/585/checks?check_run_id=2557102715#step:7:17

Compared to master: https://github.com/dhis2/ui/runs/2547217632#step:7:17 (which usually logs even more if the icons are built)

### DIFF
--- a/d2-style.config.js
+++ b/d2-style.config.js
@@ -1,0 +1,10 @@
+module.exports = {
+    patterns: {
+        js: [
+            '*.{js,jsx,ts,tsx}',
+            '!packages/icons/src/react/**/*.js',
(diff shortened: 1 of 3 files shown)


mediremi (approved):

varl (commented):
Very nice. :+1:

PR #602, opened 2021-05-13
chore(dev-deps): upgrade cypress related packages to latest

ismay:
Closes #592, closes #539, closes #540

Upgrading the above three packages at once, since they're all related.

Initially I thought the .get and .find calls would need to be updated because they've been renamed (see [here](https://github.com/dhis2/cli-utils-cypress/blob/master/CHANGELOG.md#code-refactoring-1)). However, I see we're using local custom commands so this update won't break anything (which explains why the e2e tests all pass). As a follow up we could move those to the commands supplied by the libs upgraded here.


HendrikThePendric (commented):
> As a follow up we could move those to the commands supplied by the libs upgraded here.

I agree. But then I would argue this PR shouldn't get merged until we at least have a JIRA issue to capture that work. Could you create a JIRA issue and post a link here?

ismay:
> > As a follow up we could move those to the commands supplied by the libs upgraded here.
> 
> I agree. But then I would argue this PR shouldn't get merged until we at least have a JIRA issue to capture that work. Otherwise we are going to forget. Could you create a JIRA issue and post a link here?

Sure thing, here it is: https://jira.dhis2.org/browse/LIBS-179

varl (approved):

PR #608, opened 2021-05-25
fix(chip): change prop types so they accept any content

Mohammer5:
Fixes LIBS-178
([link](https://jira.dhis2.org/browse/LIBS-178))


HendrikThePendric (commented):
@cooper-joe could you review this PR? The design system doesn't mention this super explicitly, but I remember that a `chip` should only contain text and an optional leading icon.
https://github.com/dhis2/design-system/blob/master/atoms/chip.md

Code is looking good.

varl:
This is from the Dashboard PWA design, so I assume that it comes from @cooper-joe and the design-system specs are simply old. Good to verify though, thanks @HendrikThePendric.

Code looks good to me.

varl (approved):

cooper-joe (approved):
> This is from the Dashboard PWA design, so I assume that it comes from @cooper-joe and the design-system specs are simply old.

Yes, this is correct. This is an approved update to the component.

PR #609, opened 2021-05-25
feat(modal): add "hide" prop to prevent content from unmounting when hiding (WIP)

Mohammer5:
Fixed LIBS-182
([link](https://jira.dhis2.org/browse/LIBS-182))

This introduces a new layer: background (= `z-index: -1`).
That was needed to make the layer to into the background when hiding the Modal. Rendering the content in a hidden div instead of the Layer made the components unmount, so that's not an option unfortunately.

This introduces a new prop on the Layer: `transparent` (= `background: none`)
When the modal is hidden and the page does not have a background, the translucent background is still shown. `transparent` prevents that from happening, the prop has a higher priority than `translucent` so both props can be set (see usage of Layer in the implementation of the Modal component)

### DIFF
--- a/packages/core/src/Modal/Modal.js
+++ b/packages/core/src/Modal/Modal.js
@@ -31,6 +31,12 @@ const centeredContent = resolve`
     }
 `
 
+const layerStyles = resolve`
+    div {
+        background: none;
+    }
+`
+
 /**
  * @module
  * @param {Modal.PropTypes} props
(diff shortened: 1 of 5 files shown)


varl:
Thanks for looking into this @Mohammer5. The `Layer` z-index stacking makes this a bit more annoying than I had hoped, but I forgot that we never did merge this: https://github.com/dhis2/ui/pull/474

If we merge #474 we can refactor this to make it simpler, since at that point we can set a custom mount node for the Portal, and thus hide it efficiently.

So in terms of this PR we should try to keep the changes internal, e.g. consider hard coding the layer `-1` instead of adding a constant for it. The Layer component also takes a `className`, so we could use `css.resolve` to implement the `transparent` part when we need to hide it in the Modal.

I think those two changes should isolate everything we need to the Modal component, and make it easier to refactor in #474.

Does that makes sense ?

packages/constants/src/layers.js
```diff
@@ -3,6 +3,7 @@
  * @desc DHIS2 application layers
  */
 export const layers = {
+    background: -1,
```
varl, line 6:
```suggestion
```

varl, line 6:
```suggestion
```

packages/core/src/Layer/Layer.js
```diff
@@ -33,6 +33,7 @@ const Layer = ({
     onClick,
     position,
     translucent,
+    transparent,
```
varl, line 36:
```suggestion
```

packages/core/src/Layer/Layer.js
```diff
@@ -55,6 +56,7 @@ const Layer = ({
                         `level-${level}`,
                         {
                             translucent,
+                            transparent,
```
varl, line 59:
```suggestion
```

packages/core/src/Layer/Layer.js
```diff
@@ -90,6 +92,9 @@ const Layer = ({
                         div.translucent {
                             background-color: rgba(33, 43, 54, 0.4);
                         }
+                        div.transparent {
+                            background: none;
+                        }
```
varl, line 97:
```suggestion
```

packages/core/src/Layer/Layer.js
```diff
@@ -123,6 +128,8 @@ Layer.propTypes = {
     position: PropTypes.oneOf(['absolute', 'fixed']),
     /** Adds a semi-transparent background */
     translucent: PropTypes.bool,
+    /** Has no background whatsoever **/
+    transparent: PropTypes.bool,
```
varl, line 132:
```suggestion
```

packages/core/src/Modal/Modal.js
```diff
@@ -85,26 +86,39 @@ export const Modal = ({
...
-                max-height: calc(100vh - ${2 * spacersNum.dp64}px);
-                max-width: calc(100vw - ${2 * spacersNum.dp64}px);
-                width: 600px;
-            }
+    return (
+        <Layer
+            onClick={onClose}
+            level={hide ? layers.background : layers.blocking}
+            transparent={hide}
+            translucent
```
varl, line 96:
```suggestion
            level={hide ? -1 : layers.blocking}
            className={hide ? layerStyles.className : ''}
            translucent={!hide}
```

packages/core/src/Modal/Modal.js
```diff
@@ -85,26 +86,39 @@ export const Modal = ({
             {scrollBoxCard.styles}
             {centeredContent.styles}
```
varl, line 87:
```suggestion
            {centeredContent.styles}
            {layerStyles.styles}
```

packages/core/src/Modal/Modal.js
```diff
@@ -62,14 +62,15 @@ const centeredContent = resolve`
...
-    position,
     dataTest,
-}) => (
-    <Layer onClick={onClose} level={layers.blocking} translucent>
+    hide,
+    large,
+    onClose,
+    position,
+    small,
+}) => {
```
varl, line 72: This should be on L33 in this file, but the review wouldn't let me put it there so inlining it in the render function for illustration.

```suggestion
}) => {
    const layerStyles = resolve`
        div {
            background: none;
        }
    `
```

varl (commented):
See my review for my suggestion on how to keep this isolated to the Modal component so that we can refactor it more easily when we look at #474.

packages/core/src/Modal/Modal.js
```diff
@@ -31,6 +31,14 @@ const centeredContent = resolve`
     }
 `
 
+// The double class name so the specificity is guaranteed to be higher than
+// that of the rule of the translucent background of the Layer itself
```
varl, line 35: This comment is out of date.

varl:
```suggestion
```

varl (approved):
Very nice. Looks good to me !

PR #614, opened 2021-06-01
feat(alert-bar): add hidden prop

HendrikThePendric:
This PR adds a hidden prop to the `AlertBar` component which can be used to toggle its visibility programatically. This change will enable me to clean up the implementation of https://github.com/dhis2/app-runtime/pull/820, but I think it is also in line with the changes we've been doing to the `Modal` in https://github.com/dhis2/ui/pull/609.

One notable difference between the `Modal`'s implementation and the way I have implemented the `hidden` prop in the `AlertBar` is that at the `AlertBar` will still render `null` at some point in its lifecycle. I've done so because I don't think the `AlertBar` should ever contain stateful children, and because I like my DOM to be tidy 😄 . However, it would be fairly trivial to change this and keep the `div` in the DOM with `display: none;`. Let me know if that is preferred.

I've also added a new test case and a story for toggling visibility via the new `hidden` prop.

I also did some general refactoring because even though it was me who wrote the initial implementation, I found it pretty confusing. So I changed the following:
- Port from a class to a function component
- Use much more explicit variable and function names
- Clear all timers whenever anything changes. I encountered some weird edge cases when toggling the `hidden` prop quickly and often after introducing the hidden prop to the initial implementation. This had to do with some timers still running that shouldn't have been. It was fairly hard to reason about and this "brute force" approach with `clearAllTimers` works solidly
- Moved some test code into a single file since this is what we're doing in most places


HendrikThePendric:
Input from @varl about the `AlertBar` rendering `null`:
> as long as the implementation is consistent (i.e. we use `hidden` not `show`) i think the other part is an implementation detail that we can change if there is a use case for stateful children in the alert bar without it being a breaking change.

packages/core/src/AlertBar/AlertBar.js
```diff
@@ -20,110 +20,133 @@ import { Message } from './Message.js'
...
+    duration,
+    hidden,
+    icon,
+    permanent,
+    success,
+    warning,
+    onHidden,
+}) => {
+    const [inViewport, setInViewport] = useState(() => !hidden)
+    const [inDOM, setInDOM] = useState(() => !hidden)
```
mediremi, line 38:
```suggestion
    const [inViewport, setInViewport] = useState(!hidden)
    const [inDOM, setInDOM] = useState(!hidden)
```

packages/core/src/AlertBar/AlertBar.js
```diff
@@ -20,110 +20,133 @@ import { Message } from './Message.js'
...
-        if (this.shouldAutoHide()) {
-            const elapsedTime = Date.now() - this.startTime
-            this.timeRemaining = this.timeRemaining - elapsedTime
-            clearTimeout(this.displayTimeout)
+    const runShowAnimation = () => {
+        clearAllTimeouts()
+        show()
+        showTimeout.current = setTimeout(startDisplayTimeout, ANIMATION_TIME)
+    }
+    const pauzeDisplayTimeout = () => {
```
mediremi, line 85:
```suggestion
    const pauseDisplayTimeout = () => {
```

packages/core/src/AlertBar/AlertBar.js
```diff
@@ -20,110 +20,133 @@ import { Message } from './Message.js'
...
+        <div
+            className={cx(className, {
+                info,
+                success,
+                warning,
+                critical,
+                inViewport,
+            })}
+            data-test={dataTest}
+            onMouseEnter={pauzeDisplayTimeout}
```
mediremi, line 125:
```suggestion
            onMouseEnter={pauseDisplayTimeout}
```

packages/core/src/AlertBar/AlertBar.js
```diff
@@ -20,110 +20,133 @@ import { Message } from './Message.js'
...
+            onMouseLeave={resumeDisplayTimeout}
```
mediremi, line 126: I know this is out of scope for this PR but it'd be nice if we also paused the timer when the component is focused in order to make alerts keyboard accessible.

packages/core/src/AlertBar/AlertBar.js
```diff
@@ -20,110 +20,133 @@ import { Message } from './Message.js'
...
-            }, this.timeRemaining)
+    const clearAllTimeouts = () => {
+        clearTimeout(showTimeout.current)
+        clearTimeout(displayTimeout.current)
+        clearTimeout(hideTimeout.current)
+    }
+    const runHideAnimation = () => {
+        clearAllTimeouts()
+        hide()
+        hideTimeout.current = setTimeout(remove, ANIMATION_TIME)
```
mediremi, line 67: I know you mentioned previously that you were looking into using the `animationend` event instead of using `setTimeout` - was `setTimeout` used due to browser support issues (and so we could look into using `animationend` in the future) or was this the simpler implementation?

HendrikThePendric: It was decided to not use the `animationend` event because critical functionality depends on it, and support is not rock-solid. If the event would not fire the `onHidden` callback would not be called and this could lead to serious issues. For example the DOM could get cluttered with lots of alert-bars by the alert service. See [this Slack thread](https://dhis2.slack.com/archives/CBM8LNEQM/p1622472566054000) to read how the decision came about.

So we decided to play it safe and stick with the `setTimeout` approach. We've got three timed events, the show-animation, the hide-animation and the display duration. If I had gone with the `animationend` event based implementation I would have been able to replace the timeouts for the show/hide animations with an event-listener and I would expect that to be a lot easier to reason about.

Having said that, I think the current implementation is actually not too complex now that I have switched to more explicit variable names, and opted for a "brute force" approach to clearing timeouts (I just clear all of them all the time to make things simpler).

packages/core/src/AlertBar/AlertBar.js
HendrikThePendric: Great suggestion. Could you create a new JIRA issue for this, and post a link to it here, so we don't forget?

mediremi: https://jira.dhis2.org/browse/LIBS-190

mediremi (approved):

packages/core/src/AlertBar/AlertBar.js
HendrikThePendric: Thanks!

PR #629, opened 2021-06-15
chore(release): cut reorganization

varl:
# Todo

- [x] Verify dependencies
- [x] Verify no breaking changes on the consumer side
  - [x] Tested in app-platform via `examples/simple-app`
  - [x] Tested in Usage Analytics: https://dhis2.vardevs.se/usage-analytics-result.html
  - [x] Tested in Data Visualizer: https://dhis2.vardevs.se/data-visualizer-result.html
  - [x] Tested in Scheduler: https://dhis2.vardevs.se/scheduler-result.html
  - [x] Tested in Import/Export: https://dhis2.vardevs.se/import-export-result.html
  - [x] Tested in SMS Config: https://dhis2.vardevs.se/sms-config-result.html
- [x] Verify correct use of dependencies and peer dependencies
  - In `master` we have set most internal dependencies as peer dependencies which I assume is done for de-duplication purposes. In `alpha` I don't think that this approach is suitable given the structure and amount of packages, since it makes consuming the `@dhis2-ui` components very painful. Instead internal dependencies are added as bundled dependencies.
- [x] Update to latest dev dependencies.
- [x] Update workflows
- [x] Use `ui-icons` instead of old embedded icons.


Mohammer5:
LGTM, it seems to work, tests pass & I really like that i can navigate to an individual component folder and run the storybook there.
One thing that I'd prefer is to put all components, collections and utilities into a `src` folder to keep them distinct from the project configuration files & folders. It also makes it a bit more intuitive to see at one glance what's relevant for the built library and what's not

ismay:
> In master we have set most internal dependencies as peer dependencies which I assume is done for de-duplication purposes. In alpha I don't think that this approach is suitable given the structure and amount of packages, since it makes consuming the @dhis2-ui components very painful. Instead internal dependencies are added as bundled dependencies.

Yeah if deduplication is the reason that doesn't seem like correct usage of peerdependencies. This is the intended usecase: https://nodejs.org/es/blog/npm/peer-dependencies. So if we have dependencies that we plug into, like react/react-dom, etc., that's when a peerdependency makes sense.

ismay:
- ~~This might be unrelated to the changes in this PR, but the bundle analyzer results mentioned for the 5 apps seem rather large. I assume this is before gzip/minification? Still seems big to me then, but it _seems_ as if ui isn't significantly larger than other dependencies. Anyway, main relevant point for this PR is: have we verified there's no regression in the bundled size?~~ (edit: alpha is not to blame: https://dhis2.slack.com/archives/CBM8LNEQM/p1624277112138000?thread_ts=1623915474.134600&cid=CBM8LNEQM)
- I think utilities is a bit of a catchall term. I can see the difference between components on the one hand, and icons/constants on the other. However I don't see the relation between icons and constants. I think utilities has a bit of a risk of becoming a dumping ground for anything that's not a component. What do you think of a top level folder for constants and icons, does that make sense for the setup you envision? To me that runs less risk of confusion: icons separate because it's got a separate build process, constants separate as it's not a component.
- I think the architecture doc is a little hard to digest. It explains the architecture, but it requires a lot of processing to get the information you want. I think framing it in the context of scenarios could help. For example, adding a component. Or using a component as a dev. On that note, I can imagine that this info could be split up to either the README (if it's useful for a user), or CONTRIBUTING (if it's useful for a contributor). I think probably only the most motivated will read this doc otherwise, and that seems like a shame as it's useful info.

HendrikThePendric (approved):
Looks excellent! All changes I expected were present. As @Mohammer5 already mentioned, it's great to be able to run storybook for individual components. I think that'll really speed up development time.

I considered perhaps moving the `CssReset` and `CssVariables` components to `utilities`, but after reading the architecture documentation that wouldn't be a good idea because these are react components.

So Great work 👍  🚀

varl:
> To me that runs less risk of confusion: icons separate because it's got a separate build process, constants separate as it's not a component.

Yeah, that works for me. Fixed in b98389c.

varl:
> I think the architecture doc is a little hard to digest. It explains the architecture, but it requires a bit of processing to get the information you want.

Yeah, I think there is potential to refine that document, but I wanted to get a first revision down in writing in this PR.

> I think framing it in the context of scenarios could help. For example, "adding a component to the lib". Or "using a component as a dev". On that note, I can imagine that this info could be split up to either the README (if it's useful for a user), or CONTRIBUTING (if it's useful for a contributor). I think probably only the most motivated will read this doc otherwise, and that seems like a shame as it's useful info.

I agree with this, but I thought to add them in another place. I also want docs for additional use cases, e.g.

- "this is how you are supposed to build other libraries with UI components"
- "this is how you are supposed to consume UI components in an app"
- "this is how you recompose components to customize behavior".

The ARCHITECTURE file _is_ for the motivated. I don't expect casual contributors to have to worry about it. I do expect that ~core~ contributors that propose changes to the structure itself, to first read the ARCHITECTURE file before questioning the way things are.

varl:
:warning: here we go !

PR #632, opened 2021-06-17
fix(menu-item): do not call event.preventDefault() in onClick handler if href prop is set and user is holding meta/alt/ctrl/shift key (alpha)

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-191


Mohammer5:
This PR should not be necessary, it should be enough to put it into master (cc @varl; you'll know for sure)

varl (approved):

PR #641, opened 2021-06-22
docs(props): remove jsdoc typedefs for props

varl:
We prefer to use React doc strings to document props, as that is what
most of the tools in the React ecosystem expects, e.g. Storybook and
React docgen.

Having JSDoc typedefs in additon means that we have two keep two
documentation styles in sync with the prop changes.

A follow-up for some components may be required to add relevant React
docstrings to props where we used to have typedefs coverage.


mediremi (approved):
This just reduced our maintenance burden by ~:infinity:!
Thank you :smile: 


I think we might also want to remove this: [https://github.com/dhis2/ui/blob/9e1dbf88a033981f8ca96ff43ad1f268e1777e96/components/organisation-unit-tree/src/organisation-unit-tree.js#L144-L155](https://github.com/dhis2/ui/blob/9e1dbf88a033981f8ca96ff43ad1f268e1777e96/components/organisation-unit-tree/src/organisation-unit-tree.js#L144-L155)

varl:
> I think we might also want to remove this:
> 
> https://github.com/dhis2/ui/blob/9e1dbf88a033981f8ca96ff43ad1f268e1777e96/components/organisation-unit-tree/src/organisation-unit-tree.js#L144-L155

Good catch !

varl:
The Transfer, FileInput, and OrganisationUnitTree uses JSDoc strings for helper functions, and I have left those because they make the functions a lot easier to understand. Note that these are not Components, so using JSDoc here is alright.

PR #644, opened 2021-06-23
fix(organisation-unit-tree): handle filtering when units' children have not yet been loaded

mediremi:
When initially loading an organisation unit, the organisation unit tree will not request its `children` field and so when the `OrganisationUnitNode` component calls `computeChildNodes` (see below) `node.children` will be `undefined` resulting in an error.

https://github.com/dhis2/ui/blob/e21ae4ace1ba8931dd7a5eb0d96ab4f828efe07c/components/organisation-unit-tree/src/compute-child-nodes.js#L27-L38


varl (approved):
Looks alright to me.

PR #649, opened 2021-06-28
fix: Add default value for options

SferaDev:
Options is not required but if not provided it breaks with ``map does not exist on undefined`` error

Not sure the code style of the library, alternatively mark ``options`` as a required field, however that's change over the API-facing props.


varl (approved):

varl:
I think this is a pragmatic fix that's OK to merge.

PR #655, opened 2021-06-30
Match border color with InputFields

SferaDev:
``Transfer`` and ``InputFields`` have a slightly different border color (``grey500`` vs ``grey400``) and looks strange when an interface combines them.

Since ``Transfer`` should be the least used component of both, changing to match the same border color of ``Input``.

image

Note: The [design system page](https://github.com/dhis2/design-system/blob/master/organisms/transfer.md) should be probably updated?

### DIFF
--- a/components/transfer/src/common/styles.js
+++ b/components/transfer/src/common/styles.js
@@ -1,5 +1,5 @@
 import { colors } from '@dhis2/ui-constants'
 
 export const backgroundColor = colors.white
-export const borderColor = colors.grey400
+export const borderColor = colors.grey500
 export const borderRadius = '3px'


cooper-joe:
Hi @SferaDev, thanks for the suggestion.
The different border colors here are intentional. The reasoning is that inputs are individual components where everything inside is interactive, whereas the Transfer is a container around multiple elements that aren't necessarily all interaactive (like headers). The lighter border is intended as a container-border.

I've found that using Transfers in dedicated sections works well, see example screenshot below. This can help frame the Transfer as a contained task as part of the form, rather than another, simpler input field.

Example usage:
image

PR #659, opened 2021-07-06
fix(tag): support 16px icons

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-198


cooper-joe (approved):
Design changes look good to me. Thanks for fixing 👍.

PR #666, opened 2021-07-15
fix(select): do not throw error when selected prop does not correspond to value of existing option

mediremi:
Thanks to [recent work integrating `react-query`](https://github.com/dhis2/app-runtime/pull/824) into the app runtime, apps will soon be able to enable and take advantage of the stale-while-revalidate fetching strategy. Unfortunately, some of our components were not designed with SWR and background refresh in mind - including the select components `SingleSelect` and `MultiSelect`.

These components expect the value(s) of their `selected` prop to correspond to the value of a `SingleSelectOption`/`MultiSelectOption` and will throw an error if this is not the case. This assumption can fail under SWR in the case that the options (and hence the value of the `selected` prop) are generated based on fetched data:

```JSX
/*
 * If the current value of `selected` refers to an item that was previously present but has now
 * been removed on the backend - and background refresh results in `data` being updated - then
 * SingleSelect will throw an error.
 * 
 * One way to guard against this issue is to use
 * `selected={data.map(i => i.value).includes(selected) ? selected : ''}`
 */
const [selected, setSelected] = useState('')
const { data } = useDataQuery(someQuery)

<SingleSelect
  selected={selected}
  onChange={({ selected }) => setSelected(selected)}
>
  {data.map(item => <SingleSelectOption key={item.value} label={item.label} value={item.value} />)
</SingleSelect>
```

### DIFF
--- a/components/select/src/multi-select/features/invalid_selected.feature
+++ b/components/select/src/multi-select/features/invalid_selected.feature
@@ -1,0 +1,6 @@
+Feature: Invalid selections are ignored
+
+    Scenario: Some of the MultiSelect's selected values do not have a corresponding option
+        Given a MultiSelect with selected values that do not have a corresponding option
+        Then the valid selected values are rendered
+        And some errors are written to the console
--- a/components/select/src/multi-select/features/invalid_selected/index.js
+++ b/components/select/src/multi-select/features/invalid_selected/index.js
@@ -1,0 +1,28 @@
+import { Given, Then } from 'cypress-cucumber-preprocessor/steps'
+
+Given(
+    'a MultiSelect with selected values that do not have a corresponding option',
+    () => {
+        cy.visitStory('MultiSelect', 'With some invalid selected options', {
+            onBeforeLoad(win) {
+                cy.stub(win.console, 'error').as('consoleError')
+            },
+        })
+    }
+)
+Then('the valid selected values are rendered', () => {
+    cy.get('[data-test="dhis2-uicore-select-input"]').should(
+        'contain',
+        'option three'
+    )
+})
+Then('some errors are written to the console', () => {
+    cy.get('@consoleError').should(
+        'be.calledWith',
+        'There is no option with the value: "1". Make sure that all the values passed to the selected prop match the value of an existing option.'
+    )
+    cy.get('@consoleError').should(
+        'be.calledWith',
+        'There is no option with the value: "2". Make sure that all the values passed to the selected prop match the value of an existing option.'
+    )
+})
--- a/components/select/src/multi-select/input.js
+++ b/components/select/src/multi-select/input.js
@@ -6,8 +6,28 @@ import {
     InputClearButton,
     InputPlaceholder,
     InputPrefix,
+    findOptionChild,
+    removeOption,
 } from '../select/index.js'
 import { SelectionList } from './selection-list.js'
+
+const createSelectedOption = ({ selected, onChange, value, options }) => {
+    const optionChild = findOptionChild(value, options)
+    if (optionChild) {
+        const { label, disabled } = optionChild.props
+        return {
+            value,
+            label,
+            disabled,
+            onRemove: (_, e) => {
+                const filtered = removeOption(value, selected)
+                const data = { selected: filtered }
+
+                onChange(data, e)
+            },
+        }
+    }
+}
 
 const Input = ({
     selected,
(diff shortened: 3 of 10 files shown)


ismay:
> Thanks to [recent work integrating `react-query`](https://github.com/dhis2/app-runtime/pull/824) into the app runtime, apps will soon be able to enable and take advantage of the stale-while-revalidate fetching strategy. Unfortunately, some of our components were not designed with SWR and background refresh in mind - including the select components `SingleSelect` and `MultiSelect`.
> 
> These components expect the value(s) of their `selected` prop to correspond to the value of a `SingleSelectOption`/`MultiSelectOption` and will throw an error if this is not the case. This assumption can fail under SWR in the case that the options (and hence the value of the `selected` prop) are generated based on fetched data:
> 
> ```js
> /*
>  * If the current value of `selected` refers to an item that was previously present but has now
>  * been removed on the backend - and background refresh results in `data` being updated - then
>  * SingleSelect will throw an error.
>  * 
>  * One way to guard against this issue is to use
>  * `selected={data.map(i => i.value).includes(selected) ? selected : ''}`
>  */
> const [selected, setSelected] = useState('')
> const { data } = useDataQuery(someQuery)
> 
> <SingleSelect
>   selected={selected}
>   onChange={({ selected }) => setSelected(selected)}
> >
>   {data.map(item => <SingleSelectOption key={item.value} label={item.label} value={item.value} />)
> </SingleSelect>
> ```

I was thinking about this change: so with the conservative changes we're making currently devs shouldn't start encountering errors if they previously didn't. The current behaviour will be basically identical.

So I guess this could become an issue if the app-runtime starts doing things like refetching automatically on window focus, etc. right? As in that case data could be refreshed outside of say, a manual refetch by the developer (in which case I'd expect them to deal with supplying the right options), and in that case I suppose it's possible for devs to not have anticipated the change and guard against the changing data.

Because the current behaviour is a kind of safeguard against having a selection that can't be associated with a value. Which seems useful. Having a selection that can't be associated with a value could cause weird bugs.

So maybe, instead of smoothing over the select being used with a selection that doesn't have an associated option, we should just recommend that the select isn't used with SWR if this can cause problems. To me that seems like the safer option. What do you think?

mediremi:
> I was thinking about this change: so with the conservative changes we're making currently devs shouldn't start encountering errors if they previously didn't. The current behaviour will be basically identical.

> So I guess this could become an issue if the app-runtime starts doing things like refetching automatically on window focus, etc. right? As in that case data could be refreshed outside of say, a manual refetch by the developer (in which case I'd expect them to deal with supplying the right options), and in that case I suppose it's possible for devs to not have anticipated the change and guard against the changing data.

That's right - this change isn't in response to our current conservative changes but rather in preparation for when we enable automatic refetching.

> Because the current behaviour is a kind of safeguard against having a selection that can't be associated with a value. Which seems useful. Having a selection that can't be associated with a value could cause weird bugs.

> So maybe, instead of smoothing over the select being used with a selection that doesn't have an associated option, we should just recommend that the select isn't used with SWR if this can cause problems. To me that seems like the safer option. What do you think?

My main reason for making this PR was that when a select component throws an error due to an invalid `selected` prop, the user loses their current work and is taken to the 'something went wrong' page. If this issue were caused by background refetching then it would (correctly) appear to the user that the app they are using suddenly broke for no clear reason and would make our job of finding the root cause from bug reports a nightmare.

However, you're right that this is an important safeguard that we should keep and that's something that I overlooked. Maybe in addition to warning developers about the issues with using select and SWR together, we should also make SWR opt-in *per query* instead of per app when we do enable it?

mediremi:
I'm closing this PR since this is an important safeguard that we should keep.

ismay:
> My main reason for making this PR was that when a select component throws an error due to an invalid `selected` prop, the user loses their current work and is taken to the 'something went wrong' page. If this issue were caused by background refetching then it would (correctly) appear to the user that the app they are using suddenly broke for no clear reason and would make our job of finding the root cause from bug reports a nightmare.
> 
> However, you're right that this is an important safeguard that we should keep and that's something that I overlooked. Maybe in addition to warning developers about the issues with using select and SWR together, we should also make SWR opt-in _per query_ instead of per app when we do enable it?

Yeah I agree. I think it definitely needs a per query way of configuring it. And I wonder if we should even eventually move to making SWR opt-out. Maybe it even makes sense to always leave it on opt-in.

PR #667, opened 2021-07-20
feat: loader updates (UX-6)

cooper-joe:
Implements [UX-6](https://jira.dhis2.org/browse/UX-6)

This PR adjusts the loading components, in particular the circular loader:
- simplifies shape and animation to reduce motion, especially when using multiple spinners on a page.
- adds a static background track to improve visibility on varied backgrounds.
- adds an `invert` mode for use on colored backgrounds, (will be used in an upcoming PR to add loading modes to the `Button` component).
- adds an `xsmall` size. (I added this to `sharedPropTypes.sizePropType`, even though `Loader` is the only one using it so far).
- adjusts linear loader for consistency with circular styles.

### DIFF
--- a/components/loader/src/circular-loader/circular-loader.js
+++ b/components/loader/src/circular-loader/circular-loader.js
@@ -1,79 +1,62 @@
-import { theme, spacers, sharedPropTypes } from '@dhis2/ui-constants'
+import { colors, sharedPropTypes } from '@dhis2/ui-constants'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React from 'react'
 
-const CircularLoader = ({ small, large, className, dataTest }) => (
+const CircularLoader = ({
+    small,
+    large,
+    extrasmall,
+    invert,
+    className,
+    dataTest,
+}) => (
     <div
         role="progressbar"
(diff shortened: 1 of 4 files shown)


components/loader/src/circular-loader/circular-loader.js
```diff
@@ -86,8 +69,10 @@ CircularLoader.defaultProps = {
 CircularLoader.propTypes = {
     className: PropTypes.string,
     dataTest: PropTypes.string,
+    invert: PropTypes.bool,
     large: sharedPropTypes.sizePropType,
     small: sharedPropTypes.sizePropType,
+    xsmall: sharedPropTypes.sizePropType,
```
Mohammer5, line 75: I'm not a big fan of abbreviations in code in general..
In this particular case it's quite easy to confuse `xsmall` with `small` as there's only one letter difference and `extrasmall` requires less thinking when reading code

cooper-joe: Thanks for the tip, that makes a lot of sense. Changed to `extrasmall`.

Mohammer5 (approved):

PR #669, opened 2021-07-21
feat(button): Loading state

cooper-joe:
Fixes [UX-22](https://jira.dhis2.org/browse/UX-22).

This PR adds a loading state to the `Button` component. The loading state prepends the `CircularLoader` component to the `Button` children. If an `icon` is present, it is replaced by the `CircularLoader`.

This PR also adjusts and adds stories to make testing the loading spinner states easier.

Note: This is my (potentially) naive implementation of how a `loading` state. Please do say if there is a better way of doing this.

### DIFF
--- a/components/button/src/button/button.js
+++ b/components/button/src/button/button.js
@@ -1,3 +1,4 @@
+import { CircularLoader } from '@dhis2-ui/loader'
 import { sharedPropTypes } from '@dhis2/ui-constants'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
@@ -24,6 +25,7 @@ export const Button = ({
     onBlur,
     onClick,
     onFocus,
+    loading,
 }) => {
     const ref = useRef()
 
@@ -44,6 +46,7 @@ export const Button = ({
         large,
         'icon-only': iconOnly,
         toggled,
+        loading: loading,
     })
 
     return (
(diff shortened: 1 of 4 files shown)


cooper-joe:
One thing I don't understand, why is the `no-extraneous-dependencies` being triggered by the `CircularLoader` import at the top? I had a look at how imports were done for other components, [InputField for example](https://github.com/dhis2/ui/blob/master/components/input/src/input-field/input-field.js#L1), and those are done in the same way and don't seem to trigger this lint error?

components/button/src/button/button.js
```diff
@@ -24,6 +25,7 @@ export const Button = ({
     onBlur,
     onClick,
     onFocus,
+    isLoading,
```
mediremi, line 28: While I like having booleans/predicates start with `is` and `has`, existing components use `loading` so that's probably what we should use here (e.g. https://github.com/dhis2/ui/blob/master/components/select/src/single-select/single-select.js#L162).

mediremi:
> One thing I don't understand, why is the no-extraneous-dependencies being triggered by the CircularLoader import at the top? I had a look at how imports were done for other components, InputField for example, and those are done in the same way and don't seem to trigger this lint error?

Adding `"@dhis2-ui/loader": "6.11.0"` to the dependencies section of `components/button/package.json` should fix that error

components/button/src/button/button.js
cooper-joe: Thanks for pointing that out! Agreed that consistency is good. Fixed.

mediremi (approved):
LGTM :heavy_check_mark: 

Now we just need to update a bazillion buttons :joy:

PR #670, opened 2021-07-22
feat: update library-wide focus styles

cooper-joe:
Partially fixes [UX-20](https://jira.dhis2.org/browse/UX-20)

This PR updates the `:focus` styles for all components to establish a standard, consistent focus style. This PR adjusts styles, but does not add keyboard functionality for several items that are current missing it.

For the following elements, `:focus` styles have been added, but it's not possible to actually focus the element. This additional functionality will be covered by another ticket. This applies to `menu-item` and `file-input`.

### DIFF
--- a/components/button/src/button/button.styles.js
+++ b/components/button/src/button/button.styles.js
@@ -32,31 +32,15 @@ export default css`
     }
 
     button:focus {
-        /* Prevent default browser behavior which adds an outline */
-        outline: none;
-    }
-
-    /* Use the ::after pseudo-element for focus styles */
-    button::after {
-        content: ' ';
-        pointer-events: none;
-
-        position: absolute;
-        top: -2px;
-        right: -2px;
-        bottom: -2px;
-        left: -2px;
-
-        border: 2px solid transparent;
-        border-radius: inherit;
-
-        transition: border-color 0.15s cubic-bezier(0.4, 0, 0.6, 1);
+        outline: 3px solid ${theme.focus};
+        outline-offset: 2px;
+        transition: none;
     }
 
     /* Prevent focus styles on active and disabled buttons */
-    button:active:focus::after,
-    button:disabled:focus::after {
-        border-color: transparent;
+    button:active:focus,
+    button:disabled:focus {
+        outline: none;
     }
 
     button:hover {
@@ -73,10 +57,6 @@ export default css`
 
     button:focus {
         background-color: #f9fafb;
-    }
-
-    button:focus::after {
-        border-color: ${theme.primary600};
     }
 
     button:disabled {
@@ -119,7 +99,6 @@ export default css`
 
     .primary:active,
     .primary:active:focus {
-        border-color: ${theme.primary800};
         background: linear-gradient(180deg, #054fa3 0%, #034793 100%);
         background-color: #1c4a90;
         box-shadow: 0 0 0 1px rgba(0, 0, 0, 0.18) inset;
@@ -128,10 +107,6 @@ export default css`
     .primary:focus {
         background: linear-gradient(180deg, #1565c0 0%, #0650a3 100%);
         background-color: #285dac;
-    }
-
-    .primary:focus::after {
-        border-color: ${colors.yellow300};
     }
 
     .primary:disabled {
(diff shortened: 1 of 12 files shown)


components/tab/src/tab/tab.js
```diff
@@ -72,9 +72,11 @@ const Tab = ({
                 text-overflow: ellipsis;
                 transition: fill 150ms ease-in-out;
             }
-
-            button:focus span {
-                outline: 1px dotted ${colors.grey900};
+            /*todo: decide below*/
+            /*avoids the focus-after-click that was reported as confusing. not supported by safari though, possible to polyfill?*/
+            button:focus-visible {
```
cooper-joe, line 77: Using `:focus-visible` here is nice because it avoids the "focus-on/after-click" that has been reported as confusing. It's not yet out of [CSS4 draft](https://caniuse.com/css-focus-visible), though it has decent coverage. Is this a bad idea?

mediremi: According to https://css-tricks.com/platform-news-using-focus-visible-bbcs-new-typeface-declarative-shadow-doms-a11y-and-placeholders/, a good way to deal with lack of widespread coverage is to define focus styles in `:focus` and then undo those in `:focus-visible`.

cooper-joe: Nice! Thanks @mediremi. That seems like a reasonable workaround to me -- I can make the changes necessary.

varl:
@cooper-joe Any further action needed on this or can it be posted as a PR and merged ?

cooper-joe:
> @cooper-joe Any further action needed on this or can it be posted as a PR and merged ?

@varl - ready for review now. There's some outstanding changes necessary to completely fix [UX-20](https://jira.dhis2.org/browse/UX-20), but I'll open LIBS tickets for those changes.

components/button/src/button/button.styles.js
```diff
@@ -32,31 +32,15 @@ export default css`
...
-        top: -2px;
-        right: -2px;
-        bottom: -2px;
-        left: -2px;
-
-        border: 2px solid transparent;
-        border-radius: inherit;
-
-        transition: border-color 0.15s cubic-bezier(0.4, 0, 0.6, 1);
+        outline: 3px solid ${theme.focus};
```
varl, line 35: The table in UX-20 says 2px outline and 2px offset, is the typo in the code?

varl (changes requested):
Nice that we can use the semantic `outline` CSS property for `:focus`.

I have one comment that warrants response before merging.

cooper-joe:
Moving this back into _draft_ to work out a couple of issues raised from discussion:
- determine whether inconsistent browser handling of outline border-radius is an issue.
- use `outline` instead of `border` everywhere possible, not just the `button`.

cooper-joe:
This PR is ready for review again, after a rethink of the focus styles. [UX-20](https://jira.dhis2.org/browse/UX-20?focusedCommentId=43758&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-43758) covers the changes. The approach is now:
- using the proper semantic `outline` property wherever possible.
- using `outline` (or replacement) with `3px solid ${theme.focus}` for all components, to establish an (even more) consistent focus flow.

To note:
- `box-shadow` is used instead of `outline` (as suggested by @mediremi, thanks!) for the `input`, `textarea` and `select`, because the square `outline`s for Blink renderers looked too odd.
- Even though they're not needed, `radio`, `checkbox` and `switch` keep their `border: 2px solid transparent`, which was previously used to allow a `border` to be used as a focus style. Removing the `border` changes the size of the element, which may have weird consequences for UIs, so I'd rather leave this in.

components/button/src/button/button.styles.js
cooper-joe: It was a typo in UX-20, but it highlighted an inconsistency that I wasn't happy with. This has been changed and all elements now use the same outline width.

mediremi (approved):
LGTM :heavy_check_mark:

varl (approved):
Nice, looks great.

PR #671, opened 2021-07-22
Fix usage of updated loading spinner and icons

cooper-joe:
This PR fixes some minor issues: 
- where the updated `CircularLoader` caused layout issues
- where the new `extrasmall` size could replace a custom implementation

- also, while here, the `FileInput` was updated to use `ui-icons` rather than a custom icon.


cooper-joe:
Closed - to be covered in another ongoing PR.

PR #675, opened 2021-07-27
fix: circular loader sizing and placements

cooper-joe:
This PR fixes various issues that came up after the circular loader component was updated:
- file input: use new `extrasmall` loader size (also switches a few embedded SVGs to ui-icons)
- node: adjust spacing for updated loader.
- organisation unit tree: adjust spacing/sizing for updated loader.


cooper-joe:
Closing – covered in separate PRs.

PR #679, opened 2021-07-29
feat(org unit tree): allow node label customization (LIBS-203)

Mohammer5:
I've exposed a `renderNodeLabel` prop which can be used to customize the rendering.
The object passed to that function contains several state information as well as info about the node.
Additionally that object has a `label` property which is what the internal `Label` component will render.
That prop defaults to the display name.

I've created a story to highlight how it's supposed to work


components/organisation-unit-tree/src/__e2e__/displaying_loading_error.stories.e2e.js
```diff
@@ -1,21 +1,25 @@
...
 import React from 'react'
-import { OrganisationUnitTree } from '../organisation-unit-tree.js'
+import { OrganisationUnitTree } from '../index.js'
 import {
     StatefulMultiSelectionWrapper,
     dataProviderData,
     namespace,
 } from './common.js'
 
 const dataPriverDataWithError = {
```
mediremi, line 11:
```suggestion
const dataProviderDataWithError = {
```

components/organisation-unit-tree/src/features/path_based_filtering/index.js
```diff
@@ -31,7 +31,10 @@ Given(
  * left empty intentionally
  * ========================
  * Step is necessary to make the feature file understandable
- * Required states are prepared by the story
+ * Required states are prepared by the story.
+ * There is no way to check if the given is true by checking the dom
+ * as this can only be verified by opning the org unit tree,
```
mediremi, line 36:
```suggestion
 * as this can only be verified by opening the org unit tree,
```

components/organisation-unit-tree/src/organisation-unit-tree/default-render-node-label/label/label.js
```diff
@@ -0,0 +1,118 @@
...
+    if (singleSelection) {
+        return (
+            <LabelContainer highlighted={highlighted}>
+                <SingleSelectionLabel
+                    checked={checked}
+                    label={displayName}
+                    onChange={onClick}
+                    loading={loading}
+                >
+                    {displayName}
```
mediremi, line 76: [SingleSelectionLabel](https://github.com/dhis2/ui/blob/1712a034f9acc5e287e07d1aca452e0452dac5c8/components/organisation-unit-tree/src/organisation-unit-tree/default-render-node-label/label/single-selection-label.js) does not use the `children` prop so I think this can just be

```jsx
                <SingleSelectionLabel
                    checked={checked}
                    label={displayName}
                    onChange={onClick}
                    loading={loading}
                />
```

components/organisation-unit-tree/src/organisation-unit-tree/default-render-node-label/label/label.js
```diff
@@ -0,0 +1,118 @@
...
+                open={open}
+                hasChildren={hasChildren}
+            />
+        </LabelContainer>
+    )
+}
+
+Label.propTypes = {
+    dataTest: propTypes.string.isRequired,
+    displayName: propTypes.any.isRequired,
```
mediremi, line 102:
```suggestion
    displayName: propTypes.string.isRequired,
```

components/organisation-unit-tree/src/organisation-unit-tree/use-root-org-data/patch-missing-display-name.js
```diff
@@ -0,0 +1,29 @@
+export const fromEntries = entries =>
+    entries.reduce(
+        (collection, [key, name]) => ({
+            ...collection,
+            [key]: name,
+        }),
+        {}
+    )
```
mediremi, line 8: `Object.fromEntries` [has good browser support](https://caniuse.com/?search=fromEntries) so we should be able to use that instead

Mohammer5: Afaik we still support IE11

components/organisation-unit-tree/src/organisation-unit-tree/default-render-node-label/label/label.js
Mohammer5: I chose `any` intentionally so that this can be customized (see the story).
We could(/should?!) take a look at the arguments of the `renderNodeLabel` callback, I assume we can do better there..
Right now, in order to manipulate the node label, the node's `displayName` has to be modified. We could introduce an extra argument (`label`?), which defaults to the `displayName` but can be overwritten with `any`thing

Mohammer5: I'll add a comment for now

components/organisation-unit-tree/src/organisation-unit-tree/use-root-org-data/patch-missing-display-name.js
mediremi: Not since November 2019 :stuck_out_tongue_winking_eye:  (https://github.com/dhis2/notes/pull/45/files & https://github.com/dhis2/app-platform/blob/master/cli/config/.browserlistrc.js#L5)

Mohammer5: Oh that's nice. Wondering about opera mini (I think @amcgee or @ismay mentioned that it's being used in Africa quite a bit). But I assume this app won't be used on mobile devices for a LONG time, so I'll use `Object.fromEntries`

components/organisation-unit-tree/src/helpers/is-path-included.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @param {string[]} includedPaths
+ * @param {string} path
+ * @returns {bool}
+ */
+export const isPathIncluded = (includedPaths, path) => {
+    const isIncluded = includedPaths.some(includedPath => {
+        if (path === includedPath) return true
+        return includedPath.indexOf(`${path}/`) === 0
```
mediremi, line 9:
```suggestion
        return includedPath.startsWith(`${path}/`)
```

components/organisation-unit-tree/src/helpers/sort-node-children-alphabetically.js
```diff
@@ -0,0 +1,14 @@
+export const sortNodeChildrenAlphabetically = node => {
+    if (!node.children) return node
+    const sortedChildren = [...node.children]
+
+    sortedChildren.sort((left, right) => {
+        if (left.displayName === right.displayName) return 0
+        return left.displayName > right.displayName ? 1 : -1
```
mediremi, line 7:
```suggestion
        return left.displayName.localeCompare(right.displayName)
```

components/organisation-unit-tree/src/organisation-unit-node/use-org-data/use-org-data.js
```diff
@@ -0,0 +1,70 @@
...
+import { sortNodeChildrenAlphabetically } from '../../helpers/index.js'
+
+const ORG_DATA_QUERY = {
+    orgUnit: {
+        resource: `organisationUnits`,
+        id: ({ id }) => id,
+        params: ({ isUserDataViewFallback }) => ({
+            isUserDataViewFallback,
+            fields: ['id', 'path', 'children[id,path,displayName]'],
+            paging: false,
```
mediremi, line 12: Since this is a query for a singular org unit is `paging: false` needed?

components/organisation-unit-tree/src/organisation-unit-tree/default-render-node-label/label/label.js
```diff
@@ -0,0 +1,134 @@
...
+                {children}
+            </IconizedCheckbox>
+        </LabelContainer>
+    )
+}
+
+Label.propTypes = {
+    children: propTypes.any.isRequired,
+    dataTest: propTypes.string.isRequired,
+    // This is `any` so it can be customized by the app
```
mediremi, line 109: looks like the `propType` that this comment is associated with was moved (by eslint?) but the comment was not moved with it

components/organisation-unit-tree/src/organisation-unit-tree/organisation-unit-tree.js
```diff
@@ -43,10 +49,16 @@ const OrganisationUnitTree = ({
...
-        if (refetch && reloadId > 0) {
+        if (refetch && reloadId > 0 && reloadId !== prevReloadId) {
             refetch()
+            setPrevReloadId(reloadId)
         }
-    }, [reloadId, refetch])
+
+        return () =>
+            console.warn(
+                '@TODO: Why does this component unmount after a force reload?'
```
mediremi, line 59: An unmount would be if the second arg of `useEffect` was just `[]` - this callback was triggered as a result of `reloadId`/`prevReloadId` changing.

components/organisation-unit-tree/src/organisation-unit-node/use-org-data/use-org-data.js
Mohammer5: You're right, can be omitted. It's a leftover from when this hook was used for two different use cases

components/organisation-unit-tree/src/organisation-unit-tree/organisation-unit-tree.js
Mohammer5: I'm not sure I can follow.
Here's what I observed: The reload id does not change, so in theory this should work.
But for some reason the whole org unit tree component unmounts when the `forceReload` prop changes (or at least the returned callback gets called), which causes the hook to be run again..

mediremi: What I meant is that

```js
useEffect(() => {
  return () => {
    console.log("unmount")
  }
}, [])
```

would signal an unmount, but adding this to the component does not result in anything being logged to the console - so the `console.warn` in the `forceReaload` `useEffect` is being triggered due to a change in either `reloadId`, `prevReloadId` or `refetch`.

After investigating this a but further, I've discovered that the changes are being triggered by `refetch`, whose value changes due to `onComplete` and `onError` in `useRootOrgData` changing (see https://github.com/dhis2/app-runtime/blob/master/services/data/src/react/hooks/useQueryExecutor.ts#L44-L85 and https://github.com/dhis2/ui/blob/aad22793676c10e4dbcb103fcbd4ba930a04b7f4/components/organisation-unit-tree/src/organisation-unit-tree/use-root-org-data/use-root-org-data.js#L38-L52).

One way to prevent this issue is by wrapping `onComplete` and `onError` with `useCallback`:

```js
{
onComplete: useCallback(() => {
  // ...
}, [setState]),
onError: useCallback(() => {
  // ...
}, [setState]),
}
```

mediremi (approved):

components/organisation-unit-tree/src/organisation-unit-tree/organisation-unit-tree.js
Mohammer5: We agreed that this will be solved in #681

PR #681, opened 2021-08-03
feat(org unit tree): allow request customization

Mohammer5:


components/organisation-unit-tree/src/helpers/use-query.js
```diff
@@ -0,0 +1,54 @@
...
+        fetchFn(...args)
+            .then(response => {
+                const data = transform ? transform(response) : response
+                setState({
+                    ...state,
+                    loading: false,
+                    data: { ...defaultData, ...data },
+                })
+
+                return data
```
mediremi, line 28:
```suggestion
                return { ...defaultData, ...data }
```

components/organisation-unit-tree/src/organisation-unit-node/organisation-unit-node.js
```diff
@@ -52,14 +53,26 @@ export const OrganisationUnitNode = ({
...
-    const { loading, error, data } = useOrgData(id, {
-        isUserDataViewFallback,
-        suppressAlphabeticalSorting,
-        displayName,
+    const { loading, error, data } = useQuery(fetchOrgData, {
+        defaultData: {
+            organisationUnit: { id, displayName },
+        },
+        initialArguments: [{
+            variables: { id, isUserDataViewFallback },
```
mediremi, line 61: Does this need `paging: false`? Some other org unit node queries have it

Mohammer5: No, I've removed it from the other queries

components/organisation-unit-tree/src/__stories__/custom-node-label.js
```diff
@@ -0,0 +1,20 @@
...
+        renderNodeLabel={data => {
+            if (data.node.path !== '/A0000000000/A0000000001') {
+                return OrganisationUnitTree.defaultProps.renderNodeLabel(data)
+            }
+
+            return OrganisationUnitTree.defaultProps.renderNodeLabel({
+                ...data,
+                label: <span>--- {data.node.displayName}</span>,
+            })
+        }}
```
mediremi, line 18: Since using default props in this way is not a typical react pattern, I propose we expose `Label` as `OrganisationUnitLabel` in order to allow library consumers to do the following:

```js
        renderNodeLabel={props => {
            if (props.node.path !== '/A0000000000/A0000000001') {
                return <Label {...props}>{props.label}</Label>
            }

            return (
              <Label {...props}>
                Custom label for {props.node.displayName}
              </Label>
            )
        }}
```

mediremi: Alternatively we could make `renderNodeLabel` less flexible and use the return value as the `children` prop of the label:

```js
        renderNodeLabel={node => {
            if (node.path !== '/A0000000000/A0000000001') {
                return node.displayName
            }

            return (
              <strong>
                Custom label for <span>{props.node.displayName}</span>
              </strong>
            )
        }}
```

Mohammer5: I really like the idea of making it less flexible! WIll adapt that version

components/organisation-unit-tree/src/organisation-unit-tree/organisation-unit-tree.js
```diff
@@ -150,30 +174,23 @@ OrganisationUnitTree.propTypes = {
     /** When provided, the 'isUserDataViewFallback' option will be sent when requesting the org units */
     isUserDataViewFallback: propTypes.bool,
 
-    /** Renders the actual node component for each leaf, can be used to
-     * customize the node. In order to change the displayed node while keeping
-     * the existing functionality intact, you can re-use the original prop
-     * and overwrite the label property.
+    /** Renders the actual node label content for each leaf, can be used to
```
mediremi, line 177: JSDocs was replaced with react-docgen in https://github.com/dhis2/ui/pull/688, so this prop's description needs to be updated to:

```
    /** Renders the actual node component for each leaf, can be used to
     * customize the node. The default function just returns the node's
     * displayName
     * 
     * Shape of the object passed to the callback:
     * ```
     * {
     *    node: {
     *      displayName: string,
     *      id: string,
     *      // Only provided once `loading` is false
     *      path?: string,
     *      // Only provided once `loading` is false
     *      children?: Array.<{
     *        id: string,
     *        path: string,
     *        displayName: string
     *      }>
     *    },
     *    loading: boolean,
     *    error: string,
     *    open: boolean,
     *    selected: string[],
     *    singleSelection: boolean,
     *    disableSelection: boolean,
     *    // If the request is being customized, then all responses except the org
     *    // unit's response data will be included in this object
     *    additional?: object
     * }
     * ``` */
    renderNodeLabel: propTypes.func,
```

mediremi:
I'm converting this PR to a draft as after discussing with @HendrikThePendric we've agreed to implement these changes in the https://github.com/dhis2/approval-app repo.

Mohammer5:
keep it open

Mohammer5:
@HendrikThePendric is this PR of any relevance to you?

PR #684, opened 2021-08-05
chore(deps): update app-runtime to latest

ismay:
https://jira.dhis2.org/browse/LIBS-200, updates ui to app-runtime v3.

Follow up:

- [ ] Remove app-runtime resolution
- [ ] Update sharing dialog to new app-runtime patterns
- [ ] See if storybook webpack config can be manipulated by adding config instead of updating existing rules (and if that improves stability/separation between our and storybook's config)

### DIFF
--- a/components/organisation-unit-tree/src/organisation-unit-node/use-org-data/use-org-data.js
+++ b/components/organisation-unit-tree/src/organisation-unit-node/use-org-data/use-org-data.js
@@ -1,5 +1,5 @@
 import { useDataQuery } from '@dhis2/app-runtime'
-import { useState } from 'react'
+import { useMemo, useEffect } from 'react'
 import { sortNodeChildrenAlphabetically } from '../../helpers/index.js'
 
 const ORG_DATA_QUERY = {
@@ -35,35 +35,27 @@ export const useOrgData = (
 
     const defaultData = { id, displayName }
     const variables = { id, isUserDataViewFallback }
-    const [state, setState] = useState({
-        loading: true,
-        error: null,
-        data: defaultData,
-    })
-
-    useDataQuery(ORG_DATA_QUERY, {
+    const { loading, error, data } = useDataQuery(ORG_DATA_QUERY, {
         variables,
-        onComplete: ({ orgUnit: node }) => {
-            const sorted = !suppressAlphabeticalSorting
-                ? sortNodeChildrenAlphabetically(node)
-                : node
-
-            const nextData = { ...defaultData, ...sorted }
-            setState({
-                ...state,
-                data: nextData,
-                loading: false,
-            })
-            onComplete && onComplete(nextData)
-        },
-        onError: queryError => {
-            setState({
-                ...state,
-                error: queryError,
-                loading: false,
-            })
-        },
     })
 
-    return state
+    const transformedData = useMemo(() => {
+        if (!data) {
+            return defaultData
+        }
+
+        const { orgUnit: node } = data
+        const transformed = suppressAlphabeticalSorting
+            ? node
+            : sortNodeChildrenAlphabetically(node)
+        const merged = { ...defaultData, ...transformed }
+
+        return merged
+    }, [data, suppressAlphabeticalSorting, id, displayName])
+
+    useEffect(() => {
+        onComplete && onComplete(transformedData)
+    }, [onComplete, transformedData])
+
+    return { loading, error: error || null, data: transformedData }
 }
--- a/components/organisation-unit-tree/src/organisation-unit-tree/use-root-org-data/use-root-org-data.js
+++ b/components/organisation-unit-tree/src/organisation-unit-tree/use-root-org-data/use-root-org-data.js
@@ -1,5 +1,5 @@
 import { useDataQuery } from '@dhis2/app-runtime'
-import { useCallback, useState } from 'react'
+import { useMemo } from 'react'
 import { patchMissingDisplayName } from './patch-missing-display-name.js'
 
 export const createRootQuery = ids =>
@@ -29,37 +29,18 @@ export const createRootQuery = ids =>
 export const useRootOrgData = (ids, { isUserDataViewFallback } = {}) => {
     const query = createRootQuery(ids)
     const variables = { isUserDataViewFallback }
-    const [state, setState] = useState({
-        loading: true,
-        error: null,
-        data: null,
-    })
-
-    const { refetch: queryRefetch } = useDataQuery(query, {
+    const { loading, error, data, refetch } = useDataQuery(query, {
         variables,
-        onComplete: queryData =>
-            setState({
-                ...state,
-                data: queryData ? patchMissingDisplayName(queryData) : {},
-                loading: false,
-            }),
-        onError: queryError =>
-            setState({
-                ...state,
-                error: queryError,
-                loading: false,
-            }),
     })
 
-    const refetch = useCallback(() => {
-        setState({ ...state, data: null, error: null, loading: true })
-        queryRefetch()
-    }, [queryRefetch, setState])
+    const patchedData = useMemo(() => {
+        return data ? patchMissingDisplayName(data) : data
+    }, [data])
 
     return {
-        loading: state.loading,
-        error: state.error,
-        data: state.data,
+        loading,
+        error: error || null,
+        data: patchedData || null,
         refetch,
     }
 }
--- a/components/sharing-dialog/src/__tests__/ShareBlock.test.js
+++ b/components/sharing-dialog/src/__tests__/ShareBlock.test.js
@@ -1,5 +1,6 @@
 import { Button } from '@dhis2-ui/button'
-import { shallow } from 'enzyme'
+import { CustomDataProvider } from '@dhis2/app-runtime'
+import { mount } from 'enzyme'
 import React from 'react'
 import { AccessSelect } from '../access-select.js'
 import { Autocomplete } from '../autocomplete/autocomplete.js'
@@ -13,7 +14,9 @@ describe('SharingDialog widget - ShareBlock component', () => {
 
     const getShareBlockComponent = props => {
         if (!shallowShareBlockComponent) {
-            shallowShareBlockComponent = shallow(<ShareBlock {...props} />)
+            shallowShareBlockComponent = mount(<ShareBlock {...props} />, {
+                wrappingComponent: CustomDataProvider,
+            })
         }
 
         return shallowShareBlockComponent
--- a/components/sharing-dialog/src/__tests__/SharingDialog.test.js
+++ b/components/sharing-dialog/src/__tests__/SharingDialog.test.js
@@ -1,6 +1,7 @@
 import { Button } from '@dhis2-ui/button'
 import { Modal, ModalTitle } from '@dhis2-ui/modal'
-import { shallow } from 'enzyme'
+import { CustomDataProvider } from '@dhis2/app-runtime'
+import { mount } from 'enzyme'
 import React from 'react'
 import { DashboardSharingContent } from '../dashboard-sharing-content.js'
 import { DefaultSharingContent } from '../default-sharing-content.js'
@@ -17,7 +18,9 @@ describe('SharingDialog widget', () => {
 
     const getSharingDialogWidget = props => {
         if (!shallowSharingDialog) {
-            shallowSharingDialog = shallow(<SharingDialog {...props} />)
+            shallowSharingDialog = mount(<SharingDialog {...props} />, {
+                wrappingComponent: CustomDataProvider,
+            })
         }
 
         return shallowSharingDialog
@@ -34,8 +37,9 @@ describe('SharingDialog widget', () => {
         }
     })
 
-    it('renders a Modal', () =>
-        expect(getSharingDialogWidget(props).find(Modal)).toHaveLength(1))
+    it('renders a Modal', () => {
+        expect(getSharingDialogWidget(props).find(Modal)).toHaveLength(1)
+    })
 
     it('renders the ModalTitle with the name of the AO if passed in props', () => {
         props.initialSharingSettings = {
@@ -49,16 +53,21 @@ describe('SharingDialog widget', () => {
     })
 
     it('renders a Close button', () => {
-        const button = getSharingDialogWidget(props).find(Button)
+        const buttons = getSharingDialogWidget(props).find(Button)
+        const closeButtons = buttons.filterWhere(button =>
+            button.html().includes('Close')
+        )
 
-        expect(button).toHaveLength(1)
-        expect(button.html()).toMatch('Close')
+        expect(closeButtons).toHaveLength(1)
     })
 
     it('calls the Close callback when the Close button is clicked', () => {
-        const button = getSharingDialogWidget(props).find(Button)
+        const buttons = getSharingDialogWidget(props).find(Button)
+        const closeButtons = buttons.filterWhere(button =>
+            button.html().includes('Close')
+        )
 
-        button.simulate('click')
+        closeButtons.first().simulate('click')
 
         expect(onClose).toHaveBeenCalled()
     })
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
         "react": "16.13",
         "react-dev-utils": "^10.2.1",
         "react-dom": "16.13",
+        "react-scripts": "^4.0.3",
         "rimraf": "^3.0.2",
         "styled-jsx": "^4.0.1",
         "wait-on": "^5.3.0"
--- a/package.json
+++ b/package.json
@@ -45,8 +45,11 @@
             "description": "Components and related resources for the DHIS2 design system"
         }
     },
+    "resolutions": {
+        "@dhis2/app-runtime": "^3.0.0"
+    },
     "dependencies": {
-        "@dhis2/app-runtime": "^2.11.0",
+        "@dhis2/app-runtime": "^3.0.0",
         "@dhis2/cli-app-scripts": "^7.3.0",
         "@dhis2/cli-style": "^9.0.1",
         "@dhis2/cli-utils-cypress": "^8.0.1",
--- a/collections/ui/package.json
+++ b/collections/ui/package.json
@@ -67,14 +67,14 @@
         "prop-types": "^15.7.2"
     },
     "peerDependencies": {
-        "@dhis2/app-runtime": "^2",
+        "@dhis2/app-runtime": "^3",
         "@dhis2/d2-i18n": "^1",
         "react": "^16.8",
         "react-dom": "^16.8",
         "styled-jsx": "^4"
     },
     "devDependencies": {
-        "@dhis2/app-runtime": "^2.11.0",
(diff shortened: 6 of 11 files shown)


varl:
A neat trick to update `app-runtime` in all packages automatically is to bump the `app-runtime` dep in the root `package.json` to the latest:

```
    "dependencies": {
        "@dhis2/app-runtime": "^3.1.0",
```

And then run:

```
yarn manypkg fix
```

That will see that the highest mentioned version is `3.1.0` and update everywhere where it is lower:

```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   collections/ui/package.json
        modified:   collections/widgets/package.json
        modified:   components/header-bar/package.json
        modified:   components/organisation-unit-tree/package.json
        modified:   components/sharing-dialog/package.json
        modified:   package.json
        modified:   yarn.lock
```

Sadly it only updated the `devDeps`, and running `yarn manypkg check` again gives errors:

```
yarn manypkg check
yarn run v1.22.5
$ /home/varl/dev/dhis2/libs/ui/node_modules/.bin/manypkg check
☔️ error @dhis2/ui has a peerDependency on @dhis2/app-runtime but the range specified in devDependency is not greater than or equal to the range specified in peerDependencies
☔️ error @dhis2/ui-widgets has a peerDependency on @dhis2/app-runtime but the range specified in devDependency is not greater than or equal to the range specified in peerDependencies
☔️ error @dhis2-ui/header-bar has a peerDependency on @dhis2/app-runtime but the range specified in devDependency is not greater than or equal to the range specified in peerDependencies
☔️ error @dhis2-ui/organisation-unit-tree has a peerDependency on @dhis2/app-runtime but the range specified in devDependency is not greater than or equal to the range specified in peerDependencies
☔️ error @dhis2-ui/sharing-dialog has a peerDependency on @dhis2/app-runtime but the range specified in devDependency is not greater than or equal to the range specified in peerDependencies
☔️ info the above errors may be fixable with yarn manypkg fix
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

But, at least there is an exhaustive list where it needs to be fixed.

ismay:
> A neat trick to update `app-runtime` in all packages automatically is to bump the `app-runtime` dep in the root `package.json` to the latest:
> 
> ```
>     "dependencies": {
>         "@dhis2/app-runtime": "^3.1.0",
> ```
> 
> And then run:
> 
> ```
> yarn manypkg fix
> ```
> 
> That will see that the highest mentioned version is `3.1.0` and update everywhere where it is lower:
> 
> ```
> Changes not staged for commit:
>   (use "git add <file>..." to update what will be committed)
>   (use "git restore <file>..." to discard changes in working directory)
>         modified:   collections/ui/package.json
>         modified:   collections/widgets/package.json
>         modified:   components/header-bar/package.json
>         modified:   components/organisation-unit-tree/package.json
>         modified:   components/sharing-dialog/package.json
>         modified:   package.json
>         modified:   yarn.lock
> ```
> 
> Sadly it only updated the `devDeps`, and running `yarn manypkg check` again gives errors:
> 
> ```
> yarn manypkg check
> yarn run v1.22.5
> $ /home/varl/dev/dhis2/libs/ui/node_modules/.bin/manypkg check
> ☔️ error @dhis2/ui has a peerDependency on @dhis2/app-runtime but the range specified in devDependency is not greater than or equal to the range specified in peerDependencies
> ☔️ error @dhis2/ui-widgets has a peerDependency on @dhis2/app-runtime but the range specified in devDependency is not greater than or equal to the range specified in peerDependencies
> ☔️ error @dhis2-ui/header-bar has a peerDependency on @dhis2/app-runtime but the range specified in devDependency is not greater than or equal to the range specified in peerDependencies
> ☔️ error @dhis2-ui/organisation-unit-tree has a peerDependency on @dhis2/app-runtime but the range specified in devDependency is not greater than or equal to the range specified in peerDependencies
> ☔️ error @dhis2-ui/sharing-dialog has a peerDependency on @dhis2/app-runtime but the range specified in devDependency is not greater than or equal to the range specified in peerDependencies
> ☔️ info the above errors may be fixable with yarn manypkg fix
> error Command failed with exit code 1.
> info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
> ```
> 
> But, at least there is an exhaustive list where it needs to be fixed.

Ah nice. Weird that rg missed a couple packages. The react-scripts error is weird as well. If I install it in the root everything works just fine locally. If I remove it it doesn't seem to get hoisted from the app-shell. No idea why that's suddenly happening.

ismay:
Ok, I suspect it has something to do with our custom storybook loading. Part of the error is that there's no matching rule that matches the `find` here: https://github.com/dhis2/ui/blob/69334aa65c11feb39c74a0b6eab2b0042086f0de/storybook/src/webpack-config.js#L21

varl:
I'm not seeing any failure to resolve `react-scripts` locally on commit 6b1caa7 (before you added `react-scripts` to the dependencies) ?

I'm getting test failures on a `mock` and no `QueryClientProvider` ?

ismay:
> I'm not seeing any failure to resolve `react-scripts` locally on commit [6b1caa7](https://github.com/dhis2/ui/commit/6b1caa733c3f125850ff8fcf7af5bf6aa4c6956d) (before you added `react-scripts` to the dependencies) ?
> 
> I'm getting test failures on a `mock` and no `QueryClientProvider` ?

See here: https://github.com/dhis2/ui/pull/684/checks?check_run_id=3607978897#step:6:26. That's what I'm getting basically.

ismay:
I've tried a fresh clone, yarn install, yarn build and yarn setup. After that yarn start produces the above error for me. So Has been reproducible for me so far.

components/organisation-unit-tree/src/organisation-unit-node/use-org-data/use-org-data.js
```diff
@@ -35,35 +35,25 @@ export const useOrgData = (
...
+        const { orgUnit: node } = data
+        const transformed = suppressAlphabeticalSorting ? node : sortNodeChildrenAlphabetically(node)
+        const merged = { ...defaultData, ...transformed }
+
+        return merged
+    }, [data, suppressAlphabeticalSorting, id, displayName])
+
+    useEffect(() => {
+        onComplete && onComplete(transformedData)
+    }, [onComplete, transformedData])
```
ismay, line 56: I'm using a useEffect here instead of including it in the useMemo as it seems more suited to a side-effect like this, and not _directly_ related to the memoized value. Seems better to leave this to react's scheduling.

components/organisation-unit-tree/src/organisation-unit-node/use-org-data/use-org-data.js
```diff
@@ -35,35 +35,25 @@ export const useOrgData = (
 
     const defaultData = { id, displayName }
     const variables = { id, isUserDataViewFallback }
-    const [state, setState] = useState({
-        loading: true,
-        error: null,
-        data: defaultData,
+    const { loading, error, data } = useDataQuery(ORG_DATA_QUERY, {
```
ismay, line 38: Using the useDataQuery's values directly instead of with the intermediate state.

components/organisation-unit-tree/src/organisation-unit-node/use-org-data/use-org-data.js
```diff
@@ -35,35 +35,25 @@ export const useOrgData = (
...
-                loading: false,
-            })
-        },
-    })
+        const { orgUnit: node } = data
+        const transformed = suppressAlphabeticalSorting ? node : sortNodeChildrenAlphabetically(node)
+        const merged = { ...defaultData, ...transformed }
+
+        return merged
+    }, [data, suppressAlphabeticalSorting, id, displayName])
```
ismay, line 52: Using a useMemo to prevent unnecessary calculations if the data hasn't changed, without having to resort to setState.

components/organisation-unit-tree/src/organisation-unit-node/use-org-data/use-org-data.js
```diff
@@ -35,35 +35,25 @@ export const useOrgData = (
...
+
+    useEffect(() => {
+        onComplete && onComplete(transformedData)
+    }, [onComplete, transformedData])
 
-    return state
+    return { loading, error: error || null, data: transformedData }
```
ismay, line 58: The unit test expected error or null, instead of the error or undefined that the app-runtime returns. I kept it to the existing API for this component.

components/sharing-dialog/src/__tests__/ShareBlock.test.js
```diff
@@ -13,7 +14,9 @@ describe('SharingDialog widget - ShareBlock component', () => {
 
     const getShareBlockComponent = props => {
         if (!shallowShareBlockComponent) {
-            shallowShareBlockComponent = shallow(<ShareBlock {...props} />)
+            shallowShareBlockComponent = mount(<ShareBlock {...props} />, {
+                wrappingComponent: CustomDataProvider,
+            })
```
ismay, line 19: It seems to me that this should also work with shallow. It also has the wrappingComponent API. But I'm a react-query missing provider error, so I went with mount. Costs a little more time to run.

components/sharing-dialog/src/__tests__/SharingDialog.test.js
```diff
@@ -49,16 +53,17 @@ describe('SharingDialog widget', () => {
     })
 
     it('renders a Close button', () => {
-        const button = getSharingDialogWidget(props).find(Button)
+        const buttons = getSharingDialogWidget(props).find(Button)
+        const closeButtons = buttons.filterWhere(button => button.html().includes('Close'))
 
-        expect(button).toHaveLength(1)
-        expect(button.html()).toMatch('Close')
+        expect(closeButtons).toHaveLength(1)
```
ismay, line 59: Since we're mounting it finds more components than it did with shallow, so I have to do some extra filtering here and in the test below.

components/sharing-dialog/src/__tests__/SharingDialog.test.js
```diff
@@ -17,7 +18,9 @@ describe('SharingDialog widget', () => {
 
     const getSharingDialogWidget = props => {
```
ismay, line 19: Sidenote: these abstractions make this test harder to work on in my opinion. Useful for app code, not as useful for tests, that are now all coupled and harder to understand and refactor. Same goes for the shareblock test above.

components/organisation-unit-tree/src/organisation-unit-tree/use-root-org-data/use-root-org-data.js
```diff
@@ -29,37 +29,18 @@ export const createRootQuery = ids =>
...
 
     return {
-        loading: state.loading,
-        error: state.error,
-        data: state.data,
+        loading,
+        error: error || null,
+        data: patchedData || null,
         refetch,
     }
```
ismay, line 45: Eventually we should refactor this a little more in depth, so that the hook just passes around the default values that the app-runtime uses as well. Might be feasible to do that in this PR, didn't look into it (also goes for other areas where I'm sticking closely to the existing API of the refactored components for safety).

ismay:
@varl Btw, I rebased on next and now the react-scripts error is back. Even without a cli-app-scripts update.

ismay:
> @varl Btw, I rebased on next and now the react-scripts error is back. Even without a cli-app-scripts update.

So I'm not sure _why_ it's changed, but it does seem like the errors are related to the storybook webpack changing slightly, and our modifications to that config are failing because they're quite tied to the exact shape of the config. I've gotten the build to progress beyond the errors by tweaking webpack-config.js, but haven't gotten it to completely behave. That does seem to at least be part of the problem though.

varl:
> > @varl Btw, I rebased on next and now the react-scripts error is back. Even without a cli-app-scripts update.
> 
> So I'm not sure _why_ it's changed, but it does seem like the errors are related to the storybook webpack changing slightly, and our modifications to that config are failing because they're quite tied to the exact shape of the config. I've gotten the build to progress beyond the errors by tweaking webpack-config.js, but haven't gotten it to completely behave. That does seem to at least be part of the problem though.

I diffed them manually yesterday, and I get different Webpack configuration files when `react-script` resolves, and when it doesn't.

Storybook's `--debug-webpack` helps a lot.

ismay:
> > > @varl Btw, I rebased on next and now the react-scripts error is back. Even without a cli-app-scripts update.
> > 
> > 
> > So I'm not sure _why_ it's changed, but it does seem like the errors are related to the storybook webpack changing slightly, and our modifications to that config are failing because they're quite tied to the exact shape of the config. I've gotten the build to progress beyond the errors by tweaking webpack-config.js, but haven't gotten it to completely behave. That does seem to at least be part of the problem though.
> 
> I diffed them manually yesterday, and I get different Webpack configuration files when `react-script` resolves, and when it doesn't.
> 
> Storybook's `--debug-webpack` helps a lot.

Yeah exactly. I still don't get why that has changed.

varl:
> Yeah exactly. I still don't get why that has changed.

My hypothesis is that `react-scripts` pulls in their own Webpack configuration that's different from the default Storybook one that's used otherwise.

ismay:
> > Yeah exactly. I still don't get why that has changed.
> 
> My hypothesis is that `react-scripts` pulls in their own Webpack configuration that's different from the default Storybook one that's used otherwise.
> 
> Or do you mean why it doesn't resolve anymore ? That frustrates my brain as well.

I mean why it doesn't resolve anymore yeah. I think it's just yarn deciding that it's not necessary all of a sudden. Because it no longer hoists it to the top. This is what happens on the PR branch:

Screenshot 2021-09-16 at 11 30 34

And this is what happens on master:

Screenshot 2021-09-16 at 11 31 26

ismay:
> > > Yeah exactly. I still don't get why that has changed.
> > 
> > 
> > My hypothesis is that `react-scripts` pulls in their own Webpack configuration that's different from the default Storybook one that's used otherwise.
> > Or do you mean why it doesn't resolve anymore ? That frustrates my brain as well.
> 
> I mean why it doesn't resolve anymore yeah. I think it's just yarn deciding that it's not necessary all of a sudden. Because it no longer hoists it to the top. This is what happens on the PR branch:
> 
> Screenshot 2021-09-16 at 11 30 34
> 
> And this is what happens on master:
> 
> Screenshot 2021-09-16 at 11 31 26

So it makes sense that it can no longer resolve it.

varl:
> > > > Yeah exactly. I still don't get why that has changed.
> > > 
> > > 
> > > My hypothesis is that `react-scripts` pulls in their own Webpack configuration that's different from the default Storybook one that's used otherwise.
> > > Or do you mean why it doesn't resolve anymore ? That frustrates my brain as well.
> > 
> > 
> > I mean why it doesn't resolve anymore yeah. I think it's just yarn deciding that it's not necessary all of a sudden. Because it no longer hoists it to the top. This is what happens on the PR branch:
> > Screenshot 2021-09-16 at 11 30 34
> > And this is what happens on master:
> > Screenshot 2021-09-16 at 11 31 26
> 
> So it makes sense that it can no longer resolve it. In a way what we're doing is probably outside of the intended usecase for storybook's preset-create-react-app (assuming that that's where the resolution is failing). We're not using CRA directly, so their expectations aren't met when it comes to resolving react-scripts. Maybe we were just lucky that yarn hoisted react-scripts up until now.

Might be a blessing in a horrible disguise. I see that `react-scripts` is pulled from `app-shell`, which I really don't think is ideal for UI.

We can put `react-scripts` as a dependency of the Storybook workspace, and that should allow it to resolve correctly -- and give us a more explicit relationship to the Webpack config that we need to ~screw around with~ delicately modify.

```yarn workspace ui-storybook add react-scripts```

**Comment by ismay** [comment]
> > So it makes sense that it can no longer resolve it. In a way what we're doing is probably outside of the intended usecase for storybook's preset-create-react-app (assuming that that's where the resolution is failing). We're not using CRA directly, so their expectations aren't met when it comes to resolving react-scripts. Maybe we were just lucky that yarn hoisted react-scripts up until now.
> 
> Might be a blessing in a horrible disguise. I see that `react-scripts` is pulled from `app-shell`, which I really don't think is ideal for UI.
> 
> We can put `react-scripts` as a dependency of the Storybook workspace, and that should allow it to resolve correctly -- and give us a more explicit relationship to the Webpack config that we need to ~screw around with~ delicately modify.
> 
> `yarn workspace ui-storybook add react-scripts`

Yeah that's probably the best solution. I looked at removing preset-create-react-app but that seems more involved. One thing I was wondering, would it not be more stable for us to push extra rules to the config, instead of modifying existing ones?

We wouldn't inherit any changes from the storybook webpack config, so there's chance of the configs diverging, but it would also decouple it a bit. Might be a good thing.

**Comment by varl** [comment]
> We wouldn't inherit any changes from the storybook webpack config, so there's chance of the configs diverging, but it would also decouple it a bit. Might be a good thing.

Yeah. I think so. Maybe even making our Webpack configuration in app-scripts shared and installable as a package so we can depend on that and inject it as is into Storybook.

**Comment by ismay** [comment]
> > We wouldn't inherit any changes from the storybook webpack config, so there's chance of the configs diverging, but it would also decouple it a bit. Might be a good thing.
> 
> Yeah. I think so. Maybe even making our Webpack configuration in app-scripts shared and installable as a package so we can depend on that and inject it as is into Storybook.
> 
> Separate issue and PR ?

Yeah, will open one.

By the by, react-scripts only resolves successfully if added to the workspace root, so I've gone ahead and done that.

File: package.json [hunk 9]
```diff
@@ -66,6 +69,7 @@
         "react": "16.13",
         "react-dev-utils": "^10.2.1",
         "react-dom": "16.13",
+        "react-scripts": "^4.0.3",
```
varl, line 72: Can we add this to `storybook/package.json` instead ?

varl: Ok, saw the other comment now.

ismay: Yeah I've tried. Still doesn't resolve successfully then.

ismay:
> Changes look good to me!
> 
> OT: Hmm no cypress tests for the sharing dialog.. Not sure if that's because there was no need of if they have been omitted for other reasons, but it leaves a bit of a bitter taste.. Does it still work as intended (can the answer be inferred from passing tests or does this have to be done manually)?

Yeah I'm not sure tbh. I don't know the component very well, so I've made the changes I've made with very limited knowledge of the inner workings. I did check out the storybook, and it seems all right, but it's probably best if someone with more intimate knowledge of the component double checks it.

Mohammer5:
CC @edoardo

varl:
Cypress tests will need to be written in post, as the lack of them couldn't block 2.37, so this was component was intentionally merged without them. We should have a Jira ticket for that.

Mohammer5:
Should we then defer this PR? Or go ahead and follow-up with cypress tests?

varl:
No, we cannot defer this PR because `master` is currently blocked, so we need to move on this to `next` and then `next` to master, and then we can create a `6.x` branch to allow issues blocking 2.37 to move forward.

I know this isn't an ideal flow.

PR #686, opened 2021-08-09
feat: use filter in query params and a single get request

HendrikThePendric:
Still need to update the tests for this, but it seems to be working fine. Before we merge this in we need to check with the backend team about the possible performance implications.


mediremi (approved):

HendrikThePendric:
Not relevant

PR #687, opened 2021-08-09
test: add test script to each component's package.json file

mediremi:
Adds a test script to each component's `package.json` file, allowing us to run `yarn test` for individual components.

Also adds jest enzyme config for components whose tests use enzyme.


varl (approved):
Nice.

Would have been nifty if we could have passed in e.g. `d2-app-scripts test --jest-config ../../jest.config.js` and re-used the Jest and Enzyme configuration (the `--config` flag is [taken](https://platform.dhis2.nu/#/scripts/test)).

That can be a future thing.

mediremi:
I've implemented the `jestConfig` option here: https://github.com/dhis2/app-platform/pull/605

Unfortunately we can't reuse the top level `jest.config.js` for running individual components' tests due to the `roots` config and the fact that the current working directory in the config file will be the component directory so we'll need this additional config in a file called something like `jest.shared.config.js`:

```js
module.exports = {
    setupFilesAfterEnv: [`${__dirname}/jest.enzyme.config.js`],
    testPathIgnorePatterns: [
        '/node_modules/',
        '/build/',
        '/.d2/',
    ],
}
```

I'm happy to update this PR once https://github.com/dhis2/app-platform/pull/605 is merged.

jest.config.shared.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+    setupFilesAfterEnv: [`${__dirname}/jest.enzyme.config.js`],
```
varl, line 2: Neato !

varl (approved):

PR #688, opened 2021-08-09
docs: use react-docgen

mediremi:
This PR:
1) Fixes generation of our documentation which was previously broken due to babel issues
2) Uses react-docgen to generate better component documentation


KaiVandivier (approved):
Looks good!

PR #697, opened 2021-08-10
docs(constants): fix link to elevation in design system

mediremi:


KaiVandivier (approved):
Nice, thanks 😁 We're getting closer and closer to good docs

PR #711, opened 2021-08-23
feat(portal): introduce component

varl:
Introduces a new Portal component that can be used as a standard React
component, either with a default or custom mount point.

Nested Modals in JSX is now an unsupported technique. Modals can be
siblings, but not placed within each other.

If a modal needs to create another modal, then a wrapper component that
controls which of the sibling modals should open and when it should be
used. Modals placed as siblings will render on top of one another based
on the order in the DOM, so last sibling renders on top of the
preceeding ones.

### DIFF
--- a/components/alert/package.json
+++ b/components/alert/package.json
@@ -31,6 +31,7 @@
         "styled-jsx": "^3.2"
     },
     "dependencies": {
+        "@dhis2-ui/portal": "6.17.0",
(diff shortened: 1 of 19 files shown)


varl:
Supersedes: https://github.com/dhis2/ui/pull/474

varl:
Depends on: https://github.com/dhis2/app-platform/pull/613

HendrikThePendric (approved):

PR #713, opened 2021-08-24
fix(button): remove z-index from border style

varl:
Modifying the z-index on the border causes it to be shown above the
normal stacking context, which causes it to overlay on top of e.g.
modals and component covers.

Generally we should avoid modifying the z-index of components and allow
them to be automatically put into the relevant stacking context.

### DIFF
--- a/components/button/src/button/button.styles.js
+++ b/components/button/src/button/button.styles.js
@@ -46,7 +46,6 @@ export default css`
         right: -2px;
         bottom: -2px;
         left: -2px;
-        z-index: 1;
 
         border: 2px solid transparent;
         border-radius: inherit;
--- a/components/button/src/split-button/split-button.js
+++ b/components/button/src/split-button/split-button.js
@@ -114,9 +114,11 @@ class SplitButton extends Component {
                 <style jsx>{`
                     div {
                         display: inline-flex;
-                        position: relative;
                         color: inherit;
                         white-space: nowrap;
+                        // create a stacking context for the children
+                        position: relative;
+                        z-index: 0;
                     }
 
                     div > :global(button:first-child) {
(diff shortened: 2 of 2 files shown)


edoardo (approved):

HendrikThePendric (commented):
I'm not sure if this is a reason for not implementing this fix, but I just wanted to point out that we do have some visual regressions as a result of this change. For example, the split button now looks like this:

Screenshot 2021-08-24 at 15 04 03

Some possible ideas on how to address this:
- A layering system which uses z-index
- Provide a z-index to the portal root so it is always on top of whatever is in the app root
- Changing the button styles so that the focus outline doesn't have a greater with than the border.

HendrikThePendric:
FYI: I briefly looked into this and I believe the core of the problem is the button style specifications. Because we want the "focus outline" to effectively extend 1 px outside of the button element things become problematic. 

I expect we also opted for a slightly quirky solution using the `::after` pseudo element was also a result of this.

My recommendation would be to adjust the button styles and use the `outline` CSS attribute to produce a 2px border with -2px offset.

HendrikThePendric:
@varl I've pushed 2 commits that "pull the focus-outline" into the button. Feel free to revert them if you think this is not a good solution. Without 78d4650 the split button's focus styles don't look too bad either, the right part is just has 1 visible px width instead of 2.

varl:
Thanks @HendrikThePendric. I would like to avoid changing the Button styles. I have a fix for it that moves the `z-index` stuff to the SplitButton that also works with modals.

2021-08-24-161208_657x309_scrot

HendrikThePendric (approved):
Makes a lot of sense to solve it like this for now.

PR #720, opened 2021-08-31
feat: add 'last online' info to status indicator

KaiVandivier:
Part of https://jira.dhis2.org/browse/LIBS-48

Applying this feature will also require some changes to the app platform:
1. to read a configuration option in `d2.config.js` that decides what kind of additional text to add to the online status indicator -- I'm thinking either `headerbar.onlineStatusInfo` or `pwa.headerbarStatusInfo` for the config property, but I'd like a second opinion 🧐 
2. and add that option to the app config through the shell and adapter

### DIFF
--- a/components/header-bar/src/header-bar.stories.e2e.js
+++ b/components/header-bar/src/header-bar.stories.e2e.js
@@ -23,3 +23,15 @@ export const PWAEnabled = () => (
         <HeaderBar appName="Example!" />
     </Provider>
 )
+
+export const WithLastOnline = () => (
+    <Provider
+        config={{
+            ...config,
+            pwaEnabled: true,
+            headerbar: { onlineStatusInfo: 'LAST_ONLINE' },
+        }}
+    >
+        <HeaderBar appName="Example!" />
+    </Provider>
+)
--- a/components/header-bar/src/header-bar.stories.e2e.js
+++ b/components/header-bar/src/header-bar.stories.e2e.js
@@ -18,6 +18,12 @@ export const Default = () => (
     </Provider>
 )
 
+export const ShowOnlineStatus = () => (
+    <Provider config={{ ...config, headerbar: { showOnlineStatus: true } }}>
+        <HeaderBar appName="Example!" />
+    </Provider>
+)
+
 export const PWAEnabled = () => (
     <Provider config={{ ...config, pwaEnabled: true }}>
         <HeaderBar appName="Example!" />
@@ -28,8 +34,10 @@ export const WithLastOnline = () => (
     <Provider
         config={{
             ...config,
-            pwaEnabled: true,
-            headerbar: { onlineStatusInfo: 'LAST_ONLINE' },
+            headerbar: {
+                showOnlineStatus: true,
+                onlineStatusInfo: 'LAST_ONLINE',
+            },
         }}
     >
         <HeaderBar appName="Example!" />
--- a/components/header-bar/src/online-status.js
+++ b/components/header-bar/src/online-status.js
@@ -1,13 +1,33 @@
-import { useOnlineStatus } from '@dhis2/app-runtime'
+import { useConfig, useOnlineStatus } from '@dhis2/app-runtime'
 import cx from 'classnames'
+import moment from 'moment'
 import PropTypes from 'prop-types'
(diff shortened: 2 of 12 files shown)


yarn.lock
```diff
@@ -1690,6 +1690,16 @@
   dependencies:
     moment "^2.24.0"
 
+"@dhis2/app-runtime@^2.11.0":
```
mediremi, line 1693: Let's deduplicate the lockfile with `npx yarn-deduplicate yarn.lock` as updating `app-runtime` has resulted in multiple versions of services.

components/header-bar/src/online-status.js
```diff
@@ -1,15 +1,35 @@
...
 /** A badge to display online/offline status in the header bar */
-export function OnlineStatus({ info, dense }) {
-    const { online } = useOnlineStatus()
+export function OnlineStatus({ dense }) {
+    const { online, lastOnline } = useOnlineStatus()
+    const { headerbar } = useConfig()
+
     const displayStatus = online ? i18n.t('Online') : i18n.t('Offline')
 
+    let info
```
mediremi, line 16: Extracting this logic into a hook is a bit cleaner as there's no need for mutable variables:

```js
const useOnlineStatusInfo = ({ online, lastOnline }) => {
    const { headerbar } = useConfig()
    switch (headerbar?.onlineStatusInfo) {
        case 'last-online': 
            if (!online && lastOnline) {
                 return i18n.t('Last online {{relativeTime}}', {
                     relativeTime: moment(lastOnline).fromNow(),
                 })
            } else {
                return null
            }
        default:
            return null
    }
    return null
}
```

```js
const info = useOnlineStatusInfo()
```

As there's currently only one supported value for `headerbar.onlineStatusInfo`, I'd even be inclined to do:

```js
const useOnlineStatusInfo = ({ online, lastOnline }) => {
    const { headerbar } = useConfig()
    if (headerbar?.onlineStatusInfo === 'last-online' && !online && lastOnline) {
        return i18n.t('Last online {{relativeTime}}', {
            relativeTime: moment(lastOnline).fromNow(),
        })
    }
    return null
}
```

components/header-bar/src/header-bar.stories.e2e.js
```diff
@@ -23,3 +23,15 @@ export const PWAEnabled = () => (
         <HeaderBar appName="Example!" />
     </Provider>
 )
+
+export const WithLastOnline = () => (
+    <Provider
+        config={{
+            ...config,
+            pwaEnabled: true,
+            headerbar: { onlineStatusInfo: 'last-online' },
```
mediremi, line 32: Using uppercase snake case (`LAST_ONLINE`) would make it clearer that this value is an enum and won't be rendered directly.

yarn.lock
KaiVandivier: Interesting - this has been happening everywhere I've update the runtime recently

components/header-bar/src/online-status.js
KaiVandivier: That looks clean 👍

yarn.lock
mediremi: Looks like there's a conflict with `master` now. What I usually do in this situation is:

1. Delete the `yarn.lock` in this branch and replace it with the one from `master`
2. Run `yarn install`
3. Run `npx yarn-deduplicate yarn.lock`

KaiVandivier: Luckily there are only a few changes to yarn.lock so it's easy to fix the merge conflicts, but this seems like a useful tip for the future. Do you commit after step 1, then again after step 3? Btw after running `npx yarn-deduplicate yarn.lock` I ended up with the old 2.9.2 versions of all the runtime services still in yarn.lock, and running `yarn` again cleared those

mediremi: I do `git merge master` first (which fails because of the merge conflict), do steps 1 to 3, and then do `git commit` to commit the merge. So only one commit needed in total.

mediremi: Just tried locally and yep for some reason yarn dedup is incorrectly adding the old versions back so looks like this time there's a step 4 of `yarn install`.

mediremi (approved):

PR #725, opened 2021-09-01
feat(tab bar): add "autoHideScrollButtons" functionality

Mohammer5:
Closes LIBS-211
([link](https://jira.dhis2.org/browse/LIBS-211))

- [x] Add functionality
- [x] Add story
- [x] Add tests
- [x] ~~Add docs~~ (there are no docs for this component that could be updated)
- [x] Discuss: Should auto hiding be the default functionality? (Yes, can't be turned off, because why?)

### DIFF
--- a/components/tab/src/tab-bar/scroll-bar.js
+++ b/components/tab/src/tab-bar/scroll-bar.js
@@ -1,27 +1,72 @@
 import propTypes from '@dhis2/prop-types'
-import { colors } from '@dhis2/ui-constants'
 import { IconChevronRight16, IconChevronLeft16 } from '@dhis2/ui-icons'
 import cx from 'classnames'
 import React, { Component, createRef } from 'react'
 import { animatedSideScroll } from './animated-side-scroll.js'
 import { detectHorizontalScrollbarHeight } from './detect-horizontal-scrollbar-height.js'
+import { ScrollButton } from './scroll-button.js'
 
 class ScrollBar extends Component {
     scrollBox = createRef()
     scrollArea = createRef()
-    state = {
-        scrolledToStart: false,
-        scrolledToEnd: false,
-    }
     horizontalScrollBarHeight = detectHorizontalScrollbarHeight()
+    scrollBoxResizeObserver = null
+
+    constructor(props) {
+        super(props)
+        this.state = {
+            scrolledToStart: true,
+            scrolledToEnd: false,
+            // used to initially hide the entire content to prevent flickering
+            hideScrollButtonsInitialized: false,
+            // hide buttons initially to simplify calculations
+            hideScrollButtons: false,
+        }
+        this.scrollBoxResizeObserver = new ResizeObserver(
+            this.calculateShouldHideButtons
+        )
+    }
 
     componentDidMount() {
         this.scrollSelectedTabIntoView()
         this.attachSideScrollListener()
+        this.manageShouldHideButtons()
     }
 
     componentWillUnmount() {
         this.removeSideScrollListener()
+        this.scrollBoxResizeObserver.disconnect()
+    }
+
+    manageShouldHideButtons() {
+        const { current: scrollBox } = this.scrollBox
+        this.scrollBoxResizeObserver.observe(scrollBox)
+        this.calculateShouldHideButtons()
+    }
+
+    calculateShouldHideButtons = () => {
+        this.setState({ hideScrollButtonsInitialized: false })
+
+        const { current: scrollBox } = this.scrollBox
+        const { current: scrollArea } = this.scrollArea
+
+        const areaWidth = scrollArea.offsetWidth
+        const boxWidth = scrollBox.offsetWidth
+        const hideScrollButtons = areaWidth <= boxWidth
+
+        this.setState({ hideScrollButtons })
+
+        if (!hideScrollButtons) {
+            if (this.state.scrolledToStart) {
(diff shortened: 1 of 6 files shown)


components/tab/src/tab-bar/scroll-bar.js
```diff
@@ -204,6 +225,7 @@ class ScrollBar extends Component {
 ScrollBar.propTypes = {
     children: propTypes.node.isRequired,
     dataTest: propTypes.string.isRequired,
+    autoHideScollButtons: propTypes.bool,
```
HendrikThePendric, line 228: There's a typo here: `autoHideScollButtons` => `autoHideScrollButtons`
This change would have to be made in various places, if we should chose to keep this prop.

components/tab/src/tab-bar/scroll-bar.js
```diff
@@ -1,27 +1,85 @@
...
 
 class ScrollBar extends Component {
     scrollBox = createRef()
     scrollArea = createRef()
-    state = {
-        scrolledToStart: false,
-        scrolledToEnd: false,
-    }
     horizontalScrollBarHeight = detectHorizontalScrollbarHeight()
+    disconnectResizeObserver = null
```
HendrikThePendric, line 13: I would probably prefer to create the resize observer in the constructor, start observing in `componentDidMount` and disconnect in `componentWillUnmount`.
_Note: this suggestion only makes sense if everyone is on board with making this the default behaviour._

components/tab/src/tab-bar/scroll-button.js
```diff
@@ -0,0 +1,72 @@
+import propTypes from '@dhis2/prop-types'
+import { colors } from '@dhis2/ui-constants'
+import cx from 'classnames'
+import React from 'react'
+
+export const ScrollButton = ({ children, disabled, hidden, onClick }) => (
```
HendrikThePendric, line 6: Makes sense to extract this 👍

HendrikThePendric (commented):
Good work 👍 

I've left a few comments about details, but I think there is one bigger question: Do we need the `autoHideScollButtons` prop?

I really struggle to think of situations where one would want to show these scroll-buttons when there is nothing to scroll towards (i.e. all tabs are in view already). And having this prop in place makes the implementation more complex.

I would very much be in favour of just making this the default behaviour.

components/tab/src/tab-bar/scroll-bar.js
Mohammer5: Removed

components/tab/src/tab-bar/scroll-bar.js
Mohammer5: Done

components/tab/src/tab-bar/scroll-bar.js
```diff
@@ -1,27 +1,70 @@
...
 
 class ScrollBar extends Component {
     scrollBox = createRef()
     scrollArea = createRef()
-    state = {
-        scrolledToStart: false,
-        scrolledToEnd: false,
-    }
     horizontalScrollBarHeight = detectHorizontalScrollbarHeight()
+    scrollBoxResizeObserver = null
```
HendrikThePendric, line 13:
```suggestion
```

components/tab/src/tab-bar/scroll-bar.js
```diff
@@ -1,27 +1,70 @@
...
+        super(props)
+        this.state = {
+            scrolledToStart: true,
+            scrolledToEnd: false,
+            // used to initially hide the entire content to prevent flickering
+            hideScrollButtonsInitialized: false,
+            // hide buttons initially to simplify calculations
+            hideScrollButtons: false,
+        }
+    }
```
HendrikThePendric, line 25:
```suggestion
    constructor(props) {
        super(props)
        this.state = {
            scrolledToStart: true,
            scrolledToEnd: false,
            // used to initially hide the entire content to prevent flickering
            hideScrollButtonsInitialized: false,
            // hide buttons initially to simplify calculations
            hideScrollButtons: false,
        }
        this.scrollBoxResizeObserver = new ResizeObserver(this.calculateShouldHideButtons)
    }
```

components/tab/src/tab-bar/scroll-bar.js
```diff
@@ -1,27 +1,70 @@
...
+    constructor(props) {
+        super(props)
+        this.state = {
+            scrolledToStart: true,
+            scrolledToEnd: false,
+            // used to initially hide the entire content to prevent flickering
+            hideScrollButtonsInitialized: false,
+            // hide buttons initially to simplify calculations
+            hideScrollButtons: false,
+        }
```
HendrikThePendric, line 24:
```suggestion
        }
        this.scrollBoxResizeObserver = new ResizeObserver(this.calculateShouldHideButtons)
```

components/tab/src/tab-bar/scroll-bar.js
```diff
@@ -1,27 +1,70 @@
...
     }
 
     componentWillUnmount() {
         this.removeSideScrollListener()
+        this.scrollBoxResizeObserver.disconnect()
+    }
+
+    manageShouldHideButtons() {
+        const { current: scrollBox } = this.scrollBox
+        this.scrollBoxResizeObserver = new ResizeObserver(this.calculateShouldHideButtons)
```
HendrikThePendric, line 40:
```suggestion
```

PR #728, opened 2021-09-02
feat: expose portal component from ui and core

HendrikThePendric:
Not 100% sure I've done this correctly. Could be there are additional steps to take. In any case, this relates to https://github.com/dhis2/data-visualizer-app/pull/1859 and **should not be merged before @varl approves**.


varl:
New components should **NOT** be exported from `@dhis2/ui-core`.

Consider `@dhis2/ui-core` frozen. Only components that were available before we deprecated the `@dhis2/ui-core` package should be available from it. Same with `@dhis2/ui-widgets`. If consumers want new stuff, they need to migrate to `@dhis2/ui`.

varl (approved):

PR #729, opened 2021-09-02
fix(sharing-dialog): prevent search field remount and focus loss

mediremi:
Editing the search field results in the input field being remounted, which causes focus to be lost.

This PR fixes this by moving the field rendering from a component to a function.


varl:
Fix works nicely. Strange about the failing e2e test.

varl (approved):

PR #735, opened 2021-09-03
fix: flaky tests

KaiVandivier:
_Do not merge!_

Cypress config is altered temporarily and most workflows are disabled.

Testing why 'Last online' tests are flaky


KaiVandivier:
Currently unable to reproduce the issue enough to detect any clues :/

PR #740, opened 2021-09-10
feat(header-bar): add onLogout callback prop

KaiVandivier:
Intended for use with PWA logic, i.e. clearing caches when a user logs out.

Related to this [thread](https://dhis2.slack.com/archives/C0BP0RABF/p1631101021091900) and this [issue](https://jira.dhis2.org/browse/DHIS2-11759)


components/header-bar/src/profile/profile-menu.js
```diff
@@ -69,6 +69,7 @@ const ProfileContents = ({ name, email, avatarUrl, helpUrl }) => {
                             baseUrl,
                             'dhis-web-commons-security/logout.action'
                         )}
+                        onClick={onLogout}
```
amcgee, line 72: @KaiVandivier are we sure this will fire and complete (could take some time if the cache is large) before the redirect to the logout action?

amcgee (changes requested):
Where will we show some kind of loading mask while logging out if it takes some time?

components/header-bar/src/profile/profile-menu.js
KaiVandivier: I was just realizing that -- no I don't think there's a guarantee that this will happen before the logout action. 

If this app is hosted on the same domain as the rest of the instance, I think that's okay though -- CacheStorage and IndexedDB can be accessed domain-wide

KaiVandivier: Still I think it would be good to wait to navigate until the `onLogout` function is complete to make this more robust

varl (dismissed):

KaiVandivier:
> Where will we show some kind of loading mask while logging out if it takes some time?

Not sure yet -- that's a good idea

components/header-bar/src/profile/profile-menu.js
varl: The `MenuItem` prevents the default, so I don't think any navigation would happen if the `onLogout` callback is passed: https://github.com/dhis2/ui/blob/ee4005bd83cdabb4896444f05cd2ceedf3ef8e9d/components/menu/src/menu-item/menu-item.js#L13-L27

varl, line 72: Using the `onClick` of the `MenuItem` blocks navigation, so as-is, using the `onLogout` overrides the `href` and stops navigation.

On my first review, I figured that this was intentional so would like confirmation on that. 

The way I thought this would work is that the supplier of the `onLogout` function deals with logging out, so that function would make sure that:

1. Everything is cleaned up properly
2. Trigger the logout logic
3. Redirect the browser

varl (changes requested):

components/header-bar/src/profile/profile-menu.js
amcgee: @KaiVandivier actually thinking about this again... why doesn't the Headerbar logout button just call a "clearSensitiveCaches" directly from the app runtime?  I know that's another level of indirection...  Either way, this onClick needs to call the callback (if it exists) and then trigger the logout redirect

KaiVandivier: Yeah I was unaware of the `MenuItem` implementation -- it's useful

I thought it wasn't navigating because of something with the storybook test environment 😅 

Pushed a change below

components/header-bar/src/header-bar.stories.js
```diff
@@ -278,3 +278,23 @@ WithLastOnlineInfo.parameters = {
...
+    onLogout: () => {
+        console.log('onLogout called')
+        // tests callback async handling
+        return new Promise(res =>
+            setTimeout(() => {
+                res()
+                console.log('onLogout resolved')
+            }, 2000)
+        )
+    },
```
KaiVandivier, line 299: This verifies that navigation waits for onLogout to resolve. Loading mask and cypress test coming soon

components/header-bar/src/profile/profile-menu.js
```diff
@@ -69,6 +69,20 @@ const ProfileContents = ({ name, email, avatarUrl, helpUrl }) => {
...
+                            if (onLogout) {
+                                await onLogout()
+                            }
+                            window.location.assign(
+                                joinPath(
+                                    baseUrl,
+                                    'dhis-web-commons-security/logout.action'
+                                )
+                            )
+                        }}
```
KaiVandivier, line 85: await `onLogout` and navigate

KaiVandivier: This navigation method is limited when it is done in an `iframe` -- is there a way to emulate exactly what happens when clicking on an `a` with an `href`?

components/header-bar/src/profile/profile-menu.js
KaiVandivier: @amcgee weird, didn't see this comment when I posted 🤔 sorry

My reasoning for using the `onLogout` callback is so that the header bar doesn't need to be 'PWA aware', and the `clearSensitiveCaches` behavior may not be generalizable to non-platform apps that are using the headerbar

KaiVandivier: Although maybe the 'PWA aware' ship has sailed with the online status inclusion ⛵

KaiVandivier: Actually we probably want non-platform apps to clear caches too 🤔

KaiVandivier: Okay I think your suggestion is best -- I thought the `clearSensitiveCaches` function would live best in `@dhis2/pwa` because it's so similar to [other functions](https://github.com/dhis2/app-platform/blob/0da26e8c055122920e55791272a63dfd76ba0aa9/pwa/src/service-worker/utils.js#L64-L82) there, but I can see the app runtime would be better so `clearSCs()` can be imported and used by non-platform apps

components/header-bar/src/profile/profile-menu.js
amcgee: I'm not sure myself... maybe just send a `fetch`?

KaiVandivier: It seems like `fetch` also needs to be followed up by a `window.location` type action too

KaiVandivier:
App runtime 'clear sensitive caches' PR: https://github.com/dhis2/app-runtime/pull/1002

KaiVandivier:
Added _some_ testing materials, but multiple approaches for stubbing `clearSensitiveCaches` and `window.location.assign` were unsuccessful 🤦‍♂️ the correct behavior is manually verifiable in the storybook though

components/sharing-dialog/i18n/en.pot
```diff
@@ -149,8 +149,8 @@ msgstr "Can view and edit"
 msgid "View and edit"
 msgstr "View and edit"
 
-msgid "Sharing & Access: {{objectName}}"
-msgstr "Sharing & Access: {{objectName}}"
+msgid "Sharing & Access: {{- objectName}}"
+msgstr "Sharing & Access: {{- objectName}}"
```
amcgee, line 153: This seems like an artifact from some other change?

amcgee (approved):
Looks good, one minor comment

components/sharing-dialog/i18n/en.pot
KaiVandivier: Yeah I think so - probably just didn't get committed with the change that created it

KaiVandivier: Is it okay to leave in?

KaiVandivier:
Added real clearSensitiveCaches implementation; still failed to write good automated tests for clearing caches and navigating ☹️ but verified the correct behavior manually in storybook by using the console to add caches with `caches.open('workbox-precache-etc')` (should be preserved) and `caches.open('test-cache')` (should be removed), then clicking logout.

In the interest of moving things forward, I'll come back to the tests another time

KaiVandivier:
Navigation is causing following test to fail - going to try fixing that with [cypress network tools](https://docs.cypress.io/guides/guides/network-requests)

components/header-bar/package.json
```diff
@@ -26,7 +26,7 @@
         "test": "d2-app-scripts test --jestConfig ../../jest.config.shared.js"
     },
     "peerDependencies": {
-        "@dhis2/app-runtime": "^2.11.0",
+        "@dhis2/app-runtime": "^2.12.1",
```
varl, line 29: You need to update this in every package to be able to resolve to a single version of it.

Tip: run `yarn manypkg check` and `yarn manypkg fix`in the UI root.

varl (changes requested):

varl (approved):

PR #750, opened 2021-09-20
chore: upgrade @dhis2/prop-types to v3

HendrikThePendric:
### ToDo
- [ ] Upgrade to "@dhis2/prop-types": "^3.0.0" (don't use the beta)

### Findings regarding the prop-types v1 -> v3 migration
This PR contains relevant changes but also served as a test case for the codemod I produced in https://github.com/dhis2/prop-types/pull/254 and the migration process from v1 to v3 in general. Below are some findings:
- The codemod only transforms component files, and it includes transformations to import statements. These could lead to some changes in the dependencies that are needed.
- These dependency changes will most likely result in changes in `package.json` file(s):
     - [ALMOST ALWAYS]: `prop-types` will need to be added as a dependency if it wasn't there already. The only exception would be a theoretical scenario where only our custom prop-types functions are used and no regular prop-types from the `prop-types` lib.
     - [QUITE LIKELY]: `@dhis2/prop-types` should be removed completely if the project only uses regular prop-types from the `prop-types` package.
     - _(Upgrading the `@dhis2/prop-types` package to v3 is another change to the `package.json` file, but that can be ignored here because I am assuming dependabot PRs which already include this change.)_
 - It's probably technically possible to write a codemod that updates the package.json files, but it would be a bit cumbersome to make as it has to look into other files to identify the deps there. And also, doing this manually shouldn't be too much work, so I think it's fine to keep this step manual.
 - For mono-repos this migration will be more work because there will be several `package.json` files to update.
 
I am certain we need to document how to migrate from v1 to v3 (and probably also for v2), but I have a lot of questions about this still, which I will write in **bold**:
- We need to provide some general information on the changes, so developers scan work out how they should update their `package.json` file(s), and make sense of the changes generated by the codemod.
- We need to provide instructions about how to run the codemod, probably by using `@dhis2/cli-utils-codemods`. **Please explain how this would actually work?**
- We need to explain how to update the `package.json` file(s) after the codemod ran.
- **Where would we actually place this documentation?**
- Assuming this change is going to be propagated to apps and libs via dependabot PRs, **would it be possible to add additional info to these PRs?** _(for example a link to these migration docs)_


HendrikThePendric:
In light of some recent developments, I would actually be in favour of perhaps merging this PR with the beta version of the prop-types lib. Manually resolving merge conflicts caused by version bumps is not fun.

varl (approved):

PR #753, opened 2021-09-20
feat(pagination): add hidePageSummary prop and story for custom summary

HendrikThePendric:
Let me know:
- Do we need that addiitonal prop? I'm afraid we do since it matches the design specs. I suppose we could allow for hiding it by setting `pageSummaryText` to an empty string too, but that's not very consistent with everything else.
- Is the way we can customise summary text OK?

### DIFF
--- a/components/pagination/src/pagination.js
+++ b/components/pagination/src/pagination.js
@@ -12,6 +12,7 @@ const Pagination = ({
     dataTest,
     hidePageSizeSelect,
     hidePageSelect,
+    hidePageSummary,
     page,
     pageCount,
     pageSize,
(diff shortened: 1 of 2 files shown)


varl:
I think the new prop is symmetrical with the rest of the API and explicit enough to warrant the addition.

I don't know the details of the interpolation, but looks OK to me.

amcgee (approved):
LGTM - so long as the other props (like `total`) are optional, it should be fine.  Technically we should probably make them required *if and only if* `hidePageSummary` is false.  I don't think that's critical for this change, but something to think about for the future.

PR #756, opened 2021-09-21
fix: remove External users from sharing dialog

jenniferarnesen:
The External users sharing setting does not have any effect currently


varl (approved):
Looks good to me.

PR #768, opened 2021-09-24
fix: sharing-dialog - enable internationalization and improve text for cascade sharing

jenniferarnesen:
Fixes in this PR:
- import i18n from the locales/index.js instead of d2-i18n to ensure the same instance as the app is being used
- enable transifex for sharing dialog
- the sharing settings apply to visualizations, not items (not every item is necessarily a visualization). Therefore, the query gets all the items and counts the ones that are visualizations (should import the visualization types from analytics?)
- Always use the name "count" for plurals, i18next requires it.
- Break up sentences containing 2 plurals. Technically it should be possible for i18next to handle these cases, but our tooling doesn't seem to support it yet (e.g. extracting strings and writing the pot file)
- fixed the stories by adding the `errorReports` object to the mutation response

cascade sharing of 2 items to 1 user
image

cascade sharing of 2 items to 2 users
image

cascade sharing of 1 item to 2 users
image

cascade sharing of 1 item to 1 user
image

Other messages:

image

image

image


components/sharing-dialog/i18n/en.pot
```diff
@@ -5,107 +5,116 @@ msgstr ""
...
-msgid ""
-"There aren't any sharing settings to apply to dashboard items. Public "
-"access cannot be applied to items."
-msgstr ""
-"There aren't any sharing settings to apply to dashboard items. Public "
-"access cannot be applied to items."
+msgid "These updated sharing settings will apply to {{count}} user or group."
+msgid_plural "These updated sharing settings will apply to {{count}} user or group."
+msgstr[0] "These updated sharing settings will apply to {{count}} user or group."
+msgstr[1] "These updated sharing settings will apply to {{count}} users and groups."
```
cooper-joe, line 36: I think this should be `users or groups`, rather than `users and groups`. The rest of the messages use _or_, and I can't see a reason to use _and_ here.

components/sharing-dialog/i18n/en.pot
```diff
@@ -5,107 +5,116 @@ msgstr ""
...
+msgstr[0] "{{count}} visualization could not be updated."
+msgstr[1] "{{count}} visualizations could not be updated."
+
+msgid "Apply dashboard sharing settings to visualizations"
+msgstr "Apply dashboard sharing settings to visualizations"
 
 msgid ""
-"No dashboard items were updated. Check that you have permission to change "
-"sharing for all items."
+"Applying the same sharing settings ensures that users, groups and roles "
```
cooper-joe, line 83: I suggest swapping _ensures_ with _makes sure_.

components/sharing-dialog/i18n/en.pot
```diff
@@ -5,107 +5,116 @@ msgstr ""
...
+"that have access to the dashboard also have at least \"View only\" access "
+"to its visualizations (charts, tables, maps, event charts, event reports)."
 msgstr ""
-"No dashboard items were updated. Check that you have permission to change "
-"sharing for all items."
+"Applying the same sharing settings ensures that users, groups and roles "
```
cooper-joe, line 87: I suggest swapping _ensures_ with _makes sure_.

cooper-joe (approved):
Approving with a couple of minor suggested changes to the terminology.

components/sharing-dialog/src/dashboard-cascade-sharing-content.js
```diff
@@ -7,23 +7,33 @@ import {
...
         params: {
-            fields: 'itemCount',
+            fields: 'dashboardItems[type]',
         },
     },
 }
 
+const getVisualizationsCount = queryResponse => {
+    return queryResponse?.dashboard.dashboardItems?.length > 0
+        ? queryResponse.dashboard.dashboardItems.filter(item =>
```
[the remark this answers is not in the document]
edoardo: Verified with @vietnguyen that the count only includes VISUALIZATION, MAP, EVENT_REPORT and EVENT_CHART item types, so this filter is ok.

edoardo (approved):

PR #769, opened 2021-09-29
chore(git): unignore collections/core

ismay:
This doesn't mean that we should go committing collections/core of course. It's there so that this build artifact will show up in git history when transitioning from older ui versions that did ship core. It'll fail the build when not removed so this'll at least send a little clearer of a signal to the dev.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,10 @@ node_modules/
 build/
 dist/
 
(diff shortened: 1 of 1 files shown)


ismay:
Closed in favour of: https://jira.dhis2.org/browse/LIBS-235

PR #770, opened 2021-09-30
refactor(sharing-dialog): update sharing dialog to match new app-runtime conventions

ismay:
This fixes https://jira.dhis2.org/browse/LIBS-222. Originally meant to adjust the sharing-dialog so it conforms to the new app-runtime standards. To get there I had to refactor a fair bit. Whilst doing that I also aligned the component with the code-style of our other components.

I stuck closely to the existing functionality. API is unchanged, and the integration tests that are on master pass for this branch as well.

Notes:

- Removed references to roles as this component only deals with users and groups atm.
- Pluralization did not work, see story on master. Replaced it with a message that does not use pluralization. Bit clunky, but it works.
- I've removed tests that were reliant on specific components being rendered. This locks down implementation and not necessarily functionality, hindering internal changes. I've left tests that are independent from the specific implementation untouched. Functionality should be guarded by our integration tests in the meantime, but we should add back unit tests where necessary.
- I've removed the optimistic updates. This caused quite some complexity, and it doesn't (yet) integrate well with the app-runtime. Before implementing optimistic updates we should be able to handle update failures, rollbacks, etc.
- Instead of passing complex objects, I've refactored the components to pass primitives as much as possible. That makes it clearer what data the component actually uses and keeps our components closer to native custom components.

Todo:

- [x] Rename access-list internal components
- [x] Cascade sharing info summary doesn't get updated after mutation (and remains visible after mutating)
- [x] Clean up styles (sharing-list-item)
- [x] Component cover also blocks close button when blocking (maybe we could use context to create a root `isBlocked`, which is based on both whether it's fetching and whether it's online or not and use that in the individual components, instead of blocking the entire modal)
- [x] Debounce refetch requests (need to rebase on master first for app-runtime fix)
- [x] Check prop-types (no warnings, unused props, etc.)
- [x] Check that mutate doesn't ignore updates to the id

Follow-up:

- [ ] Simplify demo stories (and address the confusing annotation show for the type param)
- [ ] Test request failures
- [ ] Add functional unit tests (alerts are untested atm.)
- [ ] The autocomplete component has a rather vague UX. It has confused users and developers in the past in the scheduler, as you currently can't easily distinguish it from a regular input. It has several shortcomings (listed below) that should probably be fixed. We could also go with a different pattern entirely.
  - [ ] Entities that are already added should probably be hidden from the autocomplete results
  - [ ] Does not notify the user when results can't be found
  - [x] Menu layout effect does not listen for window size changes and runs every render
  - [ ] Menu width is incorrect due to input padding
  - [ ] The current implementation seems too complex to me, with the search string representing both the search and the current selection. At the very least I think it would be better if on selection the search string turns into a label, like the ones we use for the multiselect. Then again, that would mean it can no longer be a vanilla input.
- [ ] Could simplify tabs by just hiding inaccessible tabs, instead of hiding the entire tab menu
- [ ] Add data-test props
- [ ] I think we should simplify the styles. We're using very specific margins, paddings, font-sizes, etc. Just sticking to a couple standard ones would simplify the styles a lot.
- [ ] Maybe the access list should just use a ui table
- [ ] Improve semantics (we were using divs for most elements, should ensure that we're using p, h1, h2, etc.). Will help with testing and a11y.
- [ ] Deprecate and remove initialSharingSettings (if it turns out it's not being used)
- [ ] Why is the user avatar ignoring everything but the first and last name: https://github.com/dhis2/ui/pull/770#discussion_r750270762
- [ ] Cypress doesn't seem to fail when the query parameters for an intercepted test are wrong
- [ ] Take a look at the autocomplete abstraction. Sharing specific part could be moved into a hook, and composed with the autocomplete in the access-add component.


components/sharing-dialog/src/autocomplete/sharing-autocomplete.js
```diff
@@ -0,0 +1,91 @@
...
+const query = {
+    search: {
+        resource: 'sharing/search',
+        params: ({ search }) => ({
+            key: search,
+        }),
+    },
+}
+
+export const SharingAutocomplete = ({ selected, onSelection }) => {
```
Mohammer5, line 16: I think this file could be turned into a hook used by the `Autocomplete` component
* `sharing-autocomplete.js -> useAutoComplete.js`
* `autocomplete.js -> sharing-autocomplete.js`

components/sharing-dialog/src/cascade-sharing/helpers.js
```diff
@@ -0,0 +1,74 @@
...
+    if (updatedItems < visualizationsCount) {
+        return i18n.t(
+            'Not all visualization sharing settings needed to be updated. {{ updated }} updated, {{ skipped }} already had sufficient sharing settings.',
+            {
+                updated: updatedItems,
+                skipped: visualizationsCount - updatedItems,
+            }
+        )
+    }
+}
```
Mohammer5, line 58:
```suggestion
    }

    return ''
}
```

Mohammer5: I guess it make sense to always have a default return value in case none of the conditions match for whatever reason

components/sharing-dialog/src/helpers/helpers.js
```diff
@@ -0,0 +1,249 @@
...
+export const nameToInitials = (name) => {
+    if (!name) {
+        return ''
+    }
+
+    const parts = name.split(' ')
+    const firstNameInitial = parts.length > 0 ? parts.shift().charAt(0) : ''
+    const lastNameInitial = parts.length > 0 ? parts.pop().charAt(0) : ''
+
+    return `${firstNameInitial}${lastNameInitial}`
```
Mohammer5, line 53: Are middle names being ignored on purpose? I think with western names this might be fine, but I wouldn't bet on every culture doing it the same way (even if they use the Latin alphabet)

components/sharing-dialog/src/helpers/helpers.js
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * Access and constant conversion
+ */
+
+export const convertAccessToConstant = (access) => {
+    if (typeof access === 'boolean') {
+        return access ? ACCESS_VIEW_ONLY : ACCESS_NONE
+    }
+
+    if (/^rw/.test(access)) {
```
Mohammer5, line 65: Not a change request, just a suggestion (same applies to the next condition as well):
```suggestion
    if (access.startsWith('rw')) {
```

components/sharing-dialog/src/helpers/helpers.js
```diff
@@ -0,0 +1,249 @@
...
+        default:
+            return useBoolean ? false : '--------'
+    }
+}
+
+/**
+ * Mapping users and groups to style used internally
+ */
+
+export const mapUsers = ({ id, name, access }) => ({
```
Mohammer5, line 91: Function name does not describe what happens, it actually even deceives a bit as I'd expect to pass it an array (applies to `mapGroups` as well

Mohammer5: maybe `convertUserAccessToConstant`?

components/sharing-dialog/src/helpers/helpers.js
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * Helper to check whether to allow removing for the selected target
+ */
+
+export const isRemovableTarget = (target) => {
+    const permanentTargets = [SHARE_TARGET_EXTERNAL, SHARE_TARGET_PUBLIC]
+
+    // Do not allow removal of permanent targets
+    return !permanentTargets.includes(target)
+}
```
Mohammer5, line 112: Just a suggestions
```suggestion
const permanentTargets = [SHARE_TARGET_EXTERNAL, SHARE_TARGET_PUBLIC]
export const isRemovableTarget = (target) => {
    // Do not allow removal of permanent targets
    return !permanentTargets.includes(target)
}
```

components/sharing-dialog/src/helpers/helpers.js
```diff
@@ -0,0 +1,249 @@
...
+
+/**
+ * Mutation handlers
+ */
+
+export const createOnChange =
```
Mohammer5, line 118: Not a big fan of a function doing two things at the same time (creating a payload & performing a mutate), but I suppose that's highly subjective. I guess my preference would be to not pass the mutate function but instead just pass the created payload to the mutate function

Mohammer5: Changing this would also make it easier to reason about the arguments that are supplied to this function:

```js
onAdd={({ type, id, access }) => {
  const data = createOnAddPayload({ type, id, access, object })
  mutate({ data })
}}
```

components/sharing-dialog/src/tabs/tabbed-content.js
```diff
@@ -0,0 +1,89 @@
...
+    publicAccess,
+    allowPublicAccess,
+    id,
+    onAdd,
+    onChange,
+    onRemove,
+}) => {
+    const [activeTabIndex, setActiveTabIndex] = useState(0)
+
+    if (type === 'visualization') {
```
Mohammer5, line 22: Could be a constant that could also be re-used in the prop types definition

Mohammer5 (changes requested):
This looks pretty good already. There's nothing I could complain about in regards to architecture or logic, I only found a few small things

There are some places where propTypes are generic (e.g. `PropTypes.array`/`PropTypes.object`).
If we know the shape of the items/properties, then I'd prefer them to be defined explicitly as that makes it a lot easier to understand a component when looking at it in isolation

components/sharing-dialog/src/sharing-dialog.js
```diff
@@ -29,269 +31,126 @@ const mutation = {
...
-            setIsDirty(false)
+        },
+        onError: (error) => {
+            showError(error)
+            onError(error)
+            refetch()
+        },
+        onComplete: () => {
+            refetch()
+            onSave()
```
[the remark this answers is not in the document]
ismay: True, but there's a noop in the default props so this should be ok.

components/sharing-dialog/src/sharing-dialog.js
```diff
@@ -29,269 +31,126 @@ const mutation = {
...
+                        publicAccess={initialSharingSettings.public}
+                        allowPublicAccess={initialSharingSettings.allowPublic}
+                        type={type}
+                        onAdd={() => {}}
+                        onChange={() => {}}
+                        onRemove={() => {}}
+                    />
+                </Modal>
+            </FetchingContext.Provider>
+        )
```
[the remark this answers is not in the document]
ismay: I think that would be a good change yeah. One thing we should probably check though: what was the intention of the initialsharingsettings prop. It seems to me to just be a way to seed the component with data before it even does any fetching. If that's true, then ignoring initial sharing settings and showing a spinner instead _should_ be harmless.

But, technically we _are_ breaking the component by removing this functionality. Say someone has a test that asserts that passing in that prop will result in the component displaying the data, we would break that test. I think it's probably best to defer this change (which I'm still all for) and release it as a breaking change later. Maybe we could create a proposal that deprecates this in a separate PR?

components/sharing-dialog/src/sharing-dialog.js
```diff
@@ -301,6 +160,7 @@ SharingDialog.propTypes = {
         public: PropTypes.string,
         users: PropTypes.object,
```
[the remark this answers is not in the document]
ismay: Yeah I agree, that seems good. Let me know what you think of https://github.com/dhis2/ui/pull/770#discussion_r750457565, and then we can see what to do about this. I think we should be able to specify a more specific shape.

components/sharing-dialog/src/sharing-dialog.js
```diff
@@ -29,269 +31,126 @@ const mutation = {
...
+    onClose: () => {},
+    onError: () => {},
+    onSave: () => {},
 }
 
 SharingDialog.propTypes = {
     /** The id of the object to share */
     id: PropTypes.string.isRequired,
     /** The type of object to share */
     type: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
ismay: Yeah I agree. Updated that in 724e30abe69e1b27e62a7a31ac446ebcd3cf1a4d

components/sharing-dialog/src/sharing-constants.js
```diff
@@ -1,72 +1,16 @@
-import i18n from './locales/index.js'
+/**
+ * Access
+ */
```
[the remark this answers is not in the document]
ismay: Yeah I agree. Did that in e0425c59aba1850516b6ecafe9c952e9411802c0

components/sharing-dialog/src/sharing-dialog.js
```diff
@@ -1,16 +1,18 @@
-import { ButtonStrip, Button } from '@dhis2-ui/button'
-import { Modal, ModalTitle, ModalContent, ModalActions } from '@dhis2-ui/modal'
 import { useAlert, useDataQuery, useDataMutation } from '@dhis2/app-runtime'
 import PropTypes from 'prop-types'
-import React, { useEffect, useState } from 'react'
-import { DashboardSharingContent } from './dashboard-sharing-content.js'
-import { DefaultSharingContent } from './default-sharing-content.js'
-import i18n from './locales/index.js'
+import React, { useEffect } from 'react'
+import { FetchingContext } from './context/index.js'
```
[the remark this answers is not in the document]
ismay: Yeah that's true, renamed it in b448a1fc1c116444fb955dbcf07e6b3eb04c9816

components/sharing-dialog/src/sharing-dialog.js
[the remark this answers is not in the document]
Mohammer5: Agreed

components/sharing-dialog/src/sharing-dialog.js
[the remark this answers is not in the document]
Mohammer5: I agreed with your remark in that comment, so this one should be more specific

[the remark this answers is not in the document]
ismay: Fixed in 98de5c0ee01df6a01b7136f0019b3d97e5d84c0c

components/sharing-dialog/src/cascade-sharing/helpers.js
ismay: Yeah I agree, fixed in c7ee062925d53b9ef0e18902a8abf2b35b94b3c7

components/sharing-dialog/src/helpers/helpers.js
ismay: Yeah good question. This area of the code has now been replaced by the user-avatar, but I think that that does the same: https://github.com/dhis2/ui/blob/master/components/user-avatar/src/text-avatar.js. Probably good to find a less western-name-specific solution.

components/sharing-dialog/src/sharing-dialog.js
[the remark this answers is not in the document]
ismay: Just unresolving this so I don't forget to follow up.

components/sharing-dialog/src/helpers/helpers.js
ismay: Yeah good idea, changed in 9658e900fc46b3a9a3c15bd21ea40b0ff33e5504

components/sharing-dialog/src/tabs/tabbed-content.js
ismay: Fixed in 724e30abe69e1b27e62a7a31ac446ebcd3cf1a4d

components/sharing-dialog/src/helpers/helpers.js
ismay: Yeah this bit didn't have to be as complex, I've replaced it with a function that just picks out the access property here: faaee59cba77a36abae4ad767ce2af4b5408de15

components/sharing-dialog/src/helpers/helpers.js
ismay: Yeah sounds good, fixed in 1e77401dd9815696c812f88cd36ba9a2b0d9aa5a.

components/sharing-dialog/src/helpers/helpers.js
ismay: Yeah I don't like this section either. It's also a bit annoying that here type and id are shadowing the sharing-dialog type and id props, but they're actually referring to something completely different. I'll see if I can improve this.

components/sharing-dialog/src/sharing-dialog.js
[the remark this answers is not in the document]
ismay: Added it to the top level follow ups.

components/sharing-dialog/src/helpers/helpers.js
ismay: Added it to the follow ups.

components/sharing-dialog/src/helpers/helpers.js
ismay: Good suggestion, I like this pattern a lot better. Changed it in b3cade77b1854ee386f329420559f2f5faec51bc, and I've also updated mutate here: 828ef46f003c59fce70234f0fb365929bba545d2, so that it picks up on prop changes in the root component (otherwise it'd keep using the initial props and update the wrong object).

ismay: As you can see in the last commit, the type and id params in the handlers shadow the global type and id (which is not ideal). I've updated the naming only for the root component. But we could propagate those names downwards if you think that's clearer.

ismay: I'll resolve it, but let me know if you think anyone still needs to be addressed.

components/sharing-dialog/src/autocomplete/sharing-autocomplete.js
ismay: So judging from this line: `sharing-autocomplete.js -> useAutoComplete.js` I assume you mean that `autocomplete.js` could be a hook? Because that component is mostly markup, so to me having it be a component feels most appropriate. Or do you mean something else?

Mohammer5: The other way. What's currently in `sharing-autocomplete.js` is basically no markup whatsoever but using hooks.
So instead of wrapping that in a component, it could be wrapped in its own hook instead (hook logic inside hook instead of hook logic inside component).

Then the contents of what's currently in `autocomplete.js` could be moved to `sharing-autocomplete.js` and simply use the newly created hook

ismay: Yeah I see what you mean. The reason for the current abstraction is because the `autocomplete.js` component represents the generic part. Conceptually it doesn't make as much sense to me to abstract the logic in `sharing-autocomplete.js` to a hook and then move what's in `autocomplete.js` to the `sharing-autocomplete.js` component. 

Because if you'd want another type of autocomplete you'd have to either extract the markup we moved to `sharing-autocomplete.js` to a generic component (which is what we have now), or repeat the markup.

I know it's not a situation we currently face, but that's why I tend towards the current abstraction. The other might be convenient, but to me it puts the inheritance opposite of how I conceptualize it. What do you think?

Mohammer5: We talked about this on slack quickly and agreed that this is mostly a matter of perspective, so I'm marking this as resolved and leave it up to @ismay what he wants to do

Mohammer5 (approved):

PR #772, opened 2021-10-01
fix: sharing dialog text improvements

jenniferarnesen:
These text changes were requested by Joe but omitted from the previous PR


components/sharing-dialog/i18n/en.pot
```diff
@@ -33,7 +33,7 @@ msgstr[1] ""
 msgid "These updated sharing settings will apply to {{count}} user or group."
 msgid_plural "These updated sharing settings will apply to {{count}} user or group."
 msgstr[0] "These updated sharing settings will apply to {{count}} user or group."
-msgstr[1] "These updated sharing settings will apply to {{count}} users and groups."
+msgstr[1] "These updated sharing settings will apply to {{count}} users or groups."
```
edoardo, line 36: I guess it can actually be both "and" and "or".

edoardo (approved):

components/sharing-dialog/i18n/en.pot
jenniferarnesen: Joe wanted "or" and far be it from me to argue with an Englishman :)

PR #795, opened 2021-10-07
feat(headerbar): display instance's & app's info in profile menu - LIBS-176

Mohammer5:
Blocked by #820 
Relates to LIBS-176 ([link](https://jira.dhis2.org/browse/LIBS-176))

### Open ToDos

- [x] Cypress test

### Open questions

- [x] When the instance's version is too long, the build version breaks into the next line.
    With the current styles it's a lot harder to parse
- [x] I have to load the instance's data from the server, so there's a short loading time (the app data is available though). What should we display during that time?
- [x] If an error occurs while loading the instance's info, right now I just don't render anything. Is that correct or should we display some kind of error message?
- [x] Right now there's no proper way of retrieving the app's build revision, so I just left it out. Is that okay? Or is that information crucial?


varl (changes requested):
@cooper-joe @amcgee @Mohammer5 

I think we should avoid hard coding "DHIS2" for the same reason we have custom logo support. E.g. DATIM may want to show its own version string and product name, instead of DHIS2 and its version.

Mohammer5:
Still relevant, just went under the radar

amcgee:
I'm reviving this and adding PWA update notification, with slightly updated styling and using context instead of loading the version directly.  So I'll close this PR and have opened #1153 - but it's based on your work @Mohammer5 so big hats off in thanks to you!

PR #799, opened 2021-10-11
test(sharing-dialog): add e2e tests

ghost:
See https://jira.dhis2.org/browse/LIBS-222

These tests aren't great. But the component does not have any data-test attributes, and semantically it's hard to find the proper selectors (without a lot of refactoring). So I decided to test with what I have, and then we'll have to fix the selectors whilst refactoring the component in the refactoring PR.

Add tests for:

- [x] Setting the dialog title
- [x] Disabling control of the access level
- [x] Search functionality
- [x] Dashboard and visualization types

Follow up:

- Remove descriptive unit tests, replace with functional unit tests (i.e., no asserting _that_ a certain component is rendered somewhere, but instead test that given certain _input_ for the sharing dialog, it yields the expected _output_. To allow the internals to change whilst ensuring we don't break the API)
- I wonder if we could eventually decouple the network layer from this component. 
- Search results do not seem to match search string
- Search seems to disappear when displayed in regular storybook, but not in cypress' storybook
- Add unit tests for the different kinds of messages possible in the dashboard tab
- Assert expected network traffic as well

Notes:

- It seems that the all users item can't be hidden (did not verify in source yet)
- It seems that users and groups can't be set to no access, instead it can be removed. You can break the component by passing a string that does not grant access

### DIFF
--- a/components/sharing-dialog/src/features/add-user/index.js
+++ b/components/sharing-dialog/src/features/add-user/index.js
@@ -93,38 +93,50 @@ When('the user gives group view and edit access', () => {
  * the <target> should be added to the access list with <level> access
  */
 
-Then('the user should be added to the access list with view only access', () => {
-    cy.contains('.sharing-list-item', 'A user')
-        .should('be.visible')
-        .as('user-item')
+Then(
+    'the user should be added to the access list with view only access',
+    () => {
+        cy.contains('.sharing-list-item', 'A user')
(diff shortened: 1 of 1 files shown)


ismay:
> Only thing left: Some of the "then" statements still don't use the "should" grammar

Yup, just finished that one. Should be good now.

PR #828, opened 2021-11-02
feat(modal): add support for custom dimensions

HendrikThePendric:
See https://jira.dhis2.org/browse/LIBS-2

### DIFF
--- a/components/modal/src/modal/modal.js
+++ b/components/modal/src/modal/modal.js
@@ -67,8 +67,8 @@ export const Modal = ({
                 aside {
                     overflow-y: hidden;
                     height: ${height || 'auto'};
-                    max-height: ${height ? 'auto' : 'calc(100vh - 128px)'};
-                    max-height: ${width ? 'auto' : 'calc(100vw - 128px)'};
+                    max-height: ${height ? 'none' : 'calc(100vh - 128px)'};
+                    max-height: ${width ? 'none' : 'calc(100vw - 128px)'};
                     width: ${width || '600px'};
                 }
 
--- a/components/modal/src/modal/modal.js
+++ b/components/modal/src/modal/modal.js
@@ -7,16 +7,20 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import { resolve } from 'styled-jsx/css'
 
-const scrollBoxCard = resolve`
-    div {
-        display: flex;
-        flex-direction: column;
-        max-height: calc(100vh - ${2 * spacersNum.dp64}px);
-        max-width: calc(100vw - ${2 * spacersNum.dp64}px);
-    }
-`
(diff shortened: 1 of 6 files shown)


Mohammer5 (changes requested):
No test (not sure what'd be the best test here, but I think jest is more appropriate than cypress)

---

I'm wondering if we should also add `maxWidth` and `maxHeight` props, which would allow the following:
```jsx
<Modal ... height="95vh" width="400px" maxWidth="95vw" />
```

varl (changes requested):
I am wary that this opens up for a lot of bad uses of the Modal. If a Modal needs to be larger than "large", that indicates that whatever we are trying to fit into it should be on its own screen.

I would like to discuss this with @cooper-joe to understand the use cases better.

HendrikThePendric:
> If a Modal needs to be larger than "large", that indicates that whatever we are trying to fit into it should be on its own screen.

That's a bold conclusion IMO, but a good starting point for a conversation nonetheless.

The use-case is to view interpretations of a given visualisation. We want to show the same visualisation as the user is currently viewing, but in the state the visualisation was when the interpretation was written. IMO some sort of modal experience is a great fit with this.

HendrikThePendric:
I have implemented dynamic dimensions using a `fluid` prop now. This comes with two downsides/gotchas/known-issues:
- One point which I decided to leave unaddressed is that when the `position` prop has value `top` (default) or `bottom` the modal has some predefined whitespace above/below. So if you would want to implement a modal with a very large height, say `99vh`, you would have to set both the `fluid` and the `position="middle"` prop on the modal. Otherwise the default `top` position would be applied and the whitespace that comes with that would cause the modal to extend beyond the bottom window border.
- Another point is the layout. In [the fluid modal story](https://pr-828--dhis2-ui.netlify.app/?path=/story/layout-modal--fluid-title-content-primary) you see the buttons (which have been placed in a `ModalActions` component) are high up on the modal, while they should really be aligned with the bottom of the modal. This is because we're now using a `Box` to control the modal's dimensions, and this is not `display: flex`. So if apps want to use a fluid modal with a bottom-aligned modal-footer, they probably need to insert some custom div with dimensions + flexbox. I decided to leave this unaddressed as well.

HendrikThePendric:
Given the two fairly substantial known-issues above, does it even make sense to implement this prop? I know we don't like overriding UI styles, but I almost feel that is a cleaner approach then having to re-implement the layout inside the component and ensuring you set the modal's position to "middle".

HendrikThePendric:
I've discussed this in a bit more detail with @mediremi and decided to close this PR. Adding custom dimension was not an acceptable solution and the alternative comes with too many downsides. In the end the cleanest solution here is just to let apps add style overrides 🤷‍♂️ .

PR #830, opened 2021-11-04
feat: user-icon component

HendrikThePendric:
## Tasks
- [x] Let user-icon construct its own image src from an avatar id
- [x] Rename `UserIcon` to `UserAvatar` and extract into a new package => `@dhis2-ui/user-avatar`
- [x] Add @dhis2-ui/user-icon as a dependency to the `HeaderBar`
- [x] Add @dhis2-ui/user-icon as a dependency to the `SharingDialog`
- [x] Add @dhis2-ui/user-icon as a dependency to `@dhis2/ui` and expose it

## Context
We currently have two individual implementations of the user-icon: one in the header-bar and one in the sharing-dialog. In the new event-reports-app we need to show user-icons too. Instead of having unique implementations in three places we should consolidate them into a package.

## Known issues
I have had to make a choice when extracting this component to be stand-alone:
1. Keep the `avatarUrl` prop and let apps construct a valid image source themselves
2. Make the user-icon component responsible for constructing a valid image source from a `avatarId`

For a reusable component the second option made more sense to me, mainly because constructing a valid img url is [a bit tricky](https://github.com/dhis2/ui/blob/d7e1693bf6c75fa68e982793d95bc8bf21cfe2fb/components/header-bar/src/user-icon/image-icon.js#L5-L14), so I went with this one. However, this does come with some trade-offs:
1. Effectively this component can only be used when used in an app-platform app, since it relies on the `useConfig` hook. I'm not sure to what extend this is an issue. The header-bar also relies on `useConfig`, so apparently there are no set rules about this.
2. The header-bar package uses a `joinPaths` helper which is also used to construct the image source. When extracting the user-icon package I simply copied this logic into the `useAvatarImgSrc` hook. So there is some code-duplication there, but I think this is to be preferred over having this logic duplicated all over our actual apps.


mediremi:
Should we call this new component `UserAvatar` instead of `UserIcon` since it'll take an `avatarId` prop? For me a user icon would be something that shows the status of user (e.g. enabled/disabled/...)

HendrikThePendric:
> Should we call this new component `UserAvatar` instead of `UserIcon` since it'll take an `avatarId` prop? For me a user icon would be an icon that shows a user's status (e.g. enabled/disabled/...)

I agree. I've updated the description.

mediremi:
> Effectively this component can only be used when used in an app-platform app, since it relies on the useConfig hook. I'm not sure to what extend this is an issue. The header-bar also relies on useConfig, so apparently there are no set rules about this.

We could always make the consumer pass `baseUrl` as a prop, but since this component relies on the fact that the consuming app is talking to a DHIS2 instance (since it's fetching avatars from `api/fileResources`) I think depending on `useConfig` is fine.

components/header-bar/src/user-icon/image-icon.js
```diff
@@ -1,22 +1,37 @@
...
-            }
-        `}</style>
-    </>
-)
+    return avatarId
+        ? [baseUrl, 'api/fileResources', avatarId, 'data']
+              .filter((part) => !!part)
+              .map((part) => part.replace(/^\/+|\/+$/g, ''))
+              .join('/')
+        : null
```
varl, line 13: I don't remember the reason for that complicated way to build the URL, to me it seems that something like this should work fine.

```suggestion
        ? new URL(`api/fileResources/${avatarId}/data`, baseUrl).href
        : null
```

mediremi: > I don't remember the reason for that complicated way to build the URL

The `filter + map + join` would be required for [a more generic `joinPath` function](https://github.com/dhis2/ui/blob/d7e1693bf6c75fa68e982793d95bc8bf21cfe2fb/components/header-bar/src/join-path.js) (which we use in the header bar for generating app URLs in the app drawer).

For the `UserAvatar` use case I agree that your solution will work fine.

varl:
> 1. Effectively this component can only be used when used in an app-platform app, since it relies on the `useConfig` hook. I'm not sure to what extend this is an issue. The header-bar also relies on `useConfig`, so apparently there are no set rules about this.

The HeaderBar is a large component that does a lot of things, and relies on App Runtime to function. The `useConfig` hook returns a [default context](https://github.com/dhis2/app-runtime/blob/master/services/config/src/ConfigContext.ts#L4-L7) if it's used outside of a Platform App so not sure if that's a problem.

> 2. The header-bar package uses a `joinPaths` helper which is also used to construct the image source. When extracting the user-icon package I simply copied this logic into the `useAvatarImgSrc` hook. So there is some code-duplication there, but I think this is to be preferred over having this logic duplicated all over our actual apps.

I added a comment about this, I wasn't sure why `new URL(`api/fileResources/${avatarId}/data`, baseUrl).href` would not work, but I found the problem when exploring it a bit more; relative URLs.

```
> let baseUrl = 'https://instance.dhis2.org/dev'

> function url1 (avatarId) {
... return [baseUrl, 'api/fileResources', avatarId, 'data']
...               .filter((part) => !!part)
...               .map((part) => part.replace(/^\/+|\/+$/g, ''))
...               .join('/')
... }

> function url2 (avatarId) {
... return new URL(`api/fileResources/${avatarId}/data`, baseUrl).href
... }

> url1('123')
'https://instance.dhis2.org/dev/api/fileResources/123/data'
> url2('123')
'https://instance.dhis2.org/api/fileResources/123/data'

> url1('weirdåéë^#*@')
'https://instance.dhis2.org/dev/api/fileResources/weirdåéë^#*@/data'
> url2('weirdåéë^#*@')
'https://instance.dhis2.org/api/fileResources/weird%C3%A5%C3%A9%C3%AB^#*@/data'

> url1()
'https://instance.dhis2.org/dev/api/fileResources/data'
> url2()
'https://instance.dhis2.org/api/fileResources/undefined/data'

> url1('')
'https://instance.dhis2.org/dev/api/fileResources/data'
> url2('')
'https://instance.dhis2.org/api/fileResources//data'

> baseUrl = '..'
'..'
> url1('foo')
'../api/fileResources/foo/data'
> url2('foo')
Uncaught TypeError [ERR_INVALID_URL]: Invalid URL
    at __node_internal_captureLargerStackTrace (node:internal/errors:464:5)
    at new NodeError (node:internal/errors:371:5)
    at onParseError (node:internal/url:552:9)
    at new URL (node:internal/url:628:5)
    at new URL (node:internal/url:625:22)
    at url2 (REPL28:2:8) {
  input: '..',
  code: 'ERR_INVALID_URL'
}
```

components/user-avatar/src/text-avatar.js
```diff
@@ -0,0 +1,43 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+export const getInitials = (name) =>
+    name
```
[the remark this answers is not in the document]
HendrikThePendric: 🤔 Good point. I think first and last word is probably going to produce the desired outcome in more cases. But actually it would probably be even more accurate to not feed a full name into this component, but provide a first and last name instead. And then I guess we could take the first letter of the first word of each field....

For my quirky language, it would actually be even better to take the first capitalised word. For example, my last name is "de Graaf" but "G" should be the letter used in my avatar.

At the very least this comment highlights that it makes a lot of sense to extract this component. We don't want the same users to have different avatars in different places in the system.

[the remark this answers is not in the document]
HendrikThePendric: I've adjusted the implementation in this component to use the first and last word in the name as @edoardo suggested.

cooper-joe:
The [design system guidelines](https://github.com/dhis2/design-system/blob/master/atoms/avatar.md) ([spec, slightly stale but still OK](https://www.sketch.com/s/2f0dba87-fb82-415b-863d-0c2ea63ee77b/a/OlqYEk) defines other sizes and modes (`no-initials`: to cover the eventuality that initials can't or shouldn't be used, `user-photo`: to populate the avatar with the user profile photo). 

Is it worth including them in this PR? The other sizes are particularly useful immediately.

components/sharing-dialog/src/__tests__/SharingListItem.test.js
```diff
@@ -1,7 +1,8 @@
+import { UserAvatar } from '@dhis2-ui/user-avatar'
 import { IconWorld24, IconUserGroup24 } from '@dhis2/ui-icons'
 import { shallow } from 'enzyme'
 import React from 'react'
-import { Avatar } from '../icons/avatar.js'
+// import { UserAvatar } from '../../../user-avatar/src/index.js'
```
mediremi, line 5:
```suggestion
```

components/sharing-dialog/src/sharing-list-item.js
```diff
@@ -1,10 +1,12 @@
+import { Box } from '@dhis2-ui/box'
 import { Divider } from '@dhis2-ui/divider'
+import { UserAvatar } from '@dhis2-ui/user-avatar'
 import { colors } from '@dhis2/ui-constants'
 import { IconWorld24, IconUserGroup24 } from '@dhis2/ui-icons'
 import PropTypes from 'prop-types'
 import React from 'react'
+// import { UserAvatar } from '../../user-avatar/src/index.js'
```
mediremi, line 8:
```suggestion
```

yarn.lock
```diff
@@ -2197,6 +2197,16 @@
     "@dhis2/pwa" "8.0.0"
     moment "^2.24.0"
 
+"@dhis2/app-runtime@^3":
+  version "3.2.4"
```
mediremi, line 2201: Since this lock file already has `@dhis2/app-runtime@3.2.0`, we need to run `npx yarn-deduplicate yarn.lock`. Might be a good time to bump cli-app-scripts, which contains a check for this.

mediremi (changes requested):

components/user-avatar/src/user-avatar.stories.js
```diff
@@ -0,0 +1,26 @@
+import { Box } from '@dhis2-ui/box'
+import React from 'react'
+import { UserAvatar } from './user-avatar.js'
+
+const description = `
```
mediremi, line 5: Defining a description in this way is now deprecated, and should instead be done using a `/**` comment above the component's definition. This allows the API docs to render the description too. We'll also want a link to the design system.

```js
// In components/user-avatar/src/user-avatar.js:

/**
An avatar is a visual icon that represents a user.

Avatars are used to give extra information when a user is mentioned or displayed in DHIS2. The avatar contains either user uploaded photograph, initials or a default icon. The avatar is intended to give context to a user and help to quickly identify different users. An avatar is usually shown alongside the user name, but can be used alone to show a visual hint of a user.

See specification: [Design System](https://github.com/dhis2/design-system/blob/master/atoms/avatar.md)
*/

export const UserAvatar = ({ avatarId, name, dataTest }) => {
```

HendrikThePendric:
> Is it worth including them in this PR? The other sizes are particularly useful immediately.

@cooper-joe: just to note, that the current implementation of this component doesn't include any size at all. The component simply takes up all available space in the parent. Effectively you put it in a `Box` with dimensions to set a size.

I don't think I would want to implement any size props right now. The design system specifies these sizes:
- 16px `extraSmall`
- 24px `small`
- 32px `medium`
- 48px `large`
- 72px `extraLarge`

However I have already spotted an instance of a user-avatar in yet another size. The most prominent one in the HeaderBar is actually 36px.

Also about this:
> `no-initials`: to cover the eventuality that initials can't or shouldn't be used

I personally don't think we should take this approach anymore when working on the design system. Sometimes it can be useful to keep future plans in mind when building something. But actually implementing something we don't have a need for right now is not a good use of our limited resources.

FYI, the user-photo part is already implemented: if you pass an avatarId the component will render a photo.

mediremi (approved):

cooper-joe:
> just to note, that the current implementation of this component doesn't include any size at all. The component simply takes up all available space in the parent. Effectively you put it in a Box with dimensions to set a size.

Alright, that's good to know.

> However I have already spotted an instance of a user-avatar in yet another size. The most prominent one in the HeaderBar is actually 36px.

Good point. I believe this is due to the stale-ness of the spec. The old headerbar (pre 2.34, when this spec was produced) used an avatar of size `32px`, so that's probably why it made it to the spec. An updated spec would use 36px.

> I don't think I would want to implement any size props right now. 

If the above point is solved (`32px` can be replaced by `36px`), do you still feel this way? Could you expand on the reason? My position is that adapting different use cases to use one of the available sizes is a worthy trade of the consistency and simplicity gained.

---

> > no-initials: to cover the eventuality that initials can't or shouldn't be used

> I personally don't think we should take this approach anymore when working on the design system. Sometimes it can be useful to keep future plans in mind when building something. But actually implementing something we don't have a need for right now is not a good use of our limited resources.

I agree we shouldn't use time on things we don't need now. I think this part of the spec is an attempt to cover edge cases like only having one name, or having a name that with characters that can't be rendered in an avatar. I don't think we can classify this as "future plans", as there are certainly instances of this already. But a decision can be taken that this is out of scope of the component. The `icon` variant is perhaps best seen as an attempt to provide a catch-all fallback.

> FYI, the user-photo part is already implemented: if you pass an avatarId the component will render a photo.

Awesome 👍.

components/user-avatar/package.json
```diff
@@ -0,0 +1,49 @@
+{
+    "name": "@dhis2-ui/user-avatar",
+    "version": "7.4.6",
```
mediremi, line 3: Remember to bump this to `7.5.0` after sorting out the merge conflicts

HendrikThePendric:
I've now implemented sizes and changed a few things to reflect the design system. Some notable deviations from the specs, which have been discussed with @cooper-joe are these:
- Font sizes for various avatar sizes have been tweaked a little bit. These are now like this:
    - `extrasmall`: 7px
    - `small`: 9px
    - `medium`: 14px
    - `large`: 22px
    - `extralarge`: 32px
- Background-color has 80% opacity instead of 90%.

cooper-joe:
> I've now implemented sizes and changed a few things to reflect the design system. Some notable deviations from the specs...

For clarity, I've also updated the [avatar design spec](https://github.com/dhis2/design-system/blob/ecdf181ea6bcd9605e3c76c0a4d4d21fd25b9fdf/images/avatar-spec.png) with the changes listed above.

PR #838, opened 2021-11-10
feat(modal): viewport constrained fluid size

varl:
## Use case

There is a recurring need for larger Modals than our `large` variant (600px wide).

## Potential solutions

- `height` and `width` props that sets the size of the Modal
- Override using `className` and `!important`
- Unconstrained fluid Modal size that scrolls with the page
- Constrained fluid Modal size that scrolls the ModalContent

## Chosen solution

> Constrained fluid Modal size that scrolls the ModalContent

Allowing the Modal to have a `fluid` prop allows the size of the Modal to be:

- adaptable to the content of the ModalContent component
- controlled by the app developer, by utilising an additional wrapper with defined sizes inside of the ModalContent

This provides enough flexibility without opening up the API too much.

The other solution's drawbacks that excluded them:

- `height`/`width` props tied into ongoing discussion about overrides/extensions/global props
- using `className`/`!important` is not recommended by us
- scrolling the page because the Modal is very large is not an UX we want as it obscures the modality traits

## Future

- Consider changing the default from `medium` to `fluid` (breaking change)
- If we want to allow page scroll, we need to open the Layer component to allow overflow.
- The Card CSS can probably be simplified.

### DIFF
--- a/components/modal/src/modal-title/modal-title.js
+++ b/components/modal/src/modal-title/modal-title.js
@@ -1,4 +1,4 @@
-import { colors, spacers } from '@dhis2/ui-constants'
+import { colors } from '@dhis2/ui-constants'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React from 'react'
@@ -14,7 +14,7 @@ export const ModalTitle = ({ children, dataTest }) => (
                 color: ${colors.grey900};
                 font-size: 20px;
                 font-weight: 500;
-                line-height: ${spacers.dp24};
+                line-height: 24px;
                 margin: 0;
             }
         `}</style>
--- a/components/modal/src/modal-title/modal-title.js
+++ b/components/modal/src/modal-title/modal-title.js
@@ -9,12 +9,13 @@ export const ModalTitle = ({ children, dataTest }) => (
 
         <style jsx>{`
             h1 {
+                order: 1;
+                align-self: flex-start;
                 color: ${colors.grey900};
                 font-size: 20px;
                 font-weight: 500;
-                line-height: 24px;
+                line-height: ${spacers.dp24};
                 margin: 0;
-                padding: ${spacers.dp24} ${spacers.dp24} 0 ${spacers.dp24};
             }
         `}</style>
     </h1>
--- a/components/modal/src/modal/modal.js
+++ b/components/modal/src/modal/modal.js
@@ -7,16 +7,18 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import { resolve } from 'styled-jsx/css'
 
-const scrollBoxCard = resolve`
-    div {
-        display: flex;
-        flex-direction: column;
-        max-height: calc(100vh - ${2 * spacersNum.dp64}px);
-        max-width: calc(100vw - ${2 * spacersNum.dp64}px);
-    }
-`
+const getScrollBoxCardStyles = () => {
+    return resolve`
+        div {
+            display: flex;
+            flex-direction: column;
+            max-height: calc(100vh - ${2 * spacersNum.dp64}px);
+            max-width: calc(100vw - ${2 * spacersNum.dp64}px);
+        }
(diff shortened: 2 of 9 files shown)


components/modal/src/modal/modal.js
```diff
@@ -7,67 +7,60 @@ import PropTypes from 'prop-types'
...
-
-const layerStyles = resolve`
-    div {
-		display: none;
-    }
-`
+const resolveLayerStyles = (hide) =>
+    resolve`
+        div {
+            padding: ${spacers.dp64};
```
varl, line 13: @cooper-joe I wonder if we should make this the default here: https://github.com/dhis2/ui/blob/master/components/layer/src/layer.js#L40

That would stop anything that is rendered on top of the Layer to cover the entire viewport.

cooper-joe: Yes, that sounds reasonable. That padding doesn't really concern the `modal`, it's more of a display property of the `Layer`s contents. So as a principle I support this change. Only issue I can see is some unknown use cases of the `Layer` that will be broken or mangled by the change.

cooper-joe (approved):

components/modal/src/modal-title/modal-title.js
```diff
@@ -9,12 +9,13 @@ export const ModalTitle = ({ children, dataTest }) => (
 
         <style jsx>{`
             h1 {
+                order: 1;
+                align-self: flex-start;
                 color: ${colors.grey900};
                 font-size: 20px;
                 font-weight: 500;
-                line-height: 24px;
+                line-height: ${spacers.dp24};
```
Mohammer5, line 17: I think this is a misuse of the spacing values. Although line height and spacing can be closely connected, they do have their differences. Until we have a dedicated theme system for font, I think we should use static values:

> `Line-height` and `vertical-align` are simple CSS properties. So simple that most of us are convinced to fully understand how they work and how to use them. But it’s not. They really are complex, maybe the hardest ones, as **they have a major role in the creation of one of the less-known feature of CSS: inline formatting context**.
> https://iamvdo.me/en/blog/css-font-metrics-line-height-and-vertical-align

components/modal/src/modal/modal.js
```diff
@@ -7,67 +7,60 @@ import PropTypes from 'prop-types'
...
     onClose,
     position,
     small,
 }) => {
+    const layerStyles = resolveLayerStyles(hide)
     return (
         <Layer
             onClick={onClose}
-            className={hide ? layerStyles.className : ''}
+            className={layerStyles.className}
```
Mohammer5, line 33: Although VERY unlikely, this could be a breaking change; therefore technically it is a breaking change

Mohammer5 (changes requested):

components/modal/src/modal-title/modal-title.js
varl: Thanks. I'll change it back.

varl, line 17:
```suggestion
                line-height: 24px;
```

components/modal/src/modal/modal.js
varl: Yeah, we'll have to investigate in a separate PR. Leaving this as is in this, but wanted to pointed it out.

components/modal/src/modal/modal.js
varl: The API has not changed, and all the changes are contained within the Modal components. I don't see how this could be a breaking change. Would you mind expanding on this ?

varl: The `hide` functionality is still handled in the `resolveLayerStyles` function: `display: ${hide ? 'none' : 'block'};` on L14.

Mohammer5: nvm.. My bad

Mohammer5 (approved):

PR #841, opened 2021-11-12
fix: make root sub org unit selection work

Mohammer5:
Closes LIBS-254 ([link](https://jira.dhis2.org/browse/LIBS-254))

- [x] Works with multiple roots on different hierarchy levels
- [x] Tested with cypress (for details see additions/changes in the feature files)


components/organisation-unit-tree/src/organisation-unit-node/organisation-unit-node.js
```diff
@@ -177,6 +191,7 @@ OrganisationUnitNode.propTypes = {
     dataTest: PropTypes.string.isRequired,
     id: PropTypes.string.isRequired,
     renderNodeLabel: PropTypes.func.isRequired,
+    rootId: PropTypes.string.isRequired,
```
varl, line 194: This addition should be a `feat` if it's part of the public API for the component.

varl (approved):
Nice job fixing this without breakages. I noticed that there is an addition to the API of the OrgUnitNode component, which seems like a feature ?

components/organisation-unit-tree/src/organisation-unit-node/organisation-unit-node.js
Mohammer5: This is internal. The component we expose is the `OrganisationUnitTree`, which renders the `OrganisationUnitNode` and provides the `rootId` [here](https://github.com/dhis2/ui/blob/c05de2bbd90a3293af698917ec44d93e3423fd22/components/organisation-unit-tree/src/organisation-unit-tree/organisation-unit-tree.js#L84).

varl: Ah ! Great, thanks for clarifying.

PR #844, opened 2021-11-15
chore(deps): update app-runtime

ismay:
This updates the app-runtime to the latest version, which includes this fix: https://github.com/dhis2/app-runtime/commit/1b7a42f3592074bbe0fc2798a60b81423159962a. I need that for the sharing-dialog to be able to debounce requests.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2197,37 +2197,37 @@
     "@dhis2/pwa" "8.0.0"
     moment "^2.24.0"
 
-"@dhis2/app-runtime@^3", "@dhis2/app-runtime@^3.0.0", "@dhis2/app-runtime@^3.2.0":
-  version "3.2.4"
-  resolved "https://registry.yarnpkg.com/@dhis2/app-runtime/-/app-runtime-3.2.4.tgz#78f73c799a764551f364642a18d8dda18fa00ef2"
-  integrity sha512-VApvxCHZ1AIVVO90a5BGSzwcmAXPZOhf2eMw/rzEF7w8ONWmidWocU5HwGUVb5Jn2/5UL8fGOvO0ZOKDgeev8A==
+"@dhis2/app-runtime@^3", "@dhis2/app-runtime@^3.0.0", "@dhis2/app-runtime@^3.2.0", "@dhis2/app-runtime@^3.2.5":
+  version "3.2.5"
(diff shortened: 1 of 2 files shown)


mediremi (approved):

yarn.lock
```diff
@@ -2207,30 +2207,64 @@
     "@dhis2/app-service-data" "3.2.4"
     "@dhis2/app-service-offline" "3.2.4"
 
+"@dhis2/app-runtime@^3.2.5":
```
mediremi, line 2210: There's two versions of app-runtime in this file so it needs to be deduplicated with `npx yarn-deduplicate yarn.lock`

mediremi (changes requested):

yarn.lock
mediremi: This would be a good opportunity to bump `cli-app-scripts`, whose latest version checks for this automatically

ismay: You're right, good catch. Should be good now.

ismay: > This would be a good opportunity to bump cli-app-scripts, whose latest version checks for this automatically

Ah yeah I can do that. Will do that now

ismay: Updated cli-app-scripts in 998e530

mediremi (approved):

mediremi:
I think we need to move `react` to a peer dependency or change its version to `^16`

mediremi:
There's currently two versions of react due to the strict version in `package.json`, which means `d2-app-scripts build` is failing.

ismay:
> There's currently two versions of react due to the strict version in `package.json`, which means `d2-app-scripts build` is failing.

Yeah I've noticed a mix of pinned and `^` versions throughout the lib. I'm not sure why it's mixed, or what the underlying rules are there.

ismay:
I've removed the cli-app-scripts update btw. If it were painless I wouldn't mind including it, but since it's causing some issues I've reverted to just the app-runtime update.

mediremi:
Having `cli-app-scripts` prevent builds with multiple versions of sensitive dependencies is the point though - we want to be able to catch and prevent issues such as having two versions of react in `ui` (and hence two versions of react in some apps).

ismay:
> Having `cli-app-scripts` prevent builds with multiple versions of sensitive dependencies is the point though - we want to be able to catch and prevent issues such as having two versions of react in `ui` (and hence two versions of react in some apps).

Yeah definitely. I'm not against update itself, I think that's a good idea. I just don't want it to block the current update. I'd prefer to solve that in a separate PR.

PR #846, opened 2021-11-16
chore(deps): update cli-app-scripts

ismay:
Variation with `manypkg fix` here: https://github.com/dhis2/ui/pull/848


ismay:
So the crux seems to be that react/react-dom is resolving to 16.14.0 for cli-app-scripts -> app-shell. The range there is not fundamentally different from the range allowed inside of ui though: https://github.com/dhis2/app-platform/blob/master/shell/package.json#L23-L24.

I don't see why all of them aren't resolving to 16.14.0.

mediremi:
Looks like react `16.13` is still in `yarn.lock` because of `storybook/package.json`, `icons/package.json`, `collections/ui/package.json`, and `components/*/package.json`

ismay:
> Looks like react `16.13` is still in `yarn.lock` because of `storybook/package.json`, `icons/package.json`, `collections/ui/package.json`, and `components/*/package.json`

I ran a yarn upgrade with manypkg that should have taken care of all of them, but apparently that didn't work.

PR #865, opened 2021-11-29
feat(dropdown-button): add controlled mode

HendrikThePendric:
The idea is as follows:
- If an `open` prop is provided this component goes into "controlled mode", so the parent component can determine when the `Popper` should render.
- If that's the case, the `onClick` becomes a required prop, because that callback will be needed.

Note:
I noticed one strange thing, only in the tests, which I couldn't really explain and I have had to work around it. I have left a code comment to clarify [here](https://github.com/dhis2/ui/pull/865#discussion_r758343075).

### DIFF
--- a/components/button/src/dropdown-button/dropdown-button.js
+++ b/components/button/src/dropdown-button/dropdown-button.js
@@ -68,7 +68,7 @@ class DropdownButton extends Component {
     }
     anchorRef = React.createRef()
 
-    controlledOnClick = ({ name, value } = {}, event) => {
+    controlledOnClick = ({ name, value }, event) => {
         this.props.onClick(
             {
                 name,
--- a/components/button/src/dropdown-button/dropdown-button.js
+++ b/components/button/src/dropdown-button/dropdown-button.js
@@ -1,5 +1,6 @@
 import { Layer } from '@dhis2-ui/layer'
 import { Popper } from '@dhis2-ui/popper'
+import { requiredIf } from '@dhis2/prop-types/build/cjs/propTypes'
 import { spacers, sharedPropTypes } from '@dhis2/ui-constants'
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
@@ -67,7 +68,18 @@ class DropdownButton extends Component {
     }
     anchorRef = React.createRef()
 
-    onToggle = ({ name, value }, event) => {
+    controlledOnClick = ({ name, value } = {}, event) => {
+        this.props.onClick(
+            {
+                name,
+                value,
+                open: !this.props.open,
+            },
+            event
+        )
+    }
+
+    toggleOpenState = ({ name, value }, event) => {
         this.setState({ open: !this.state.open }, () => {
             if (this.props.onClick) {
                 this.props.onClick(
@@ -83,7 +95,6 @@ class DropdownButton extends Component {
     }
 
     render() {
-        const { open } = this.state
         const {
             component,
             children,
@@ -102,7 +113,13 @@ class DropdownButton extends Component {
             initialFocus,
             dataTest,
         } = this.props
-
+        const isControlled =
+            typeof this.props.open === 'boolean' &&
+            typeof this.props.onClick === 'function'
+        const open = isControlled ? this.props.open : this.state.open
+        const onClick = isControlled
+            ? this.controlledOnClick
+            : this.toggleOpenState
         const ArrowIconComponent = open ? ArrowUp : ArrowDown
 
         return (
@@ -116,7 +133,7 @@ class DropdownButton extends Component {
                     primary={primary}
                     secondary={secondary}
                     small={small}
-                    onClick={this.onToggle}
+                    onClick={onClick}
                     name={name}
                     value={value}
                     tabIndex={tabIndex}
@@ -128,7 +145,7 @@ class DropdownButton extends Component {
                 </Button>
 
                 {open && (
-                    <Layer onClick={this.onToggle} transparent>
+                    <Layer onClick={onClick} transparent>
                         <Popper
                             dataTest={`${dataTest}-popper`}
                             placement="bottom-start"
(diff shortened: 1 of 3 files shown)


components/button/src/dropdown-button/dropdown-button.js
```diff
@@ -67,7 +68,18 @@ class DropdownButton extends Component {
     }
     anchorRef = React.createRef()
 
-    onToggle = ({ name, value }, event) => {
+    controlledOnClick = ({ name, value } = {}, event) => {
```
HendrikThePendric, line 71: I've had to add a default value for the first argument here. To prevent the error in the screenshot below. It seems to only be needed by the Jest tests though, I could replicate this behaviour in the story.

In theory, when clicking the backdrop, the first arg (`payload`) should be `{}` ([see `Layer`](https://github.com/dhis2/ui/blob/afc6eb5ef84a9796330b7eeac8addd2dd337f786/components/layer/src/layer.js#L9), but for some reason in the Jest test this is not true.
Screenshot 2021-11-29 at 13 54 21

So while this default arg value fixes the problem, I don't fully understand the reason this problem occurred in the first place, which is slightly unnerving.

I just wanted to point this out.

varl: Solved in dcd9098

components/button/src/dropdown-button/dropdown-button.stories.js
```diff
@@ -89,3 +89,25 @@ InitialFocus.args = { initialFocus: true }
  * not the normal 'canvas' story viewer)
  */
 InitialFocus.parameters = { docs: { disable: true } }
+
+const ControlledTemplate = (args) => {
```
varl, line 93: I think that we could drop the interactivity here (for the state) as we are trying to demonstrate a controlled component, and right now the uncontrolled and controlled behave the same way even if the implementation is different.

It would be enough to have a demo with the `open` prop set to `true`, and allow it to be toggled through the "props" panel in Storybook.

Toggling the prop currently does nothing in the story: https://pr-865--dhis2-ui.netlify.app/?path=/story/actions-buttons-dropdown-button--controlled&args=open:true

components/button/src/dropdown-button/dropdown-button.js
```diff
@@ -187,8 +206,12 @@ DropdownButton.propTypes = {
     /**
      * Callback triggered on click.
      * Called with signature `({ name: string, value: string, open: bool }, event)`
+     * Is required when using this component in controlled mode, i.e. when
      */
-    onClick: PropTypes.func,
+    onClick: requiredIf(
+        (props) => typeof props.open === 'boolean',
+        PropTypes.func
+    ),
```
varl, line 214: I don't think that `onClick` needs to be required when using the component in a controlled sense.

If the consumer passes in the `open` prop then the component is controlled, full stop.

Making the `onClick` function a requirement when the component is controlled just makes the API more complex.

components/button/src/dropdown-button/dropdown-button.js
```diff
@@ -67,7 +68,18 @@ class DropdownButton extends Component {
     }
     anchorRef = React.createRef()
 
-    onToggle = ({ name, value }, event) => {
+    controlledOnClick = ({ name, value }, event) => {
```
varl, line 71: Having one handler that does both uncontrolled/controlled flows seems a bit simpler to me. :thinking:

varl (changes requested):
I have been thinking about this for a bit, and I think we can simplify the API by omitting the `requiredIf` from the `onClick` function.

Suggested changes:

- If `open` is used, the component is controlled.
- `isControlled` checks if `open` is a boolean: `typeof this.props.open === 'boolean'`
- `onClick` is always optional.
- use one click handler that manages both uncontrolled/controlled flows.

Here are all my suggestions in code form: https://github.com/dhis2/ui/pull/872

components/button/src/dropdown-button/dropdown-button.js
```diff
@@ -1,5 +1,6 @@
 import { Layer } from '@dhis2-ui/layer'
 import { Popper } from '@dhis2-ui/popper'
+import { requiredIf } from '@dhis2/prop-types/build/cjs/propTypes'
```
varl, line 3: Looks like the IDE auto-import screwed up the import.

varl:
I've reconsidered my position on the utility of the `requiredIf` prop. @HendrikThePendric made a good case for it on Slack:

> I often think of prop-type-warnings as developer hints rather than an errors. At the end of the day, it is simply not possible to implement a fully-functional drop-down button without passing an `onClick`, and I personally would appreciate that hint if I forgot it.

It is also easier for us to remove this requirement without a breaking change if we need to loosen it.

I've updated my suggestions in #872 accordingly.

varl (approved):

components/button/src/dropdown-button/dropdown-button.js
```diff
@@ -65,25 +66,32 @@ class DropdownButton extends Component {
...
             }
-        })
+        }
+        if (typeof this.props.open === 'boolean') {
+            handleClick(!this.props.open)
+        } else {
+            this.setState({ open: !this.state.open }, () => {
+                handleClick(this.state.open)
+            })
+        }
```
Mohammer5, line 91: The following seems more parsable to me:

```js
    onClickHandler = ({ name, value }, event) => {
        if (!this.props.onClick) return

        const open = typeof this.props.open === 'boolean'
            ? this.props.open
            ? this.state.open

        this.props.onClick({ name, value, open }, event)
    }
```

Mohammer5 (approved):
Just one minor thing, nothing functional or critical, so I approve this PR with a change proposal

components/button/src/dropdown-button/dropdown-button.js
Mohammer5: Oops.. Nevermind, I didn't see the `this.setState`..
Resolving this one now

PR #872, opened 2021-12-01
refactor(dropdown-button): single click handler

varl:
Review of: #865.


HendrikThePendric (approved):
We discussed on Slack that the value

PR #875, opened 2021-12-01
feat: modal close button

cooper-joe:
Implements [LIBS-174](https://jira.dhis2.org/browse/LIBS-174).

This PR adds a close button to any `Modal` with an `onClose` function. 

## Issues
- [x] **Test failing**. I added a test to check the close button functionality, but I'm not able to get it to pass. Guidance would be appreciated.


Mohammer5:
The close button and optional title are very close to each other (could even overlap I think)

image

cooper-joe:
> The close button and optional title are very close to each other (could even overlap I think)

Yes, the boundaries for these elements do overlap: 
image

I thought this was an OK tradeoff because:
- positioning the X button far enough away from the `ModalTitle` boundary creates an odd looking imbalance.
- `ModalTitle` renders text. The overlap, while technically not ideal, doesn't harm the experience. Text in the title is selectable and is never overlapped by the X icon, only the clickable frame of the button. The button itself is interactive all over and is not blocked by the title.

Interested to hear what others think about this tradeoff.

Mohammer5:
If we set right/top to `0`, then it's align perfectly with the rest of the contents:
image

The button has a width of 24 pixels and padding of the modal container is 24 pixels as well:
image

cooper-joe:
> Of course this isn't the nicest way to position the button, but IMO that'd be a better trade-off than potential overlapping with / zero margin to the text

Yeah, I think you're right @Mohammer5. I also did some further testing and found some cases where it did look clunky when the two elements overlapped.

I've just pushed a change that positions it `top: 0` and `right: 0`. While it does look _a bit_ odd right up there, it's a pragmatic fix for something that is needed, so I'm happy with it.

Mohammer5 (approved):
I'm wondering why the netlify preview is not changing..

varl:
@Mohammer5 good question, the workflow hasn't run. :thinking:

cooper-joe:
Does the workflow need to be fixed before this can be approved? Otherwise, any other issues blocking merge?

varl (approved):

PR #881, opened 2021-12-03
feat(context selection): add components & stories

Mohammer5:
- [x] implementation
- [x] tests
- [x] finalize design details


cooper-joe:
When this component was used in Data Approval app, @HendrikThePendric suggested avoiding the _context_ term because it could be confused with other _context_ s present.

We struggled to find a generic enough name, [in the end landing on _Top bar_](https://github.com/dhis2/approval-app/tree/master/src/top-bar).

For a `ui` component, _Top bar_ is too generic. Perhaps _Selection bar_ / _Selection toolbar_  could work. It implies that it's a toolbar for selections, but doesn't specify anything about the type, number or consequence of those selections.

Mohammer5:
@cooper-joe This will export two components, what about the following names (according to your suggestions):
* `SelectorBar`
* `SelectorBarItem` (right now it's called `ContextSelector`)

cooper-joe:
I think the `Selector` part works well. Already today though we talked about how this component could be flexible enough to have different display modes, so I don't think it's wise to use the display mode (bar) in the name. To make it as generic as possible, how about `SelectorGroup` and `SelectorGroupItem`?

Mohammer5:
@cooper-joe That sounds good to me!

HendrikThePendric (commented):
The code is looking good, but I do have some comments:

----
Although this is very subjective, I don't think the component names we've landed on are very descriptive. For the container component, I would prefer `SelectorBar` as @Mohammer5 suggested already. @cooper-joe mentioned this:
> Already today though we talked about how this component could be flexible enough to have different display modes, so I don't think it's wise to use the display mode (bar) in the name.

But as I recall we discussed these display modes could either be a top**bar** or a side**bar**, so I would argue `SelectorBar` (or `SelectionBar`) is a perfect name.

As for the items, to me these are clearly some form of dropdown-button, so I'd call them something like `SelectorBarDropdown` / `SelectorBarDropdownButton`

----

In the stories we see quite a few examples with a single select. I haven't looked at the design specs, and haven't participated in any discussions about this component, so I might be missing stuff here.... But I would not have expected a single select as a typical child for the dropdown. Instead I would have expected a (compact) `Menu` with `MenuItem`s. I think it would be better to change the stories so that they feature more typical use cases. 

Bonus points for already including the org-unit-tree in the stories though ;)

----

How the right hand side content has been implemented is super flexible, so that's great. I wonder if perhaps we should rename the prop so we don't get into trouble later down the line if/when we start implementing RTL support. Maybe sth like `additionalContent`?

cooper-joe:
> But as I recall we discussed these display modes could either be a topbar or a sidebar, so I would argue SelectorBar (or SelectionBar) is a perfect name.

Yeah, that's a good point @HendrikThePendric! I prefer `bar` over `group` as well. If we're all happy that `bar` is flexible enough, then I'm happy to go for that.

Also agreed with the points about the `Menu` examples and the `additionalContent` name 👍

HendrikThePendric (approved):

cooper-joe (approved):

PR #885, opened 2021-12-03
fix(modal): reduce spacing between elements

cooper-joe:
Fixes [LIBS-277](https://jira.dhis2.org/browse/LIBS-277).

This PR reduces margins and changes how they are applied. Instead of applying top/bottom margins to `ModalContent`, the respective margins are applied to `ModalTitle` and `ModalActions`. If these elements aren't present, those margins aren't needed (the `Modal` itself has frame padding).


ismay (approved):
Looks good 👍!

varl (approved):

varl:
I'd call this a fix over a feature.

PR #895, opened 2021-12-08
fix(button): secondary button contrast

cooper-joe:
Fixes [LIBS-280](https://jira.dhis2.org/browse/LIBS-280).

This PR adjusts the `Button` `secondary` style to be more visible on various background colors.

| Before | After |
| --- | --- |
| image | image |

Notes
---
**This PR does not fix the issue of this button type being unusable on a dark background**. This is out of scope of this change. Adding a light background to the `secondary` type makes it indistinguishable from the `basic` variant. Potential solutions are an `invert` property, or `dark` theme support.

### DIFF
--- a/components/button/src/button/button.styles.js
+++ b/components/button/src/button/button.styles.js
@@ -118,13 +118,13 @@ export default css`
     }
 
     .secondary {
-        border-color: ${colors.grey400};
+        border-color: rgba(74, 87, 104, 0.25);
         background-color: transparent;
     }
 
     .secondary:hover {
-        border-color: ${colors.grey400};
-        background-color: rgba(160, 173, 186, 0.08);
+        border-color: rgba(74, 87, 104, 0.5);
+        background-color: rgba(160, 173, 186, 0.05);
     }
 
     .secondary:active,
@@ -138,7 +138,7 @@ export default css`
     }
 
     .secondary:disabled {
-        border-color: ${colors.grey400};
+        border-color: rgba(74, 87, 104, 0.25);
         background-color: transparent;
         box-shadow: none;
         color: ${theme.disabled};


varl (approved):

Mohammer5:
There are no technical issues and I understand that the new colors make more sense from the design perspective.
I'm wondering though if that's the correct approach (using colors that are not part of the spec'ed colors).

* Are these colors going to be in the colors collection? Then I'd be in favor of adding them there and use them in the changed file (I suppose this could require @cooper-joe to request some technical help as these colors use opacity as well, I'm not sure how we could add that to the colors and css variables right now).
* If not, these are non-standard, then why use them in the first place (which will make creating/adding custom themes more complicated)?

varl:
We use specialised colour values in places where the reference colours do not cut it. To name a few examples where we use non-reference colours: Button, Table, TableCell, TableCellHead, Chip, DataTable, HeaderBar, and Transfer.

Correct me if I'm wrong, @cooper-joe, but the reference colours are not intended to be exclusive.

Potential theme support is not a priority (the `theme` variables is a drag of the past) as we do not have any clear use cases for it right now. When it becomes a priority, there needs to be a significant clean-up across the entire UI library to make it work right so I don't think we need to open that can of worms in this PR.

cooper-joe:
That's right @varl, the `ui` colors are not intended to cover every use case. This use case in particular, semi-transparent color usage, is not covered and needs to be tweaked by hand to ensure good enough contrast.

Regarding the difficulty of theming, as @varl says, this is not a priority and will require a refactor of variables anyway, so it shouldn't stop fixes like this right now.

---
### Aside

The color used is _kind of_ part of the `ui` `colors`. It is `grey700` used at `25%`/`50%` opacity. CSS variables set in hex cannot be maniuplated, so that's why it's done using a hardcoded `rgba()`. 
Having a read, what we _could_ do is define a second set of rgb variants of the variables, like `--grey700-rba: 74, 87, 104`, and then reference it as `border: rgba(var(--grey700-rgb), 0.25);`. I don't think is necessary right now, but could be a solution to this problem when we cover theming.

Mohammer5 (approved):

varl:
Thanks @cooper-joe and a very good question, @Mohammer5 ! This isn't clarified anywhere in our docs; I checked.

Mohammer5:
Let's document it somewhere! Maybe @cooper-joe could document this in the appropriate place (do we have dedicated docs for the colors?)

cooper-joe:
Yeah we have these [color guidelines](https://github.com/dhis2/design-system/blob/master/principles/color.md), but I'm thinking it might be best to document this is generic _Using constants_ document. The concept applies also to `spacers`, `elevations` and so on. Will make a note to document, good idea 👍

PR #897, opened 2021-12-08
chore: add SelectorBar to ui collection

Mohammer5:
Question: @varl how do I make this PR make the workflows publish a new version? All the commits are chores


varl:
There has to be a fix or feat commit.

storybook/config/preview.js
```diff
@@ -64,16 +64,18 @@ export const parameters = {
...
-            <>
-                <Title />
-                <Subtitle />
-                <Description />
-                <Primary />
-                <ArgsTable story={PRIMARY_STORY} />
-                <Stories title="Examples" />
-            </>
-        ),
+        page: function Page() {
```
varl, line 67: Out of curiousity, why the change to a named function ?

varl (approved):
We could consider it a bug that we forgot to export it in the initial feat release (7.10.0) and release this as a patch.

Then they would get this fix even if their version range on UI is strict.

storybook/config/preview.js
Mohammer5: I got a prettier error.. Dunno why

varl: Huh.

Mohammer5:
> We could consider it a bug that we forgot to export it in the initial feat release (7.10.0) and release this as a patch.

I reworded the second commit from `feat` to `bug`

Birkbjo (approved):

PR #908, opened 2021-12-14
fix: focus styles

cooper-joe:
Fixes [UX-20](https://jira.dhis2.org/browse/UX-20). 

This PR fixes focus styles across the library. Focus styles were previously updated, but there were issues where the focus outline was cut off by the parent container. All focus styles are now contained within the element.

For example, issues like:
image

--- 

## Screenshots

| Component | Before | After |
| --- | --- | --- |
| `Button` | image | image |
| `Button secondary` | image |  image |
| `Button primary` | image | image  |
| `Button destructive` | image | image |
| `Checkbox` | image | image |
| `Radio` | image | image |
| `Input` / `Select` | image | image |
| `Segmented control` | image | image |
| `Switch` | image | image |

_Note: The focus styles appear outside of the `Checkbox`, `Radio`, and `Switch`, but are actually inside the bounds of the element._

### DIFF
--- a/components/text-area/src/text-area/text-area.styles.js
+++ b/components/text-area/src/text-area/text-area.styles.js
@@ -29,7 +29,8 @@ export const styles = css`
 
     textarea:focus {
         outline: none;
-        box-shadow: 0 0 0 3px ${theme.focus};
+        box-shadow: inset 0 0 0 2px ${theme.focus};
+        border-color: ${theme.focus};
     }
 
     textarea.valid {
--- a/components/button/src/button/button.styles.js
+++ b/components/button/src/button/button.styles.js
@@ -33,14 +33,22 @@ export default css`
 
     button:focus {
         outline: 3px solid ${theme.focus};
-        outline-offset: 2px;
+        outline-offset: -3px;
         transition: none;
+        text-decoration: underline;
+    }
+
+    /* Prevent focus styles when mouse clicking */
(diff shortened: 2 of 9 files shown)


mediremi (approved):

components/text-area/src/text-area/text-area.styles.js
```diff
@@ -29,7 +29,8 @@ export const styles = css`
 
     textarea:focus {
         outline: none;
-        box-shadow: 0 0 0 3px ${theme.focus};
+        box-shadow: inset 0 0 0 2px ${theme.focus};
```
ismay, line 32: Should the spread radius here (the 2px) not be 3px?

HendrikThePendric: I think this is correct, he's using an inset box-shadow with a width of 2px plus a blue border of 1px, so the total visual width is 3px.

HendrikThePendric (approved):

components/text-area/src/text-area/text-area.styles.js
cooper-joe: That's right, the `box-shadow` and `border` combine to make 3px.

ismay: Ah nice. Yeah I assumed it was intentional since it was changed from 3 to 2, but it stood out to me so I thought I'd ask.

PR #916, opened 2022-01-11
fix: remove focus outline from disabled select components

HendrikThePendric:


mediremi (changes requested):
Should the input being wrapped also receive the `disabled` prop? Hiding the focus styles when the input is receiving focus would be confusing for keyboard users.

HendrikThePendric:
> Should the input being wrapped also receive the `disabled` prop? Hiding the focus styles when the input is receiving focus would be confusing for keyboard users.

As discussed on Slack, there isn't really an `input` tag there, but setting the `tabindex` of the `div` pretending to be an `input` to `-1` has a similar effect, so that'\s what I did in 3110111 .

Mohammer5 (approved):
LGTM

mediremi (approved):

PR #921, opened 2022-01-12
fix: reduce size of date/time icon in Chrome (LIBS-212)

martinkrulltott:
Resolves [LIBS-212](https://jira.dhis2.org/browse/LIBS-212), which is a Chrome-specific issue caused by the icon added for `date` and `time` type inputs. The size of the icon has been reduced to fit properly within a `dense` input.

Live example here: https://codesandbox.io/s/chrome-date-time-picker-height-x1f6l?file=/src/styles.css

Example of `text` compared to `date` / `datetime-local` / `time`:

image

### DIFF
--- a/components/input/src/input/input.js
+++ b/components/input/src/input/input.js
@@ -82,6 +82,15 @@ const styles = css`
         border-color: ${theme.focus};
     }
 
+    input[type="date"]::-webkit-inner-spin-button,
+    input[type="date"]::-webkit-calendar-picker-indicator,
+    input[type="time"]::-webkit-inner-spin-button,
+    input[type="time"]::-webkit-calendar-picker-indicator,
+    input[type="datetime-local"]::-webkit-inner-spin-button,
+    input[type="datetime-local"]::-webkit-calendar-picker-indicator  {
+        height: 14px;
+    }
+
     input.warning {
         border-color: ${theme.warning};
     }
--- a/components/input/src/input/input.js
+++ b/components/input/src/input/input.js
@@ -82,13 +82,21 @@ const styles = css`
         border-color: ${theme.focus};
     }
 
-    input[type="date"]::-webkit-inner-spin-button,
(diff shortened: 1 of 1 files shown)


components/input/src/input/input.js
```diff
@@ -82,6 +82,23 @@ const styles = css`
...
+    input[type='datetime-local']::-webkit-calendar-picker-indicator {
+        height: 14px;
+        padding-top: 1px;
+        padding-bottom: 1px;
+    }
+
+    input[type='date']::-webkit-datetime-edit,
+    input[type='datetime-local']::-webkit-datetime-edit,
+    input[type='time']::-webkit-datetime-edit {
+        line-height: 14px;
```
HendrikThePendric, line 99: I'm assuming this is to style the placeholder text e.g. `mm / dd / yyyy`, if this is not correct you can stop reading because the text below is based on this assumption.

It seems wrong to adjust the placeholder line-height to a value other than the regular line-height (16px). It would be good to fix this issue in another way. If the font and font-size of this placeholder and the regular text are the same (which they should be), then surely something else must be causing the placeholder to take up too much height. I don't think we are tackling the root cause here.

HendrikThePendric (commented):
I'm not quite sure about this one.

components/input/src/input/input.js
HendrikThePendric: I think the root cause is the padding on the placeholder wrapper. Setting that to 0 fixes the height issue too (only tested in the devtools).

HendrikThePendric, line 99:
```suggestion
    input[type='date']::-webkit-datetime-edit-fields-wrapper,
    input[type='datetime-local']::-webkit-datetime-edit-fields-wrapper,
    input[type='time']::-webkit-datetime-edit-fields-wrapper {
        padding: 0;
```

HendrikThePendric (approved):

mediremi (approved):

PR #930, opened 2022-01-13
feat(org unit tree): expose <RootError /> & <RootLoading />

Mohammer5:
I've changed the data-test value of the `RootError` component.
It could be considered a bug fix as it was the same as the data test value of the `RootLoading` component..

### DIFF
--- a/collections/ui/src/index.js
+++ b/collections/ui/src/index.js
@@ -49,6 +49,8 @@ export { Node } from '@dhis2-ui/node'
 export { NoticeBox } from '@dhis2-ui/notice-box'
 export {
     OrganisationUnitTree,
+    OrganisationUnitTreeRootError,
+    OrganisationUnitTreeRootLoading,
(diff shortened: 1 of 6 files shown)


collections/ui/src/index.js
```diff
@@ -49,6 +49,8 @@ export { Node } from '@dhis2-ui/node'
 export { NoticeBox } from '@dhis2-ui/notice-box'
 export {
     OrganisationUnitTree,
+    RootError,
+    RootLoading,
```
mediremi, line 53: Should we prefix these exports?

```suggestion
    RootError as OrganisationUnitTreeRootError,
    RootLoading as OrganisationUnitTreeRootLoading,
```

Mohammer5: Good idea

mediremi (approved):

Birkbjo (approved):

PR #941, opened 2022-01-20
fix(status-icon): only apply margin between input and status icon if icon rendered

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-80

Replaces the margin between various inputs and their status icons with a flexbox gap. Also extracts the `StatusIcon` component.


HendrikThePendric (approved):
Great!

varl (approved):

PR #943, opened 2022-01-21
feat(chip): allow custom margin

cooper-joe:
This PR allows custom margin values to be passed to the `Chip`. The implementation is based upon the proposal below.

---

## Background
Some UI components apply a default margin. For example, the `chip` has a default `margin: 4px`. This was added to provide a reasonable default for the expected usage of a group `chip`s. The alternative, applying no default margin, would mean that in the majority of cases the consumers of UI would need to wrap the `chip` and apply margins.

## Problem
The default margins applied to some components are not always appropriate. In many cases they are, but in other cases they make the component less flexible. This leads to workarounds that are not recommended, like using `className`. @varl and I have discussed alternatives and have a proposed solution.

## Proposed solution
Define a standard way for components to expose a set of `margin` props. Though this might be defined as a _global_ prop, it is not required for every component. It should be manually included for any component where a custom `margin` value may provide extended flexibility.

The proposed props would be:
```
margin: PropTypes.number,
marginBottom: PropTypes.number,
marginLeft: PropTypes.number,
marginRight: PropTypes.number,
marginTop: PropTypes.number,
```

Implementation in a components styles looks like:
```
margin: ${margin}px;
${marginBottom && `margin-bottom: ${marginBottom}px;`}
${marginLeft && `margin-left: ${marginLeft}px;`}
${marginRight && `margin-right: ${marginRight}px;`}
${marginTop && `margin-top: ${marginTop}px`}
```

### Alternatives / Other discussions
- Use `PropTypes.string` instead to allow full css declarations to be passed in. 
  - Values are intentionally restricted to pixels. Though allowing a full css declaration to be passed in would be the most flexible, that is not the intended use of this feature. If full custom css declarations are required (like `margin: auto 1rem 50% unset`), then a wrapper element should be used.
-  Automatically apply the `margin` props to all components
   - Potentially useful, but not necessary to start using this pattern. Also, `margin` is not always an appropriate prop to expose.
- Use a `base` wrapper component with a `margin` prop that wraps every other component. Thus, a `margin` prop doesn't need to be added to any other component.
  - Creates an unecessary additional wrapping element, which creates other issues for accessibility, layout and so on. Doesn't seem worth the tradeoff for the convenience it offers.

# Update

After discussion, the decision was made to defer adding the generic `margin` prop and instead add the 4 direction-specific `margin...` props. This avoids any confusion about specificity and override behaviour. If the need for a generic `margin` prop is shown, it can be added in the future.

### DIFF
--- a/components/chip/src/chip.js
+++ b/components/chip/src/chip.js
@@ -1,4 +1,4 @@
-import { colors, theme, spacers } from '@dhis2/ui-constants'
+import { colors, theme } from '@dhis2/ui-constants'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React from 'react'
@@ -18,6 +18,10 @@ const Chip = ({
     onClick,
     icon,
     dataTest,
+    marginBottom,
+    marginLeft,
+    marginRight,
+    marginTop,
 }) => (
     <span
         onClick={(e) => {
@@ -42,7 +46,6 @@ const Chip = ({
                 display: inline-flex;
                 align-items: center;
                 height: 32px;
-                margin: ${spacers.dp4};
                 border-radius: 16px;
                 background-color: ${colors.grey200};
                 font-size: 14px;
@@ -92,11 +95,23 @@ const Chip = ({
                     0 1px 5px 0 rgba(0, 0, 0, 0.12);
             }
         `}</style>
+        <style jsx>{`
+            span {
+                ${marginBottom && `margin-bottom: ${marginBottom}px;`}
+                ${marginLeft && `margin-left: ${marginLeft}px;`}
+                ${marginRight && `margin-right: ${marginRight}px;`}
+                ${marginTop && `margin-top: ${marginTop}px`}
+            }
+        `}</style>
     </span>
 )
 
 Chip.defaultProps = {
     dataTest: 'dhis2-uicore-chip',
+    marginBottom: 4,
+    marginLeft: 4,
+    marginRight: 4,
+    marginTop: 4,
 }
 
 Chip.propTypes = {
@@ -107,6 +122,14 @@ Chip.propTypes = {
     disabled: PropTypes.bool,
     dragging: PropTypes.bool,
     icon: PropTypes.element,
+    /** `margin-bottom` value, applied in `px`  */
+    marginBottom: PropTypes.number,
+    /** `margin-left` value, applied in `px`  */
+    marginLeft: PropTypes.number,
+    /** `margin-right` value, applied in `px`  */
+    marginRight: PropTypes.number,
+    /** `margin-top` value, applied in `px`  */
+    marginTop: PropTypes.number,
     overflow: PropTypes.bool,
     selected: PropTypes.bool,
     onClick: PropTypes.func,


mediremi (approved):

Mohammer5 (approved):
The only "issue" I have with this is the default `margin` prop value.
It will force the app do so something like this:

```js
<Chip margin={0} marginRight={4} />
```

And that looks a bit odd. Maybe removing all single-direction prop makes more sense here.
But it's not a biggie whatsoever, code-wise this looks good, so I'll approve and hope someone else (@dhis2/team-platform) will comment on this OR simply approves it as well

ismay:
> The only "issue" I have with this is the default `margin` prop value. It will force the app do so something like this:
> 
> ```js
> <Chip margin={0} marginRight={4} />
> ```
> 
> And that looks a bit odd. Maybe removing all single-direction prop makes more sense here:
> 
> ```js
> <Chip margin="0 4px 0 0" />
> ```
> 
> But it's not a biggie whatsoever, code-wise this looks good, so I'll approve and hope someone else (@dhis2/team-platform) will comment on this OR simply approves it as well.

I agree. If props are mutually exclusive it shouldn't be possible to use them both at the same time on a component. That's confusing API in my opinion.

cooper-joe:
> The only "issue" I have with this is the default `margin` prop value. It will force the app do so something like this:
> 
> ```js
> <Chip margin={0} marginRight={4} />
> ```
> 
> And that looks a bit odd.

The app isn't forced to do that, because they can also do: 
```
<Chip MarginRight={4} MarginTop={0} MarginBottom={0} MarginLeft={0} />
```

The `Margin` prop, as I see it, is a utility or shorthand. Perhaps the specifity needs to be made clearer, as the single-direction props will always override the generic.

> If props are mutually exclusive it shouldn't be possible to use them both at the same time on a component

Are they mutually exclusive though? `Margin` still applies to the other 3 directions if `MarginRight` is set.

>
> And that looks a bit odd. Maybe removing all single-direction prop makes more sense here:
> 
> ```js
> <Chip margin="0 4px 0 0" />
> ```

This allows any kind of css declaration, like `0 5rem auto inherit`, and I'm not sure that's a good thing (this is mentioned in the original comment, under _Alternatives_). Though it's ultimately more flexible, it's not how these components are intended to be used. Restricting to pixel values reinforces the "if you need some complex margins, you should be wrapping this in a layout component like a `Box`."

---

Is there another prop approach where margins are limited to pixel values? The only alternative I can see that avoids what's outlined above is omitting the `Margin` and only including direction-specific margin props.

Mohammer5:
Sorry I closed this accidentally.. Reopening

cooper-joe:
> ...but <Chip marginRight={4} margin={0} /> reads like margin-right: 4px; margin: 0; but still behaves like the former example. This is the potentially confusing part.

Yeah, I can see how this is confusing. I'm trying to think of a way to imply the behaviour through the prop names. Like `marginBase` instead of just `margin`, might imply that a single-directional prop will override the "base". But still, even that wouldn't be without issue.

So, there are two options:
1. Include generic `margin` and ensure API documentation explains the specificity override behaviour.
2. Remove the generic `margin` and use single-directional props only.

I personally vote for 1, as I think the convenience is worth any confusion in the override behaviour. But I'm mostly concerned about making this functionality available, so I'm OK with either option.

Mohammer5:
> Include generic margin and ensure API documentation explains the specificity override behaviour.

I think this is the correct way. We allow people to do whatever they want but have very strict guidelines as to what is part of the official dhis2 design specs and how things should be used.

cooper-joe:
> I think this is the correct way. We allow people to do whatever they want but have very strict guidelines as to what is part of the official dhis2 design specs and how things should be used.

👍 I've added some notes to the props here, and I'll also make sure to cover the margins in the usage guidelines.

If there are no other objections, I'm going to go ahead and merge.

cooper-joe:
Update after an offline discussion: @Mohammer5 was on-board with the idea of dropping the generic `margin` prop in favour of only having the 4 direction-specific props. The previous 3 comments were miscommunication between us.

I'm happy to move ahead with @Mohammer5's suggestion. Adding the 4 direction-specific props now does not preclude adding a generic `margin` helper prop in the future, if we think the benefits outweight the possible confusion caused by specificity overrides. I've updated this PR accordingly.

PR #977, opened 2022-02-07
fix(select): don't render circular loader as flex item in order to preserve dimensions

mediremi:
Closes https://jira.dhis2.org/browse/LIBS-283

- Ensures circular loader remains a circle even when loading text wraps
- Adds spacing between circular loader and loading text


KaiVandivier (approved):
LGTM 👍

varl (approved):

PR #978, opened 2022-02-08
fix(data-table): enable auto-height in flexbox container

HendrikThePendric:
In the new Event Reports app we display data using our `DataTable` and it’s got to be scrollable, plus adapt to the available width and height. The container is using flexbox, and after playing around with the element inspector a bit, I realised that if I add `flex-grow: 0;` to the `TableScrollBox`, I can get the correct behaviour without having to compute the available height and set it….

So, I am now considering adding that style rule to the component. Obviously, it’s only going to have an effect if it’s placed in a flexbox container, but I think that’s probably OK. If the container isn’t a flexbox, the style rule is a noop.

Could be I’m missing something. Any objections to adding this?

### DIFF
--- a/components/table/src/data-table/table-elements/table-scroll-box.js
+++ b/components/table/src/data-table/table-elements/table-scroll-box.js
@@ -14,6 +14,7 @@ export const TableScrollBox = forwardRef(
                     max-height: ${maxHeight};
                     max-width: ${maxWidth};
                     overflow: auto;
+                    flex-grow: 0;
                 }
             `}</style>
         </div>


mediremi:
`0` is the default value of `flex-grow` so in what kind of use case does adding that rule affect layout? Is it to override a previously existing style?

HendrikThePendric:
> `0` is the default value of `flex-grow` so in what kind of use case does adding that rule affect layout? Is it to override a previously existing style?

This is an excellent point. I retraced my steps in how I created the table scroll container with automatic height in the events reports app, and it turns out it wasn't this CSS rule that fixed things, but some changes I made to the container element's styles.

Closing this now, and sorry for wasting your time 😊

PR #993, opened 2022-02-22
feat(org unit tree): load children on expand only

Mohammer5:
Prevent org unit tree from loading data of nodes' children because it's unnecessary. Instead load the children when expanding the tree.

About the breaking change: There are two reasons why this PR introduces a breaking change:
1. The version on `master` calls the callback twice per node.
   * Once with incomplete data (`{ id: String, displayName: String }`)
   * Then with complete data: (`{ id: String, displayName: String, path: String, children: Object[] }`)
   * -> I considered this to be a bug because the `onChildrenLoaded` has been called before the children were loaded. I consider this a breaking change because apps who use this feature must have built a workaround for this behavior which might break the app after the update
2. The timing of the call has changed.
   * On `master` the callback gets called once the root nodes are loaded, even if they remain collapsed.
   * This means that apps might expect that the callback will be called after the tree has been rendered
   * The changes on this PR will make the children load only when expanding the node, so naturally, the `onChildrenLoaded` callback can only be called after the user manually opens a node (or when the app alters the `expanded` prop imperatively)
   * -> If apps expect the previous timing behavior, this would be a breaking change 

BREAKING CHANGE: The payload and behavior of the `onChildrenLoaded` prop have changed slightly! Previously the callback would be called for every child that's been loaded. Now the callback will be called once with the node that has the children, so fewer calls but with a payload with an org unit one level above of what would be that case before this change.


components/organisation-unit-tree/src/__stories__/development-stories.js
```diff
@@ -80,7 +80,7 @@ export const DxWithRealBackend = () => (
         <DataProvider baseUrl="https://debug.dhis2.org/dev" apiVersion="">
             <Wrapper
                 //initiallyExpanded={['/ImspTQPwCqd/eIQbndfxQMb']}
-                suppressAlphabeticalSorting
+                // suppressAlphabeticalSorting
```
ismay, line 83: I'd just remove this if it's no longer necessary.

components/organisation-unit-tree/src/organisation-unit-node/organisation-unit-node.js
```diff
@@ -32,6 +32,115 @@ const LoadingSpinner = () => (
...
+    if (!filter?.length) {
+        return orgChildren
+    }
+
+    return orgChildren.filter((child) => {
+        return isPathIncluded(filter, `${node.path}/${child.id}`)
+    })
+}
+
+export const OrganisationUnitNodeChildren = ({
```
ismay, line 45: This file is getting rather large. Personally I'd extract this component to a separate file.

components/organisation-unit-tree/src/organisation-unit-node/use-org-children.js
```diff
@@ -0,0 +1,56 @@
...
+ * @param {boolean} [options.withChildren]
+ * @returns {Object}
+ */
+export const useOrgChildren = ({
+    node,
+    suppressAlphabeticalSorting,
+    onComplete,
+}) => {
+    const { called, loading, error, data } = useDataQuery(ORG_DATA_QUERY, {
+        variables: { id: node.id },
```
ismay, line 28: This query will not respond to changes in the node. Is it necessary for it to do so? Because if so, a useEffect with the node as a dependency and a refetch that updates the node would address that.

components/organisation-unit-tree/src/organisation-unit-node/use-org-children.js
```diff
@@ -0,0 +1,56 @@
...
+
+        return suppressAlphabeticalSorting
+            ? orgUnit.children
+            : sortNodeChildrenAlphabetically(orgUnit.children)
+    }, [data, suppressAlphabeticalSorting])
+
+    useEffect(() => {
+        if (onComplete && orgChildren) {
+            // For backwards compatibility: Pass entire node incl. children
+            onComplete({ ...node, children: orgChildren })
```
ismay, line 51: It's possible that this callback will be called more often than just when data and suppressAlphabeticalSorting change as useMemo does not guarantee to be referentially stable. Is that ok here?

components/organisation-unit-tree/src/organisation-unit-node/use-org-children.test.js
```diff
@@ -0,0 +1,330 @@
+import { CustomDataProvider } from '@dhis2/app-runtime'
+import { renderHook } from '@testing-library/react-hooks'
+import React from 'react'
+import { useOrgChildren } from './use-org-children.js'
+
+describe('OrganisationUnitTree - useOrgChildren', () => {
+    // @TODO: Figure out why this is necessary at all...
+    const origError = console.error
+    jest.spyOn(console, 'error').mockImplementation((...args) => {
```
ismay, line 9: What's the error you're getting here? Because it might be that you're getting errors logged from react-query's logger (though hthat should be disabled).

components/organisation-unit-tree/src/organisation-unit-node/use-org-children.js
Mohammer5: The id will always remain the same, so `useEffect` + `refetch` is not necessary here

components/organisation-unit-tree/src/organisation-unit-node/use-org-children.test.js
ismay: I don't know if it's any help, and I don't see any obvious differences, but these tests aren't generating any warnings: https://github.com/dhis2/app-runtime/blob/master/services/data/src/react/hooks/useDataQuery.test.tsx. Maybe there's something useful there.

ismay:
Looks good! Left a couple comments on things that stood out. Other than that it seems good to merge to me. Might be good to double check the breaking change with someone else just to be sure everyone agrees.

components/organisation-unit-tree/src/organisation-unit-node/use-org-children.js
Mohammer5: Fixed

components/organisation-unit-tree/src/organisation-unit-node/use-org-children.test.js
Mohammer5: I was getting a log with "Warning: An update to TestComponent inside a test was not wrapped in act(...)."
I added something to the first `it` and removed the mocking

ismay (approved):

varl (changes requested):
The stories for the Org Unit Tree don't seem to work right anymore.

Only the root is visible and I cannot seem to toggle or expand in any story:

- https://pr-993--dhis2-ui.netlify.app/?path=/story/forms-organisation-unit-tree--custom-expanded-imperative-open
- https://pr-993--dhis2-ui.netlify.app/?path=/story/forms-organisation-unit-tree--expanded
- https://pr-993--dhis2-ui.netlify.app/?path=/story/forms-organisation-unit-tree--selected-multiple

What's going on here ?

Mohammer5:
@varl fixed the stories

varl:
>Previously the callback would be called for every child that's been loaded. Now the callback will be called once with the node that has the children, so fewer calls but with a payload with an org unit one level above of what would be that case before this change.

This is good for the changelog. Would you mind expanding on why we need to change this in the PR so we get more context about the change ?

Mohammer5:
@varl I added an explanation to the PR's description

varl (approved):

PR #994, opened 2022-02-24
refactor(docs): merge design system and ui

varl:
Related to: https://github.com/dhis2/design-system/pull/47

Follow-ups:

- ~Open question about the error message: https://github.com/facebook/docusaurus/discussions/6960~ we've worked around this for now


HendrikThePendric:
Just wondering about all these `API.md` files... I think these were auto-generated from the prop-type definitions. If so, then ideally they shouldn't be in source control. Is that a possibility?

varl:
> Just wondering about all these `API.md` files... I think these were auto-generated from the prop-type definitions. If so, then ideally they shouldn't be in source control. Is that a possibility?

Yes, it's possible. We discussed it and decided to commit them for the following reasons:

- They will show up in code reviews and make it easier to spot API changes that can be (unintentionally) breaking.
- They are useful when browsing the source code on GitHub.
- The documentation does not work without first building the API docs locally.

While I would like generated code to be kept out of source control, they are benefits to committing generated code:

- The generated code can (and should) be code reviewed to catch e.g. bugs with the generator (e.g. if svgr produces a different result between different versions, we can catch that before releasing).
- We wouldn't need to run `yarn setup` after `yarn install` to be able to run the project locally.

collections/ui/API.md
```diff
@@ -0,0 +1,2653 @@
+### AlertBar
+
+#### Usage
+
+**Note**: If possible, import the component from the main UI (`@dhis2/ui`) package.
+
```
HendrikThePendric, line 6: If possible, these notes should only appear in the component packages, not in the collection packages, since the import below is already correct.

HendrikThePendric:
> While I would like generated code to be kept out of source control, there are practical benefits to committing generated code

There are also some downsides to checking in the generated code, which I think we should also keep in mind ((the first point was discussed already during a meeting, but the second point is new):
1. By checking in the generated markdown files we are implicitly choosing to have the docs generated locally. So we are effectively making a human being responsible for remembering to generate the docs whenever a new prop or behaviour is added. The docs generation process takes a bit too long to include it as a pre-push step. So we probably have to invent some sort of check on CI that catches cases where a doc update is required but no docs have been generated. This is probably doable, but not something we would have to do if we let the files be generated during the CI process.
2. I also worry about merge conflicts in the `collections/ui/API.md` file. If one dev has implemented a feature in component A and another dev a feature in component B, then both PRs will produce changes to that collections markdown file. I guess most of the time it shouldn't cause any merge conflicts because the changes are in non-overlapping parts of the file. If there are merge conflicts, these shouldn't be too hard to resolve either. But still there will be conflicts from time to time, and we wouldn't have those if we chose to keep the generated code out of source control.

On balance, I think the arguments in favour of checking in these markdown files still outweigh the arguments against doing so. But if it turns out that it's hard to make some automated check for this, or we end up with a lot of merge conflicts all the time, we should probably reconsider.

varl:
> I also worry about merge conflicts in the collections/ui/API.md file.

I'll mention it here too:

The correct way to resolve conflicts in `collections/ui/API.md` is to do: `git checkout --theirs collections/ui/API.md` (`--ours` if rebasing) and then regenerate the API docs.

varl:
Will merge this to `beta` today.

amcgee:
@germanviscuso @HendrikThePendric I hacked a workaround for the big red error bar, let me know if we should merge this

docs/src/css/custom.css
```diff
@@ -4,6 +4,11 @@
  * work well for content-centric websites.
  */
 
+/* error box hack */
+div[style="text-align:center;color:#fff;background-color:#fa383e;border-color:#fa383e;border-style:solid;border-radius:0.25rem;border-width:1px;box-sizing:border-box;display:block;padding:1rem;flex:0 0 50%;margin-left:25%;margin-right:25%;margin-top:5rem;max-width:50%;width:100%"] {
```
varl, line 8: Oh wow. I didn't know that was even possible. :joy:

varl:
Do it before @cooper-joe sees that I didn't do it before I left ! :cold_sweat:

HendrikThePendric (approved):

Mohammer5 (approved):

docs/src/css/custom.css
amcgee: Google taught me!  (hi @varl 👋 )

PR #999, opened 2022-03-01
feat(pagination): add support for pagers without a total

HendrikThePendric:
This PR adds support to `Pagination` component for pager objects without `total` and `totalPages`. This is relevant because computing the total is expensive and some endpoints have already added support for omitting the totals.

There is some complexity regarding the information we are able still show in the page summary for these pagers without total. 

For example a pager object without a total could look like this:
```
{
    page: 2,
    pageSize: 50,
    isLastPage: false,
}
```
Here we could say `Page 2, items 51-100` and we would know this to be correct, because we know there is a next page, so the current page must have 50 items.

But suppose `isLastPage` would be `false`, i.e.:
```
{
    page: 2,
    pageSize: 50,
    isLastPage: true,
}
```
Now saying `Page 2, items 51-100` is not really correct/accurate. It’s _possible_ that this last page contains exactly 50 items, but not at all likely.

I considered two options to deal with this problem:
1. We could only mention the page number in the page summary, i.e. `Page 2`. This would be the simplest solution but you lose quite a bit of useful info reg. the items numbers, and this info is actually correct for all pages apart from the last one. Because of this I decided not to implement this solution.
2. What I did instead was to add another new (optional) prop to the `Pagination` component, called `pageLength`. When this is provided it will be used to compute the last item and this will be correct. To ensure developers understand they need to populate this prop for pagers without a total, I've made this prop conditionally required when `isLastPage` is `true`. If the prop is not provided, the component will not crash, but simply use the `pageSize` and `page` properties to compute the last item (which could be an incorrect value 🤷‍♂️)

### DIFF
--- a/components/pagination/src/page-controls.js
+++ b/components/pagination/src/page-controls.js
@@ -37,7 +37,7 @@ const PageControls = ({
             secondary
             className="button-next"
             small
-            disabled={isLastPage}
+            disabled={!!isLastPage}
             onClick={() => onClick(page + 1)}
             dataTest={`${dataTest}-page-next`}
         >
--- a/components/pagination/src/page-controls.js
+++ b/components/pagination/src/page-controls.js
@@ -19,7 +19,6 @@ const PageControls = ({
     onClick,
     nextPageText,
     page,
-    pageCount,
     previousPageText,
 }) => (
     <div data-test={`${dataTest}-pagecontrols`}>
@@ -38,7 +37,7 @@ const PageControls = ({
             secondary
             className="button-next"
             small
-            disabled={isLastPage || page === pageCount}
+            disabled={isLastPage}
             onClick={() => onClick(page + 1)}
             dataTest={`${dataTest}-page-next`}
         >
@@ -71,7 +70,6 @@ PageControls.propTypes = {
         .isRequired,
     onClick: PropTypes.func.isRequired,
     isLastPage: PropTypes.bool,
-    pageCount: PropTypes.number,
 }
 
 export { PageControls }
--- a/components/pagination/src/page-controls.js
+++ b/components/pagination/src/page-controls.js
@@ -15,6 +15,7 @@ const translate = (prop, interpolationObject) => {
 
 const PageControls = ({
     dataTest,
+    isLastPage,
     onClick,
     nextPageText,
     page,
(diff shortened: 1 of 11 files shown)


mediremi (approved):

components/pagination/src/page-controls.js
```diff
@@ -37,7 +38,7 @@ const PageControls = ({
             secondary
             className="button-next"
             small
-            disabled={page === pageCount}
+            disabled={isLastPage || page === pageCount}
```
amcgee, line 41: (minor) could this move up to the parent component so we don't need to deal with this branching logic (and additional component) here?  Basically, this component would only have a boolean `isLastPage` prop and the caller would pass `isLastPage={isLastPage || page === pageCount}`

amcgee (changes requested):
Looks like the storybook doesn't correctly hide the total elements for the `no totals` stories.

Approach sounds good and the code looks good from a high level!

amcgee:
@HendrikThePendric 
Screen Shot 2022-03-01 at 16 21 59

amcgee:
>  If the prop is not provided, the component will not crash, but simply use the pageSize and page properties to compute the last item (which could be an incorrect value 🤷‍♂️)

This is probably fine, but maybe we should hide the item numbers if `pageSize` isn't provided...?

HendrikThePendric:
> > If the prop is not provided, the component will not crash, but simply use the pageSize and page properties to compute the last item (which could be an incorrect value 🤷‍♂️)
> 
> This is probably fine, but maybe we should just always hide the item numbers if `pageSize` isn't provided...?

@amcgee I think you mean `pageLength` here... But I think the point you're making is mainly about not showing a value that may be incorrect, which I think is a very valid point. I was initially concerned about always hiding information which will actually be correct almost all of the time. But it turns out it's perfectly doable to identify the cases where the `lastItem` can't reliably be calculated. So I went back and implemented it as follows:
- If there is a total, the pager will work as it always did
- Without a total it will show information about the page nr, and the item-range.
- In a very specific case, where the last item can't be computed reliably, it will only show information about the current page nr. But this is not something that will happen often, only when all of the conditions below are met: 
    - No `total` provided
    - No `pageLength` provided
    - At last page (i.e. `lastPage` is `true`)
    
In the process I ended up refactoring things quite a bit. I think the different moving parts have more clearly separated responsibilities now.

amcgee (approved):
Looks good, nice work @HendrikThePendric !

PR #1002, opened 2022-03-09
fix(SharingDialog): clear search field once access is granted, fix style

edoardo:
#### Once an access is granted to a user/group, the search input should be cleared.
Before:
https://user-images.githubusercontent.com/150978/157421634-2a3b09cb-1cb4-455c-80a4-13001bac6054.mov

After:
https://user-images.githubusercontent.com/150978/157423660-adfe43fd-46ff-4dd4-a786-3092f3827d57.mov

#### Style fix, space elements in the form.
Before:
Screenshot 2022-03-09 at 11 18 02

After:
Screenshot 2022-03-09 at 11 25 17

### DIFF
--- a/components/sharing-dialog/src/access-add/access-add.js
+++ b/components/sharing-dialog/src/access-add/access-add.js
@@ -1,7 +1,7 @@
 import { Button } from '@dhis2-ui/button'
 import { SingleSelectField, SingleSelectOption } from '@dhis2-ui/select'
 import { useOnlineStatus } from '@dhis2/app-runtime'
-import { colors } from '@dhis2/ui-constants'
+import { colors, spacers } from '@dhis2/ui-constants'
 import PropTypes from 'prop-types'
 import React, { useState, useContext } from 'react'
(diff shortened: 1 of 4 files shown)


HendrikThePendric (approved):
LGTM, but perhaps @ismay could have a look too.

ismay:
Looks good. I'll post my response from slack here as well:

> I think the core issue here is that this input is trying to deal with two types of information at the same time. That's why we need these types of hacks. On the one hand it can represent a selection, and on the other hand it can have just a search string. I think eventually a cleaner solution would be to improve the connection between search string and selection.

I didn't manage to find a nice solution for this at the time that still satisfied the autocomplete pattern. But eventually I think that'd be a more robust solution.

ismay (approved):
In lieu of a test demonstrating the fix I'll assume it works. Judging purely from the code it LGTM 👍

HendrikThePendric:
> In lieu of a test demonstrating the fix

Good point @ismay. I added an additional step to one of the existing feature files to demonstrate the fix.

PR #1003, opened 2022-03-09
fix(pagination): hide page select when only 1 page or too many pages

HendrikThePendric:
The main reason for doing this PR is because I remember seeing the page-select completely freezing the browser when there was a huge volume of data. Suppose there are 1,500,000 records and the page size is 50, then the page select would be dynamically populated with 30,000 options and the browser would freeze.  Navigating from page to page is not going to help users find the record they were looking for when there are this many records either. By introducing this max page count limit for showing the page-select, we prevent the browser from freezing...

I see this more of a safety catch type of fix because really if we are dealing with this many records a pager with totals is probably going to cause performance issues on the server as well, so it shouldn't really happen.

Additionally, I realised it's a bit silly to show a page select when there is only one page to select.

And I thought it made more sense to use the `pageCount` prop instead of the `total` prop to compute the visibility because that prop is actually used by the page-select.

(and then there is also a completely unrelated change in this PR: I realised there was a console log in one of the tests)

### DIFF
--- a/components/pagination/src/pagination.js
+++ b/components/pagination/src/pagination.js
@@ -38,6 +38,11 @@ const Pagination = ({
         pageSize,
         total,
     })
+    const showPageSelect =
+        !hidePageSelect &&
+        typeof pageCount === 'number' &&
+        pageCount > 1 &&
+        pageCount <= 2000
 
     return (
         <div className={cx('container', className)} data-test={dataTest}>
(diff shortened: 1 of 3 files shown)


components/pagination/src/pagination.js
```diff
@@ -38,6 +38,11 @@ const Pagination = ({
         pageSize,
         total,
     })
+    const showPageSelect =
+        !hidePageSelect &&
+        typeof pageCount === 'number' &&
+        pageCount > 1 &&
+        pageCount <= 2000
```
Mohammer5, line 45: [Magic number](https://rules.sonarsource.com/typescript/RSPEC-109), I'd say we either put that in a const or a prop (with default value)

Mohammer5 (changes requested):

Mohammer5 (approved):

mediremi (approved):

PR #1004, opened 2022-03-10
fix(sharing-dialog): prevent extra request and render of autocompletion list after selection

mediremi:
Closes https://dhis2.atlassian.net/browse/LIBS-288

Currently, when a user clicks on an autocompletion result, an extra search request is triggered due to the value of the input field being updated. This PR fixes this issue by only fetching search results if the search string is not equal to the current selection.


components/sharing-dialog/src/autocomplete/sharing-autocomplete.test.js
```diff
@@ -0,0 +1,59 @@
...
+        userEvent.click(screen.getByRole('listitem').querySelector('a'))
+        expect(screen.getByRole('textbox')).toHaveValue(userDisplayName)
+
+        try {
+            await waitFor(() => screen.getByRole('listitem'), { timeout: 1 })
+        } catch (error) {
+            if (!error.message.startsWith('Unable to find role="listitem"')) {
+                throw error
+            }
+        }
```
mediremi, line 56: What we're trying to do here is wait for the two `useEffect`s in `SharingAutocomplete` to have been triggered (the 1st `useEffect` triggered is [this one](https://github.com/dhis2/ui/blob/9c4e0c428cf731ff6d18a042d3c397ce9ae63d87/components/sharing-dialog/src/autocomplete/sharing-autocomplete.js#L30), which then triggers [the second](https://github.com/dhis2/ui/blob/9c4e0c428cf731ff6d18a042d3c397ce9ae63d87/components/sharing-dialog/src/autocomplete/sharing-autocomplete.js#L39-L50)). 

An alternative way of waiting for these two `useEffect`s is to do the following:
```js
const sleep = ms => new Promise(resolve => setTimeout(resolve, ms))
await sleep(0)
```

components/sharing-dialog/src/autocomplete/sharing-autocomplete.js
```diff
@@ -37,6 +37,10 @@ export const SharingAutocomplete = ({ selected, onSelection }) => {
     const debouncedRefetch = useCallback(debounce(refetch, 250), [refetch])
 
     useEffect(() => {
+        if (search && search === selected) {
+            return
+        }
+
```
mediremi, line 43: The fix

ismay (approved):

components/sharing-dialog/src/autocomplete/sharing-autocomplete.test.js
ismay: Yeah it's always hard to test for something not being there. So, if what we're trying to do is wait for logic that will have executed after a set amount of renders, we could also use `await waitForNextUpdate()`.

Would that be an option here?

mediremi: Unfortunately `waitForNextUpdate` is only for react hook testing library not react testing library :(

ismay: Ah I figured they'd have something like that for react-testing as well. And how about https://testing-library.com/docs/dom-testing-library/api-async/#waitforelementtoberemoved?

HendrikThePendric (approved):
As discussed during the team meeting:
- This PR should be merged now because the hard freeze is today and this change fixes the bug
- The long term fix is captured in a [JIRA issue](https://dhis2.atlassian.net/browse/LIBS-290)

components/sharing-dialog/src/autocomplete/sharing-autocomplete.test.js
ismay: Ok I see yeah

PR #1007, opened 2022-03-16
fix(sharing-dialog): support eventVisualization and interpretation types TECH-1037

edoardo:
Solves [TECH-1037](https://dhis2.atlassian.net/browse/TECH-1037).

`eventVisualization` and `interpretation` are now passed as types to the `SharingDialog` component.
This PR resolves the warnings about invalid prop type.

Before:
Screenshot 2022-03-14 at 11 33 35


HendrikThePendric (approved):

ismay:
This might be a dumb question, but I assume that this still works as intended here right: https://github.com/dhis2/ui/blob/040198a476a5c71a529951a1be711b70cc4e0a0b/components/sharing-dialog/src/tabs/tabbed-content.js#L18-L86? Dashboard is still the only type that should have an effect on the tabs?

PR #1027, opened 2022-04-12
feat(pagination): disable navigation elements

HendrikThePendric:
Fixes [LIBS-310](https://dhis2.atlassian.net/browse/LIBS-310)

### Update

After discussing things with Ismay, we decided to simplify the implementation a lot. The automatic enabling/disabled of buttons based on an internal navigation state was ultimately based on an assumption, and would not be enough for a lot of use cases (i.e. filtering/sorting a list). So most apps would need to use the new `disabled` prop anyway. Adding the `disabled` prop alone is enough. So in my last commit I removed the internal state from the component.

So now this PR does the following:
- Add a `disabled` prop to `Pagination`
- Rewrite tests to react-testing-library
- Add some general config for react-testing library

What follows below can be regarded as historical information. 

---

### Description

The `Pagination` component is meant to display information from the pager object of the currently visible data. While requests for a new page are pending, the navigation elements should be disabled. This will prevent multiple request for the same page from firing if the user clicks multiple times. Also, it wil make it clearer to the user that clicking next a bunch of times won't actually forward the page by the amount of clicks.

There is some granularity in what needs to be disabled when:
- Previous and next buttons are always disabled while navigating, because a "next" page is relative to a "current" page, and the current page won't get updated until the data comes in.
- The page-select only becomes disabled when the page size is updating. It's OK to request a particular page while a request is pending, because the page-select uses absolute page numbers instead of relative ones.
- The page summary text always gets grayed out when a request is pending. This serves as a visual cue to let the user know that a) something is happening, and b) the page summary information is stale

And since this component does not tap into the actual HTPP request status, it fully relies on the following assumptions, which I think are 100% correct:
- When a user interacts with one of the pager controls, a calllback is fired. 
- This callback is responsible for delivering the next page number (or page size) to the app
- The app will then do whatever it needs to do to get a new page and will update the props of the Pagination component accordingly

So based on the above, the internal state for the `Pagination` component is updated as follows:
- When the user selects a new page / page-size the internal `isNavigatingToPage` / `isChangingPageSize` is set to `true`
- Once the corresponding props are updated by the app, they will be set to `false` again

Having some internal state that automatically disables relevant controls is nice, but there are other cases where the app simply needs to disable these. A relevant use-case here is having a filter or sortable columns. The `Pagination` component has no way of knowing what is going on with these requests, so the pagination will need to be disabled via the app, which has been made possible in this PR by introducing a new `disabled` prop.

### Alternative approach

I've been going back-and-forth a bit with two possible strategies for this. The one I've settled on above is fairly blocking but also predictable. 

An alternative implementation could replace internal boolean state variables `isNavigatingToPage` and `isChangingPageSize` with analogue state variable of type number, i.e. `nextPage` and `nextPageSize`. We could then base the values we pass to the `onPageChange` and `onPageSizeChange` on those and say a the stuff has done loading once `nextPage === page && nextPageSize === pageSize`.

This approach would be a lot less blocking, especially when you have a pager object with a total and totalPages. But I'm worried there will be a lot of edge cases to consider. Don't really have anything super specific in mind, but just foresee it will be hard to work out when the internal state would actually need to get updated when the `page` and `pageSize` props change value....

But let me know if the alternative approach sounds much more promising than what I have now.

### Bonus features

Some of the page transition tests were hard to write in Jest+Enzyme and I found working with Jest + `react-testing-library` a lot more straightforward. I don't mind having both enzyme and react-testing-library test within one project, but within one file is a bit too much. So I decided to port `pagination.test.js`  to react-testing-library. In the process I also tweaked the Jest config a bit:
- There is now a separate config file for react-testing-library which is added to the Jest config's `setupFilesAfterEnv` array
- This config file ensures the `..ByTestId` queries now work with our `data-test` data-attribute 
- I have also added `@testing-library/jest-dom/extend-expect` to the `setupFilesAfterEnv` array, which enables us to write nice DOM assertions such as `expect(...).toBeDisabled()`


mediremi (approved):

components/pagination/src/__tests__/get-item-range.test.js
```diff
@@ -60,6 +60,6 @@ describe('getItemRange', () => {
             isLastPage: true,
         })
 
-        expect(lastItem).toEqual(NaN)
+        expect(lastItem).toBe(NaN)
```
ismay, line 63: Could also use https://jestjs.io/docs/expect#tobenan

ismay (approved):
Nice, looks good. The tests make it very easy to understand what the changes are meant to do 👍. Also thanks for description, gives a lot of context.

I find our jest.config.js vs the jest.config.shared.js setup a little confusing btw. We might have to take a look at that later on. Figure out why we have two 🤷‍♂️

PR #1052, opened 2022-04-26
fix: increase prefix spacing so icon is closer to the option text

ismay:
Before this fix:

before

After the fix:

after


jenniferarnesen:
Some of examples of what these changes would look like:

image

image

image

jenniferarnesen (approved):

Mohammer5 (approved):

PR #1055, opened 2022-05-09
fix: correct api reference, show changelog

amcgee:
This fixes the broken api reference generation which required node >= 15 to use `replaceAll` (I updated it to use `replace` with a global regex instead, so older node versions work)

This also moves `Changelog` to a top-level sidebar entry and imports the semantic-release generated CHANGELOG (escaping some html that appeared in changelog entries)

I also committed the generated API docs, we can remove this in the future if we want @HendrikThePendric


KaiVandivier (approved):
Looks good to go once linting is fixed :)

I don't see where the root level `API.md` file is shown on the docs site, is it in use? Same for `icons/API.md`

KaiVandivier:
Ah gotcha. Some documentation on how to use the icons is important as that's a common question we get, but that can happen in a different PR

scripts/generate-api-docs.js
```diff
@@ -63,7 +63,7 @@ const format_type = (type) => {
                 rst[key] = format_type(shape[key])
             })
 
-            return JSON.stringify(rst, null, 2).replaceAll('\n', '<br/>')
+            return JSON.stringify(rst, null, 2).replace(/\n/g, '<br/>')
```
HendrikThePendric, line 66: I looked up why you might have switched from `.replaceAll` to `.replace`, which led me [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#non-global_regex_throws). It seems odd that you always have to pass a global flag wo whatever regex you pass to `.replaceAll`, since a non-global replace-all action seems to be an impossible combination to me.

HendrikThePendric (approved):
I'm not 100% sure why we are seeing so many new `API.md` files as I would have expected all of these to have been there already after https://github.com/dhis2/ui/pull/1042. Having said that, everything in here does seem correct, so I'm approving.

PR #1056, opened 2022-05-10
fix: prevent SharingDialog from displaying id as modal name (LIBS-322)

martinkrulltott:
Fixes [LIBS-322](https://dhis2.atlassian.net/browse/LIBS-322)

`SharingDialog`: Prevents `id` from being used as the `name` of the modal

Below is an example from the provided story, where `name` and `displayName` resemble the `id`, which results in the title just being "Sharing and access":

```
object: {
    displayName: 'sharing-test',
    externalAccess: false,
    id: 'sharing-test',
    name: 'sharing-test',
    publicAccess: '--------',
    userAccesses: [],
    userGroupAccesses: []
}
```

_before_
image

_after_
image

### DIFF
--- a/components/sharing-dialog/src/sharing-dialog.js
+++ b/components/sharing-dialog/src/sharing-dialog.js
@@ -159,7 +159,13 @@ export const SharingDialog = ({
 
     return (
         <FetchingContext.Provider value={mutating || fetching}>
-            <Modal onClose={onClose} name={object.displayName || object.name}>
+            <Modal
+                onClose={onClose}
+                name={
+                    (object.displayName !== object.id && object.displayName) ||
+                    (object.name !== object.id && object.name)
+                }
+            >
                 <TabbedContent
                     id={id}
                     users={users}
--- a/components/sharing-dialog/src/sharing-dialog.stories.js
+++ b/components/sharing-dialog/src/sharing-dialog.stories.js
@@ -108,6 +108,18 @@ const customData = {
     },
 }
 
+const customDataIdAsName = {
+    ...customDefaultData,
+    sharing: {
+        ...customDefaultData.sharing,
+        object: {
+            ...customDefaultData.sharing.object,
+            name: customDefaultData.sharing.object.id,
+            displayName: customDefaultData.sharing.object.id,
+        },
+    },
+}
+
 const customDataDisabledAccess = {
     ...customData,
     sharing: {
@@ -185,6 +197,13 @@ export const WithName = (args) => (
     </CustomDataProvider>
 )
 WithName.storyName = 'With name'
+
+export const WithIdAsName = (args) => (
+    <CustomDataProvider data={customDataIdAsName}>
+        <SharingDialog {...args} />
+    </CustomDataProvider>
+)
+WithIdAsName.storyName = 'With id as name'
 
 export const WithDisabledAccess = (args) => (
     <CustomDataProvider data={customDataDisabledAccess}>


components/sharing-dialog/src/sharing-dialog.js
```diff
@@ -159,7 +159,13 @@ export const SharingDialog = ({
 
     return (
         <FetchingContext.Provider value={mutating || fetching}>
-            <Modal onClose={onClose} name={object.displayName || object.name}>
+            <Modal
+                onClose={onClose}
+                name={
+                    (object.displayName !== object.id && object.displayName) ||
```
edoardo, line 165: I think this can be simplified but just using `displayName`.
`displayName` should always have a value, either the translated version or the original version as default.
But we should confirm first.

edoardo (approved):

components/sharing-dialog/src/sharing-dialog.js
martinkrulltott: From the perspective of the fix, both `displayName` and `name` use the `id`, as seen e.g. here https://debug.dhis2.org/dev/api/39/sharing?type=interpretation&id=HOFzdanjMHE. But I see your point that there might not be a case where `displayName` is empty but `name` is set, that's what I've seen in my testing, however I can't confirm that to always be true.

amcgee (changes requested):
I'm not sure I understand this change... what if someone WANTS to use the id as the name of an object?

amcgee:
So the issue we're trying to address here is that the server returns `name` and `displayName` as `id` for the interpretations endpoint, correct?

amcgee:
It's counter to the goal of contextualized sharing (which was part of the design spec @cooper-joe created) to hide the id here, but in the case where there isn't a reasonable name I think it's ok to do so

HendrikThePendric (changes requested):
Since entities like comments don't have a name I guess the only possible value to use for `displayName` is the ID. I get that usually we don't want to show that to users. In fact, I can't really think of any scenario where showing the UID would add value to the UX.

I would personally prefer to try to solve this issue without adding a new prop to the component. We can always add a `showName` / `hideName` prop later on, if it turns out there are cases where developers need to show the ID.

Also, I agree with what @edoardo wrote about the `displayName` field. We should rely on that exclusively.

So I would recommend this for now:
```jsx
name={object.displayName === object.id ? '' : object.displayName}
```

If it turns out later on we do need to sometimes show the UID and decide to add an additional `showName` prop, then we would have something like this:
```js
name={showName ? object.displayName : ''}
```

janhenrikoverland:
I would prefer backend to stop returning ids in the `displayName` field. Technically it's a breaking change, but if clients are intentionally using the id they are rather doing `object.id`.

Regarding a new prop, I prefer booleans to be false by default, so `hideName` makes more sense to me. But I don't think it's needed if backend makes the above change.

HendrikThePendric:
> I would prefer backend to stop returning ids in the `displayName` field. Technically it's a breaking change, but if clients are intentionally using the id they are rather doing `object.id`.

I think that'd make sense. 

In the context of this PR, it definitely wouldn't cause any issues if this breaking change was implemented in the backend. If `displayName` is empty we would see the correct behaviour with or without this PR.

> Regarding a new prop, I prefer booleans to be false by default, so `hideName` makes more sense to me. 

Agreed

> But I don't think a new prop is needed if backend makes the above change.

Agreed

PR #1060, opened 2022-05-18
fix(layer): prevent events from children bubling up by improved markup

HendrikThePendric:
[LIBS-192](https://dhis2.atlassian.net/browse/LIBS-192) describes something that is actually natural browser behaviour, so not a bug per se. However, it can be confusing, and other component libraries (MUI & Chakra UI) don't have this issue.

By making the backdrop a sibling instead of a parent of the component's children, click events won't buble up from things like modal to the layer anymore. So this is a much more natural way of solving things than we had before.


ismay (commented):
To clarify what was happening here:

> If the button is pressed on one element and the pointer is moved outside the element before the button is released, the event is fired on the most specific ancestor element that contained both elements.
https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event

This means that if you click a modal and drag to the backdrop and release there, the click will fire on the most specific common ancestor, which was the layer.

In thinking about it, I think seeing the original behaviour as a bug comes from understanding the layer's onclick as onBackgroundClick, and/or not being of browser handling of dragging whilst clicking. onBackgroundClick is what we've now technically implemented here. However, the new behaviour will be confusing to someone who understands the onclick as an onclick that's on the layer itself.

I would be in favour of introducing onBackgroundClick as a feat, and if we want to, deprecating and eventually removing onClick. I think that most clearly communicates our intent with this api.

HendrikThePendric:
I think what you are saying makes sense: we added an `onClick` to the `Layer`, but always intended it to be an `onBackdropClick` / `onBackgroundClick`. 

However, it's not like the `onClick` we had prior to this PR allowed clicks to bubble up naturally. It has been blocking clicks originating from its children all along. We merely found an edge case where this wasn't working (i.e. when the mousedown was on the children but the mouseup on the layer).

And the above makes it a bit problematic to have both an `onClick` and `onBackdropClick` around, because there is no clear difference between the two based on the behaviour they would provide. Implementing things so that both props are there and the distinction is as expected, would mean changing the `onClick` prop so that it would be triggered by clicks originating in its children. And that is a breaking change which would affect lots of apps and components. I don't think we should do this (and also don't think this is what you were suggesting).

I guess we _could_ leave `onClick` as it is, introduce `onBackdropClick` as a new prop which would be implemented as I've done in this PR. And then at some point, we remove the `onClick`. But by doing so, aren't we just substituting one form of unclarity (the `onClick` prop isn't named accurately) with another (we have two props doing almost the same thing)?

HendrikThePendric:
How about this solution, which is sort of a little spin on what you said?
- Implement the PR as is
- But in this PR also introduce an alias `onBackdropClick` (i.e. new prop, exact same behaviour)
- Deprecate `onClick` next major release
- Remove it the release after

ismay:
In that case I'd deprecate onClick now, remove in the next major version. But yeah, that sounds good to me.

ismay:
So I think if we want to deprecate onClick we can do that with a note on a feat commit, where the feat is adding onBackgroundClick.

Mohammer5:
I second the idea of deprecating the current prop at the same time when introducing a better prop. Otherwise we'd have two props with the same behavior, but none of them are marked as deprecated, so you wouldn't know which one to use

Mohammer5 (approved):
Changes look good.

In the storybook I can still see `onClick` when I click on the `jsx` tab when looking at a layer story. Also the API.md still contains the `onClick` reference. I suppose the API.md will be autogenerated, but what about the story? In the story code it looks like you've already used `onBackdropClick`, no idea why the storybook (e.g. [here](https://ui.dhis2.nu/demo/?path=/story/layer--with-click-handler)) still shows `onClick`

ismay (approved):
Looks good. When merging I'd squash this and add a comment so that the deprecation shows up in the changelog as well.

PR #1069, opened 2022-05-24
feat(data-table-cell): add background-color prop

HendrikThePendric:
Closes [LIBS-324](https://dhis2.atlassian.net/browse/LIBS-324)

### DIFF
--- a/components/table/src/data-table/data-table-cell.js
+++ b/components/table/src/data-table/data-table-cell.js
@@ -8,6 +8,7 @@ export const DataTableCell = forwardRef(
         {
             active,
             align,
+            backgroundColor,
             bordered,
             children,
             className,
@@ -37,6 +38,7 @@ export const DataTableCell = forwardRef(
             <TableCell
                 active={active}
                 align={align}
+                backgroundColor={backgroundColor}
                 bordered={bordered}
                 className={className}
                 colSpan={colSpan}
@@ -79,9 +81,10 @@ const requiredIfFixedPropType = requiredIf(
     PropTypes.string
 )
 DataTableCell.propTypes = {
-    /** To toggle background color, for example for editing */
+    /** To toggle border color, for example for editing */
     active: PropTypes.bool,
     align: PropTypes.oneOf(['left', 'center', 'right']),
+    backgroundColor: PropTypes.string,
     bordered: PropTypes.bool,
     children: PropTypes.node,
     className: PropTypes.string,
--- a/components/table/src/data-table/table-elements/table-data-cell/table-data-cell.js
+++ b/components/table/src/data-table/table-elements/table-data-cell/table-data-cell.js
@@ -9,6 +9,7 @@ export const TableDataCell = forwardRef(
         {
             active,
             align,
+            backgroundColor,
             bordered,
             children,
             className,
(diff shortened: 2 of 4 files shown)


Mohammer5 (changes requested):
I'd prefer the custom styles to be set via the `style` prop. Seems cleaner to me (less css magic and no need to think about specificity)

HendrikThePendric:
> I'd prefer the custom styles to be set via the `style` prop. Seems cleaner to me (less css magic and no need to think about specificity)

Good suggestion, I've implemented things like that just now and indeed it is much easier to mentally digest this way. I have kept the part where I add `staticStyle` classname to prevent row hover styles from interfering.

Mohammer5 (approved):

components/table/src/data-table/table-elements/table-data-cell/table-data-cell.js
```diff
@@ -79,6 +81,7 @@ TableDataCell.propTypes = {
     /** To toggle background color, for example for editing */
```
KaiVandivier, line 81:
```suggestion
    /** To toggle border color, for example for editing */
```

amcgee:
This should probably have some documentation right?

KaiVandivier:
Talked with Hendrik in DMs about this, but copying summary here for posterity:

I suggest setting the background color with the `!important` flag in `styled-jsx` instead of inline styles, because I think it will be better for the library to stick with one way of setting styles and avoid spreading them out between inline and `styled-jsx` 🙂

components/table/src/data-table/data-table-cell.js
```diff
@@ -79,9 +81,10 @@ const requiredIfFixedPropType = requiredIf(
     PropTypes.string
 )
 DataTableCell.propTypes = {
-    /** To toggle background color, for example for editing */
+    /** To toggle border color, for example for editing */
     active: PropTypes.bool,
     align: PropTypes.oneOf(['left', 'center', 'right']),
+    backgroundColor: PropTypes.string,
```
KaiVandivier, line 87:
```suggestion
    /** Sets background color of the cell. Disables dynamic background colors from active, hover, and selected states */
    backgroundColor: PropTypes.string,
```

components/table/src/data-table/table-elements/table-data-cell/table-data-cell.js
```diff
@@ -76,9 +91,10 @@ const stylePropType = mutuallyExclusive(
 )
 
 TableDataCell.propTypes = {
-    /** To toggle background color, for example for editing */
+    /** To toggle border color, for example for editing */
     active: PropTypes.bool,
     align: PropTypes.oneOf(['left', 'center', 'right']),
+    backgroundColor: PropTypes.string,
```
KaiVandivier, line 97:
```suggestion
    /** Sets background color of the cell. Disables dynamic background colors from active, hover, and selected states */
    backgroundColor: PropTypes.string,
```

components/table/src/data-table/table-elements/table-header-cell/table-header-cell.js
```diff
@@ -84,9 +97,10 @@ const stylePropType = mutuallyExclusive(
 )
 
 TableHeaderCell.propTypes = {
-    /** To toggle background color, for example for editing */
+    /** To toggle border color, for example for editing */
     active: PropTypes.bool,
     align: PropTypes.oneOf(['left', 'center', 'right']),
+    backgroundColor: PropTypes.string,
```
KaiVandivier, line 103:
```suggestion
    /** Sets background color of the cell. Disables dynamic background colors from active, hover, and selected states */
    backgroundColor: PropTypes.string,
```

KaiVandivier (approved):
Looks good! Added a prop description suggestion :)

PR #1070, opened 2022-05-25
feat: add "onKeyDown" prop to various input components

Mohammer5:
Relates to [TECH-856](https://dhis2.atlassian.net/browse/TECH-856)


components/file-input/src/file-input-field-with-list/__tests__/file-input-field-with-list.test.js
```diff
@@ -0,0 +1,32 @@
+import { configure, render, fireEvent, screen } from '@testing-library/react'
+import React from 'react'
+import { FileInputFieldWithList } from '../file-input-field-with-list.js'
+
+configure({ testIdAttribute: 'data-test' })
```
HendrikThePendric, line 5: I believe this isn't needed? We have configured that already in `./jest.testing-library.config.js`.

HendrikThePendric (changes requested):
I've left one code comment, but it actually applies to most test files. Please remove these lines before merging.

Apart from that, it all looks good to me. Most important thing is that these new event handlers implement the same callback signature as the existing event handlers.

HendrikThePendric (approved):
Thanks for removing those calls to `configure`

components/select/src/single-select-field/single-select-field.js
```diff
@@ -168,6 +170,7 @@ SingleSelectField.propTypes = {
     onChange: PropTypes.func,
     /** Called with signature `({ selected: string }, event)` */
     onFocus: PropTypes.func,
+    onKeyDown: PropTypes.func,
```
KaiVandivier, line 173:
```suggestion
    /** Called with signature `({ selected: string }, event)` */
    onKeyDown: PropTypes.func,
```

components/select/src/multi-select-field/multi-select-field.js
```diff
@@ -169,6 +171,7 @@ MultiSelectField.propTypes = {
     onChange: PropTypes.func,
     /** Called with signature `({ selected: [String] }, event)` */
     onFocus: PropTypes.func,
+    onKeyDown: PropTypes.func,
```
KaiVandivier, line 174:
```suggestion
    /** Called with signature `({ selected: [String] }, event)` */
    onKeyDown: PropTypes.func,
```

components/button/src/button/button.js
```diff
@@ -151,4 +155,6 @@ Button.propTypes = {
     onClick: PropTypes.func,
     /** Callback to trigger on focus. Called with same args as `onClick` */
     onFocus: PropTypes.func,
+    /** Callback to trigger on focus. Called with same args as `onClick` */
```
KaiVandivier, line 158:
```suggestion
    /** Callback to trigger on key-down. Called with same args as `onClick` */
```

KaiVandivier (approved):
Just a few prop description edits :) everything else looks good though so will approve!

PR #1112, opened 2022-06-22
fix(chip): accept strings and numbers for margin props

HendrikThePendric:
The chip component now accepts strings and numbers for its margin props. We chose this approach because we typically accept strings for CSS attribute props, but this component only accepted numbers. Letting the component _only_ accept strings only would have represented a breaking change, so to prevent this we now accept both strings and numbers.

### DIFF
--- a/components/chip/src/remove.js
+++ b/components/chip/src/remove.js
@@ -1,7 +1,7 @@
 import { colors } from '@dhis2/ui-constants'
 import PropTypes from 'prop-types'
 import React from 'react'
-import { css, resolve } from 'styled-jsx/css'
+import css from 'styled-jsx/css'
 
 function CancelOutline({ className }) {
     return (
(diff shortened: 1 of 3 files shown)


components/chip/src/remove.js
```diff
@@ -1,7 +1,7 @@
 import { colors } from '@dhis2/ui-constants'
 import PropTypes from 'prop-types'
 import React from 'react'
-import { css, resolve } from 'styled-jsx/css'
+import css from 'styled-jsx/css'
```
HendrikThePendric, line 4: I changed this import because Jest didn't like the named imports.

Mohammer5 (approved):

jenniferarnesen (approved):

cooper-joe (approved):

PR #1118, opened 2022-07-07
fix(validators): ensure integer validator rejects any decimal separator

kabaros:
Implements [TECH-1230](https://dhis2.atlassian.net/browse/TECH-1230)

---

### Description

with the current implementation, a value such as `4.0` would be accepted by the integer validator but the API rejects it. This brings the client-side validator inline with the API implementation.

---

### Checklist

-   [x] Tests were added

### DIFF
--- a/collections/forms/src/validators/__tests__/integer.test.js
+++ b/collections/forms/src/validators/__tests__/integer.test.js
@@ -21,9 +21,10 @@ describe('validator: integer', () => {
         ])
     })
 
-    describe('rejects other data types', () => {
+    describe('rejects other data types and leading zeros', () => {
         testValidatorValues(integer, invalidIntegerMessage, [
             'text',
+            '014',
             true,
             {},
             [],
@@ -36,6 +37,9 @@ describe('validator: integer', () => {
             0.23456,
             5.987,
             1e-12,
+            '4.0',
+            '4.000',
+            '4,0',
             '0.23456',
             '5.987',
             '1e-12',
--- a/collections/forms/src/validators/integer.js
+++ b/collections/forms/src/validators/integer.js
@@ -1,12 +1,19 @@
 import i18n from '../locales/index.js'
-import { isEmpty, isInteger, isNumeric, toNumber } from './helpers/index.js'
+import { isEmpty, isNumeric, toNumber } from './helpers/index.js'
 
 const invalidIntegerMessage = i18n.t(
     'Please provide a round number without decimals'
 )
 
+// Regex accepts only digits (no decimals even if it is trailing like 4.0)
+// it also rejects a leading 0 (i.e 04) as this is rejected by backend
+const INTEGER_PATTERN = /^(-?[1-9]\d*|0)$/
+
 const integer = (value) =>
-    isEmpty(value) || (isNumeric(value) && isInteger(toNumber(value)))
+    isEmpty(value) ||
+    (INTEGER_PATTERN.test(value) &&
+        isNumeric(value) &&
+        Number.isSafeInteger(toNumber(value)))
         ? undefined
         : invalidIntegerMessage
 
--- a/collections/forms/src/validators/helpers/index.js
+++ b/collections/forms/src/validators/helpers/index.js
@@ -3,8 +3,6 @@ export const isEmpty = (value) =>
 
 export const isString = (value) => typeof value === 'string'
 
-export const isInteger = (value) => Number.isSafeInteger(value)
-
 export const isNumber = (value) => typeof value === 'number'
 
 export const isNumeric = (value) =>


collections/forms/src/validators/integer.js
```diff
@@ -5,8 +5,13 @@ const invalidIntegerMessage = i18n.t(
...
 )
 
+const DECIMAL_SEPARATOR_PATTERN = /(\.|,)/
+
 const integer = (value) =>
-    isEmpty(value) || (isNumeric(value) && isInteger(toNumber(value)))
+    isEmpty(value) ||
+    (!DECIMAL_SEPARATOR_PATTERN.test(value) &&
+        isNumeric(value) &&
+        isInteger(toNumber(value)))
```
Mohammer5, line 14: I think instead of checking for decimal deparators, we can adjust the `isInteger` helper:

```js
const INTEGER_PATTERN = /^[1-9][0-9]*$/
export const isInteger = (value) => (
    Number.isSafeInteger(value) &&
    INTEGER_PATTERN.test(value.toString())
)
```

Mohammer5 (commented):
Not sure if what I suggested is also what others think, so I'd like to get a second opinion on that first

collections/forms/src/validators/integer.js
kabaros: Putting the logic in the `isInteger` helper doesn't quite work unless we refactor even more because in the validator, we're checking whether `isInteger(toNumber(value))` is true so the `toNumber(value)` bit would convert something like `4.0` to `4` first, making it be accepted.

As for the RegExp itself, I agree it's easier to rationalise positive booleans (has numbers) than negative booleans (does not have decimal)  .. but it's just that the regular expression in this case has to be more complicated than the one you have, we have to cater for negative integers (so allow `-` at the start) and also allow `0`. It felt it's easier to express what we don't want (the decimal separator) rather than what we want in this case.

Mohammer5: > Putting the logic in the isInteger helper doesn't quite work unless we refactor even more because in the validator, we're checking whether isInteger(toNumber(value)) is true so the toNumber(value) bit would convert something like 4.0 to 4 first, making it be accepted.

You're absolutely correct! I didn't see that.

> It felt it's easier to express what we don't want (the decimal separator) rather than what we want in this case.

I suppose your current implementation is good enough. We can change it to `/^-?[0-9]+$/` if there's ever any issue with it

Mohammer5 (approved):

Mohammer5 (approved):

collections/forms/src/validators/integer.js
```diff
@@ -5,8 +5,14 @@ const invalidIntegerMessage = i18n.t(
...
 
+// Regex accepts 0 but not a traling zero (i.e 04) as this is rejected by backend
+const INTEGER_PATTERN = /^(-?[1-9]\d*|0)$/
+
 const integer = (value) =>
-    isEmpty(value) || (isNumeric(value) && isInteger(toNumber(value)))
+    isEmpty(value) ||
+    (INTEGER_PATTERN.test(value) &&
+        isNumeric(value) &&
+        isInteger(toNumber(value)))
```
HendrikThePendric, line 15: This is only indirectly related to the fix you are making, but this PR will introduce some additional confusion so it's probably best if we fix it here...
- `isInteger` is a helper and simply does this `(value) => Number.isSafeInteger(value)`
- This helper is not being used anywhere else in our codebase, it's only used in the `integer` validator above.

So technically we now have two functions that define how we "define and integer" using a slightly different definition. I think we should:
- Remove the `isInteger` helper
- Change the `integer` function to something like this:
```js
const integer = (value) =>
    isEmpty(value) ||
    (isNumeric(value) &&
        INTEGER_PATTERN.test(value) &&
        Number.isSafeInteger(toNumber(value)))
        ? undefined
        : invalidIntegerMessage
```

collections/forms/src/validators/__tests__/integer.test.js
```diff
@@ -36,6 +37,8 @@ describe('validator: integer', () => {
             0.23456,
             5.987,
             1e-12,
+            '4.0',
+            '4,0',
```
HendrikThePendric, line 41: Does the regex also catch multiple trailing zeros? We should make sure values like `'1.000000'` are also rejected, and add a test for that.

HendrikThePendric (changes requested):
Thanks @kabaros . Please see my comments.

HendrikThePendric:
Sorry, when I reviewed I dove right into the code. I hadn't read the conversation you already had with @Mohammer5. Still, I think what I suggested probably makes sense.

collections/forms/src/validators/integer.js
```diff
@@ -5,8 +5,14 @@ const invalidIntegerMessage = i18n.t(
     'Please provide a round number without decimals'
 )
 
+// Regex accepts 0 but not a traling zero (i.e 04) as this is rejected by backend
```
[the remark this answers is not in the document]
Birkbjo: Or `leading zero`?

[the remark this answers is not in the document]
kabaros: yeah I meant leading .. updated now

collections/forms/src/validators/__tests__/integer.test.js
kabaros: yes it does .. added a test

collections/forms/src/validators/integer.js
kabaros: updated

HendrikThePendric (approved):
Awesome, thanks!

PR #1132, opened 2022-08-17
fix(tab): update hover style

cooper-joe:
Partly implements [UX-106](https://dhis2.atlassian.net/browse/UX-106)

---

### Description

This PR adjusts the tab hover style. The previous style had an issue of looking possibly "greyed out" or disabled. The updated hover style avoids using a grey background. The active style is also updated to avoid a flash of grey background, which is unecessarily distracting. The selected text color is darkened a shade, for contrast (this could've been it's own PR, actually, but perhaps it can be bundled here).
`transition`s are also removed for hover to reduce unnecessary motion, using them only for changing selected tabs.

Note that the hover style is a supporting signifier to indicate interaction. The mouse cursor change is the main signifier.

---

### Checklist

-   [X] Tests were added

---

### Screenshots

Before:
SCR-20220817-fsc

After:
SCR-20220817-frj

### DIFF
--- a/components/tab/src/tab/tab.js
+++ b/components/tab/src/tab/tab.js
@@ -44,8 +44,6 @@ const Tab = ({
                 font-size: 14px;
                 line-height: 20px;
 
-                transition: all 150ms ease-in-out;
-                transition-property: color, background-color;
                 cursor: pointer;
             }
 
@@ -61,7 +59,6 @@ const Tab = ({
                 left: 0;
                 height: 4px;
                 width: 100%;
-                transition: background-color 150ms ease-in-out;
                 background-color: transparent;
             }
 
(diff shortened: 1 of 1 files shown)


KaiVandivier (approved):
Looks good :)

components/tab/src/tab/tab.js
```diff
@@ -89,20 +86,30 @@ const Tab = ({
...
-            button:active {
-                /* Briefly highlight clicked tab */
-                background-color: ${colors.grey200};
+            button:active::after {
+                background-color: ${colors.grey800};
             }
 
             button.selected {
-                color: ${theme.primary700};
+                color: ${theme.primary800};
```
tomzemp, line 102: I'm not sure I understand making the text `theme.primary800`, but the bar underneath `theme.primary700`? If that is a design choice, good by me.

tomzemp (approved):
lgtm

Birkbjo (approved):

components/tab/src/tab/tab.js
cooper-joe: Yup, that's intentional. Text looks slightly lighter than blocks of color. Having the text a shade darker keeps it from looking too light next to the color block underneath.

PR #1153, opened 2022-09-22
feat: show version debug info and update notification in profile menu [LIBS-176] [LIBS-344]

amcgee:
Implements [LIBS-176](https://dhis2.atlassian.net/browse/LIBS-176)
Implements [LIBS-344](https://dhis2.atlassian.net/browse/LIBS-344)

Requires https://github.com/dhis2/app-runtime/pull/1252
Blocks https://github.com/dhis2/app-platform/pull/748

Builds on #795 (credit to @Mohammer5) and the previous update notification flow in app-platform (credit to @KaiVandivier)

---

### Description

This implements two features
1. Debug info (DHIS2 and application versions) in headerbar
2. Update Available notification

It also moves to using `useConfig` for the app name (rather than a prop) but maintains backwards compatibility.

Builds on and supersedes the good work by @Mohammer5 in #795

---

### Known issues

-   [x] Haven't yet fixed Cypress tests, they should be much simpler than in #795 because we don't load the systemInfo data dynamically.
-   [x] Styling might need some improvements
-   [x] We eventually would like to either (1) show a tooltip indicating that a click on debug MenuItem will copy to clipboard or (2) show a dialog with extended debug info and an explicit copy action

---

### Checklist

-   ~[ ] API docs are generated~ (no API docs for HeaderBar, currently)
-   [x] Tests were added

---

### Screenshots

Screen Shot 2022-09-26 at 13 17 32

https://user-images.githubusercontent.com/947888/192796030-948971a6-cdda-4be6-b5cc-447a697e3b09.mov
(note that some small details have changed since this video was taken)


collections/forms/i18n/en.pot
```diff
@@ -99,3 +96,11 @@ msgstr "Please provide a string"
...
 msgid "Please provide a valid url"
 msgstr "Please provide a valid url"
+
+msgctxt " @ or #"
+msgid ""
+"Please provide a username between 4 and 255 characters long and possibly "
+"separated by . "
+msgstr ""
+"Please provide a username between 4 and 255 characters long and possibly "
+"separated by . _ @ or #"
```
KaiVandivier, line 106: Looks like there's something funny going on with this string

amcgee: I don't know why this is in here, will remove from this PR

amcgee: Probably needs to be extracted somewhere else - I think the characters here are problematic

amcgee: Reverted for now

amcgee:
I'm going to go ahead and merge this to `beta` for testing in-situ

components/header-bar/src/profile-menu/instance-and-app-info.js
```diff
@@ -0,0 +1,78 @@
...
+import React from 'react'
+
+const useDebugInfo = () => {
+    const { appName, appVersion, systemInfo } = useConfig()
+
+    return {
+        dhis2_version: systemInfo?.version || 'unknown',
+        dhis2_revision: systemInfo?.revision || 'unknown',
+        app_name: appName || 'App',
+        app_version: appVersion?.full || 'unknown',
```
Mohammer5, line 14: Do we need to wrap the "unknown"s with `i18n.t(...)`?

components/header-bar/src/profile-menu/update-notification.js
```diff
@@ -0,0 +1,61 @@
...
+import { colors } from '@dhis2/ui-constants'
+import PropTypes from 'prop-types'
+import React from 'react'
+import { useHeaderBarContext } from '../header-bar-context.js'
+import i18n from '../locales/index.js'
+
+export function UpdateNotification({ hideProfileMenu }) {
+    const { appName } = useConfig()
+    const { updateAvailable, onApplyAvailableUpdate } = useHeaderBarContext()
+    console.log(updateAvailable)
```
Mohammer5, line 12: Log statement

components/header-bar/src/profile-menu/instance-and-app-info.js
amcgee: I don't think so, as this is debug information that's intended for administrators / our Jira

amcgee: `unknown` should also only ever appear if the headerbar component is used in an older app-platform version.  We could perhaps use an empty string instead, but I feel like that would be confusing...

amcgee: Or we could hide the menu item in that case.

amcgee: cc @cooper-joe

cooper-joe: Could we use `Couldn't get DHIS2 version` and `Couldn't get app version` instead of `unknown`?

Mohammer5: > Could we use Couldn't get DHIS2 version and Couldn't get app version instead of unknown? Though it's an edge case, would be good to let the user know what's happening.

I think it wouldn't hurt us to translate these. For us, it's just wrapping the string with `i18n.t('....')`

amcgee: Maybe put this in the menu item (and translate it):

```
Unknown DHIS2 version
Unknown app version
```
?

amcgee: I'd advocate for not translating the details in the modal, or anything that's copied to clipboard

amcgee: We could maybe translate the keys in the table... but as we add more and more information there it could get complicated

cooper-joe: > Maybe put this in the menu item (and translate it):
> 
> ```
> Unknown DHIS2 version
> Unknown app version
> ```
> 
> ?

Sounds good to me 👍

amcgee:
Updated for all the various edge cases - a bit silly when we don't expect these to happen, but now they're covered (and tested in Cypress)

Normal: 
Screen Shot 2022-10-03 at 14 06 12

App version unknown: 
Screen Shot 2022-10-03 at 13 57 34

Instance version unknown:
Screen Shot 2022-10-03 at 13 57 23

App name unknown:
Screen Shot 2022-10-03 at 14 03 49

App name and version unknown:
Screen Shot 2022-10-03 at 13 57 29

Update available:
Screen Shot 2022-10-03 at 13 57 54

Update available with app name unknown:
Screen Shot 2022-10-03 at 13 57 44

amcgee:
@cooper-joe should we de-duplicate the `DHIS2 2.x` ?  Probably fine for now, but a bit redundant (always our problem here)

cooper-joe:
> @cooper-joe should we de-duplicate the `DHIS2 2.x` ? Probably fine for now, but a bit redundant (always our problem here)

Yeah I think it's fine for now. There doesn't seem to be an agreed format (I see both DHIS2 2.x and DHIS 2.x all over our website), so I think for clarity let's keep as is 👍.

PR #1161, opened 2022-10-17
feat: allow alerts to be undismissable [LIBS-360]

amcgee:
Implements [LIBS-360](https://dhis2.atlassian.net/browse/LIBS-360)

---

### Description

This implements an additional `dismissible` flag in the AlertBar props which optionally disables the "dismiss" X button

---

### Checklist

-   [ ] Tests were added

---

### Screenshots

_supporting text_

### DIFF
--- a/components/alert/src/alert-bar/action.js
+++ b/components/alert/src/alert-bar/action.js
@@ -5,7 +5,9 @@ import React, { Component } from 'react'
 class Action extends Component {
     onClick = (event) => {
         this.props.onClick(event)
-        this.props.hide(event)
+        if (this.props.hideOnClick) {
+            this.props.hide(event)
+        }
     }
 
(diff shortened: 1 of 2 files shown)


Mohammer5:
So far we've tried to keep boolean prop names to be not a negation (:heavy_check_mark:) but also to default to "false" (:negative_squared_cross_mark:). [Here](https://github.com/dhis2/ui/blob/b9b9b51a42419c0c89421fb4d034c7675a88be39/components/organisation-unit-tree/src/organisation-unit-tree/organisation-unit-tree.js#L216) is a good example.

I guess in this case it's a shame that `permanent` already has a different meaning, which does not reflect the actual meaning of that word. Maybe something like `irremovable` or `undismissible`?

PR #1162, opened 2022-10-18
feat: add multi-calendar component

kabaros:
Implements [DHIS2-14055](https://dhis2.atlassian.net/browse/DHIS2-14055)

docs link to test the component: https://pr-1162--dhis2-ui.netlify.app/components/calendar

PR in Data Quality app that uses this component: https://github.com/dhis2/data-quality-app/pull/1001 

---

### Description

This implements a component to pick a day in various calendar systems. For DHIS2 , the main non-Gregorian calendars we support are Nepali and Ethiopic calendars, but the component supports other calendars as well supported by the [Temporal proposal](https://tc39.es/proposal-temporal).

- The logic that powers the component lives in a different dhis2 library: [multi-calendar-dates](https://github.com/dhis2/multi-calendar-dates/tree/beta). The library abstracts away the complexity of dealing with dates and different calendar systems, and this component should just be a visual layer on top of it. Optionally, for some extra context, you can check [the library](https://github.com/dhis2/multi-calendar-dates) or at least read [this ADR](https://github.com/dhis2/multi-calendar-dates/blob/beta/doc/architecture/decisions/0002-use-temporal-api-as-the-backbone-for-the-engine.md) about why we went with Temporal to support the engine, and these [design specs document](https://docs.google.com/document/d/19zjyB45oBbqC5KeubaU8E7cw9fGhFc3tOXY0GkzZKqc/edit?userstoinvite=hendrik%40dhis2.org#heading=h.rjt0etsbsqh6).
- The component uses one hook exposed by the engine [useDatePicker](https://github.com/dhis2/multi-calendar-dates/blob/beta/src/hooks/useDatePicker.ts) which returns all the information needed to render the calendar and its navigation in different locales.
- The calendar is localised using the browser Internationalisation API. This support 90+ languages (and 300+ locales) natively in all modern browsers. It's used to localise elements of the UI like month, day names, numbering system etc... 
    -  This is also abstracted by the `multi-calendar-dates` library and `useDatePicker` hook.
- The HTML structure is similar to the structure of other open-source calendar components (react-day-picker specifically) to make it keyboard accessible and have better WAI-ARIA support.
---

### Checklist

-   [x] Tests were added

[DHIS2-14055]: https://dhis2.atlassian.net/browse/DHIS2-14055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/calendar/src/calendar-input/index.js
+++ b/components/calendar/src/calendar-input/index.js
@@ -87,8 +87,7 @@ export const CalendarInput = ({
             </div>
             {open && (
                 <Layer
-                    onBackdropClick={(_, evt) => {
-                        evt.stopPropagation()
+                    onBackdropClick={() => {
                         setOpen(false)
                     }}
                 >
--- a/components/calendar/src/calendar/calendar-table-cell.js
+++ b/components/calendar/src/calendar/calendar-table-cell.js
@@ -2,56 +2,40 @@ import { colors, spacers } from '@dhis2/ui-constants'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React from 'react'
-import { StyleOptionsProps } from './calendar-prop-types.js'
 
-export const CalendarTableCell = ({ day, cellSize, width, styleOptions }) => {
+export const CalendarTableCell = ({ day, cellSize }) => {
+    const dayHoverBackgroundColor = colors.grey200
+    const selectedDayBackgroundColor = colors.teal700
+
     return (
-        <>
-            <td data-test={day?.calendarDate} onClick={day.onClick}>
-                <button
-                    name="day"
-                    className={cx('day', {
-                        isSelected: day.isSelected,
-                        isToday: day.isToday,
-                        otherMonth: !day.isInCurrentMonth,
-                    })}
-                >
-                    {day.label}
-                </button>
-            </td>
+        <td data-test={day?.calendarDate} onClick={day.onClick}>
+            <button
+                name="day"
+                className={cx('day', {
+                    isSelected: day.isSelected,
+                    isToday: day.isToday,
+                    otherMonth: !day.isInCurrentMonth,
+                })}
+            >
+                {day.label}
+            </button>
             <style jsx>{`
-                .calendarTable td {
+                td {
                     width: ${cellSize};
                     height: 100%;
                     height: ${cellSize};
                     text-align: center;
+                    border: 2px solid transparent;
                 }
-                .calendarTable th,
-                .calendarTable td,
-                .calendarTable tr {
-                    border: 0;
-                }
-                .calendarTable td span {
+                td span {
                     display: inline-flex;
                     padding: 2px 2px;
                     justify-content: center;
                     border: 2px solid transparent;
                     min-width: 16px;
                 }
-                .selectedDate {
-                    width: ${width};
-                    padding: 4px 8px;
-                }
-                .selectedDate span {
-                    justify-content: center;
-                    font-size: 0.8em;
-                }
                 button {
-                    background: none;
                     border: 0;
-                }
-
-                button[name='day'] {
                     overflow: hidden;
                     width: ${cellSize};
                     height: ${cellSize};
@@ -61,13 +45,13 @@ export const CalendarTableCell = ({ day, cellSize, width, styleOptions }) => {
                     margin: ${spacers.dp2};
                     color: ${colors.grey900};
                 }
-
-                button[name='day']:hover {
-                    background-color: ${styleOptions.dayHoverBackgroundColor};
+                button:hover {
+                    background-color: ${dayHoverBackgroundColor};
                     text-decoration: underline;
                 }
-                button[name='day'].isSelected {
-                    background-color: ${styleOptions.selectedDayBackgroundColor};
+                button.isSelected,
+                button.otherMonth.isSelected {
+                    background-color: ${selectedDayBackgroundColor};
                     color: white;
                 }
                 button.isToday {
@@ -77,7 +61,7 @@ export const CalendarTableCell = ({ day, cellSize, width, styleOptions }) => {
                     color: ${colors.grey600};
                 }
             `}</style>
-        </>
+        </td>
     )
 }
 
@@ -91,6 +75,4 @@ CalendarTableCell.propTypes = {
         label: PropTypes.string,
         onClick: PropTypes.func,
     }),
-    styleOptions: StyleOptionsProps,
-    width: PropTypes.string,
 }
--- a/components/calendar/src/calendar/calendar-table.js
+++ b/components/calendar/src/calendar/calendar-table.js
@@ -1,75 +1,54 @@
 import { spacers } from '@dhis2/ui-constants'
 import PropTypes from 'prop-types'
 import React from 'react'
-import { StyleOptionsProps } from './calendar-prop-types.js'
 import { CalendarTableCell } from './calendar-table-cell.js'
 import { CalendarTableDaysHeader } from './calendar-table-days-header.js'
 
 export const CalendarTable = ({
     weekDayLabels,
     calendarWeekDays,
-    styleOptions,
     width,
     cellSize,
-}) => {
-    return (
-        <>
-            <div className="calendarTableWrapper">
-                <table className="calendarTable">
-                    <CalendarTableDaysHeader
-                        weekDayLabels={weekDayLabels}
-                        styleOptions={styleOptions}
-                    />
-                    <tbody>
-                        {calendarWeekDays.map((week, weekIndex) => (
-                            <tr key={`week-${weekIndex + 1}`}>
-                                {week.map((day) => (
-                                    <CalendarTableCell
-                                        day={day}
-                                        key={day?.calendarDate}
-                                        styleOptions={styleOptions}
-                                        cellSize={cellSize}
-                                        width={width}
-                                    />
-                                ))}
-                            </tr>
+}) => (
+    <div className="calendar-table-wrapper">
+        <table className="calendar-table">
+            <CalendarTableDaysHeader weekDayLabels={weekDayLabels} />
+            <tbody>
+                {calendarWeekDays.map((week, weekIndex) => (
+                    <tr key={`week-${weekIndex + 1}`}>
+                        {week.map((day) => (
+                            <CalendarTableCell
+                                day={day}
+                                key={day?.calendarDate}
+                                cellSize={cellSize}
+                                width={width}
+                            />
                         ))}
-                    </tbody>
-                </table>
-            </div>
-            <style jsx>{`
-                .wrapper {
-                    font-family: Roboto, sans-serif;
-                    font-weight: 400;
-                    font-size: 14px;
-                    background-color: ${styleOptions.backgroundColor};
-                    display: flex;
-                    flex-direction: column;
-                    border: 1px solid ${styleOptions.wrapperBorderColor};
-                    border-radius: 3px;
-                    min-width: ${width};
-                    width: max-content;
-                    box-shadow: 0px 4px 6px -2px #2129340d;
-                    box-shadow: 0px 10px 15px -3px #2129341a;
-                }
-                .calendarTable {
-                    border: none;
-                    border-collapse: collapse;
-                    width: 100%;
-                    margin-block: ${spacers.dp4};
-                }
-                .calendarTableWrapper {
-                    padding-inline: ${spacers.dp4};
-                }
-            `}</style>
-        </>
-    )
-}
+                    </tr>
+                ))}
+            </tbody>
+        </table>
+        <style jsx>{`
+            .calendar-table {
+                border: none;
+                border-collapse: collapse;
+                width: 100%;
+                margin-block: ${spacers.dp4};
+            }
+            .calendar-table tr,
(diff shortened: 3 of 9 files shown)


Mohammer5:
Why TypeScript in a JS repository?

cooper-joe:
I see this component is going in as `experimental` and the main work has been the logic and implementation, but for reference I've been working on visual/ux design of a date picker ([draft](https://www.figma.com/proto/fwcesC5CkqSnBQcL9nNCfn/ds-scratchpad?page-id=625%3A5040&node-id=753%3A9706&viewport=-472%2C126%2C0.42&scaling=min-zoom)).

Mohammer5:
> is there a way in dhis/ui to mark a component as "experimental" or beta?

There's been no precedent for this. If I'm not mistaken, then our current tooling requires all workspaces to have the same version, so this one can't have a beta version number. Maybe we could remove the multi calendar from the list of workspaces in the root package.json? cc @amcgee 

If that's not possible, then we could simply not bundle it in @dhis2/ui and release it as a standalone. It wouldn't be a proper beta version, but at least it wouldn't show up in the main UI library

ismay:
> there is no real need to mark it as experimental .. the component's API is stable now, and if it changes, I guess we just follow semver. I was just wondering if we have a pattern for marking something so that consumer could experiment with it, while not holding ourselves too early to a contract (even if it's just by prefixing its name like React did).

With our current release tooling, if you want to release something as experimental you have to do it for all the packages. But that's a perfectly fine thing to do. You could just merge this to a prerelease branch and it would be released as a prerelease version.

ismay:
I would err on the side of caution. If you're not really sure and feel it's prudent to do an alpha or beta release first then I'd do that. See here: https://github.com/dhis2/ui/blob/master/CONTRIBUTING.md#automated-releases for recommendations on what merge types to use. Either beta or alpha are what we've used in the past for prereleases (though I'd have to look up our config to see what we've defined as acceptable prerelease branches).

tomzemp (commented):
Looks really cool to me @kabaros 🙌🗓!

I just looked here at the calendar ui-component and not the multi-calendar-dates library (so I apologize if I ask any particularly stupid questions)

A couple of things that I noticed:
- it would be nice if the documentation included something about the `numberingSystem`s that are allowed (I see the default is `geor`, which I guess I thought was gregorian, but then I was wondering if this was something else)? I guess also, I'm a bit confused as Nepalese custom calendar seems to be using Nepalese numbers despite what is specified here? 
- It would be nice if the component featured the ability to specify the starting day of the week (e.g. North American calendars start weeks on Sunday)
- wondering if there's going to be an option to specify/differentiate unselectable dates in a range? E.g. because sometimes one wants to disable selection of future dates (or past dates out of range), etc.?
- I do not really understand the Ethiopian calendar 😔, but I noticed that the Calendar Date text in the storybook component seems to not match the date in the datepicker:
image
(Not sure if this is right? I guess I would have thought the year would have been 2015 and not 7515?) The Nepalese and Arabic calendars matched the year from the date picker? I also looked at the Japanese calendar and it just showed the ISO date (which again, I'm not sure is right or not-right?)
- I noticed that the 日 character is being cut off when using Chinese or Japanese (the language, irrespective of the calendar) :
image
I don't know if it's worth addressing given that DHIS2 isn't much used in these language communities. It's a bit strange though that these languages come with this day marker in the the calendar; is that from CLDR? It's not particularly common to include 日 in date pickers in Chinese (to my knowledge) as it's a bit like putting 'st/nd/rd/th' after numbers in English. All the other calendars I looked at looked good, so maybe just having the option to change the cell size is sufficient?

components/calendar/src/calendar.js
```diff
@@ -0,0 +1,368 @@
...
+import { useDatePicker } from 'multi-calendar-dates'
+import PropTypes from 'prop-types'
+import React, { useState } from 'react'
+
+export const Calendar = ({
+    onDateSelect,
+    calendar,
+    initialDate,
+    dir = 'ltr',
+    locale = 'en',
```
tomzemp, line 13: would it be worth it to specify a default non-North American variant of English (e.g. `en-GB`, `en-NG`, etc.,) so that our dates e.g. if displayed when `showSelectedDate=true` are formatted in DD MM YYYY rather than the US/Canada MM DD YYYY format?

components/calendar/src/calendar.js
```diff
@@ -0,0 +1,368 @@
...
+    numberingSystem = 'geor',
+    showSelectedDate = false,
+    weekDayFormat = 'narrow',
+    timeZone,
+    styleOptions: {
```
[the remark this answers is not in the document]
Mohammer5: These should be flat, not inside an object. Otherwise - at least the way it's currently implemented - you'd have to supply the entire object with all of the default values.

Either
* you merge the custom style options with the default style options or 
* you flatten the props

I'm in favor of flattening:
* makes the props' shape simpler
* removes the necessity to merge options
* looks cleaner when you need to change a single style only.
* Some of the properties are not "styles" in the sense of css, so the name could be confusing
* The flat approach is what the rest of the code base is using (e.g. the [`<Box />`](https://github.com/dhis2/ui/blob/master/components/box/src/box.js#L38:L50))

components/calendar/src/calendar.js
```diff
@@ -0,0 +1,361 @@
...
+                }
+            `}</style>
+        </div>
+    )
+}
+
+Calendar.defaultProps = {
+    locale: 'en-GB',
+    numberingSystem: 'latn',
+    timeZone: Intl?.DateTimeFormat?.().resolvedOptions?.()?.timeZone || 'UTC',
```
HendrikThePendric, line 326: I'm not 100% sure that in this case providing defaults here is a good idea.... As I envisioned it, the hook should be able to resolve sensible default values for all these date/time related properties. And ultimately I'd like these values to be resolved using the following prioritisation scheme:
1. Prop-values passed to the hook/component
2. Properties available from the system/user settings resource*
3. Values coming from `Intl.DateTimeFormat().resolvedOptions()`
4. Some hardcoded defaults

[*] I'm not sure this is easy to implement right now. Quite likely we need some sort of date-internationalisation provider that holds this info first. But still good to keep this in mind, and also even when skipping this step for now, what I'm suggesting still makes sense.

So if we implement things in the hook as I suggest above, then using default props here breaks the "prioritisation scheme" I've outlined above.

HendrikThePendric (commented):
Very awesome!

I've left a few comments which we should probably discuss in more detail.

I also wonder if perhaps we should also produce a `CalendarInputField` component in the current iteration. In most apps that need multi-calendar support, we will need to actually replace date inputs, so having a similar component seems necessary.

components/calendar/src/calendar.js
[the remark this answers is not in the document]
cooper-joe: I think it probably makes sense to only allow the minimal style options (`cellSize` and `width` as you mention above). Customizing the rest of the values should be handled by theme values/tokens if we want to do something like that in future, rather than overriding via props. So I vote to remove all non-needed style options.

HendrikThePendric (changes requested):
Great work @kabaros, thanks a lot. I've left quite a few comments and I've also set the review status to "Changes requested", but not all my comments represent "required" changes, some are just suggestions. The changes I do see as required are these three:
- Remove the default prop
- Reuse components from `@dhis2/ui`
- Split the Calendar component JSX/CSS into smaller presentational components

components/calendar/src/calendar.js
```diff
@@ -0,0 +1,361 @@
...
+        options,
+    })
+
+    const PreviousIcon =
+        languageDirection === 'ltr' ? IconChevronLeft16 : IconChevronRight16
+    const NextIcon =
+        languageDirection === 'ltr' ? IconChevronRight16 : IconChevronLeft16
+
+    return (
+        <div>
```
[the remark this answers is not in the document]
HendrikThePendric: BTW: this comment really applies to everything the component returns, but I didn't want to select that many lines...

components/calendar/src/calendar-input/index.js
```diff
@@ -0,0 +1,116 @@
...
+                    {...rest}
+                    type="text"
+                    onFocus={onFocus}
+                    value={value}
+                />
+            </div>
+            {open && (
+                <Layer
+                    onBackdropClick={(_, evt) => {
+                        evt.stopPropagation()
```
HendrikThePendric, line 91: This shouldn't be needed. In the apps all layers are appended to the layer root, which is a `div` directly under the body. So it seems unlikely this event would need to stop bubbling up.

components/calendar/src/calendar/calendar-table-cell.js
```diff
@@ -0,0 +1,96 @@
...
+                    background-color: ${styleOptions.selectedDayBackgroundColor};
+                    color: white;
+                }
+                button.isToday {
+                    text-decoration: underline;
+                }
+                button.otherMonth {
+                    color: ${colors.grey600};
+                }
+            `}</style>
```
HendrikThePendric, line 79: These styles look off to me, considering styled-jsx will produce an encapsulated stylesheet. 

This component only contains a `td` without a class name and a `button` which can have 3 dynamic classes. So any style rules that don't apply to these 2 tags or 3 dynamic classes are out of place here (and also they won't have any effect). 

Furthermore, some of the selectors being used are much more specific than they need to be, considering this is an encapsulated stylesheet. No need to use things like `button[name=day]`, there is only one button tag in this component so just use `button`.

Finally, you don't have to render a Fragment (`<>...</>`), you can just nest the `<style jsx>...</style>` under the `td`.

components/calendar/src/calendar/index.js
```diff
@@ -0,0 +1,105 @@
...
+        backgroundColor: 'none',
+        chevronColor: colors.grey600,
+        dayHoverBackgroundColor: colors.grey200,
+        dayNamesColor: colors.grey700,
+        headerBackground: colors.grey050,
+        selectedDayBackgroundColor: colors.teal700,
+        wrapperBorderColor: colors.grey300,
+        width,
+        cellSize,
+    }
```
HendrikThePendric, line 37: This doesn't make sense to me. This object only has 2 dynamic properties. I see you've done this because the header and the table cells need similar styles. But these are different tags (`td` vs `th`). IMO it's fine to let both have their individual style rules which happen to be identical right now. If you don't like this, you could write a shared style sheet that targets both the `td` and the `th` tag.

In any case, the main point here is that we don't pass a style options object. Instead just pass the dynamic props directly to the components that need them, and use them in the styled-jsx there. This is a more transparent way of working imo.

components/calendar/src/calendar/navigation-container.js
```diff
@@ -0,0 +1,177 @@
...
+                }
+                .navigationContainer > .curr {
+                    justify-content: center;
+                }
+                .navigationContainer > .next {
+                    justify-content: flex-end;
+                }
+                .navigationContainer > .today {
+                    grid-area: today;
+                }
```
HendrikThePendric, line 145: I don't think these styles do anything, since there are no direct descendants of the `.navigationContainer` with these class names

HendrikThePendric (approved):
Looking good. Still some minor points that could be addressed though.

components/calendar/src/calendar/navigation-container.js
```diff
@@ -0,0 +1,177 @@
...
+                .today > button {
```
tomzemp, line 146: I don't see a today class at all? (I'm not sure if this is captured by Hendrik's previous comment or is this additional)

components/calendar/src/calendar/calendar-table.js
```diff
@@ -0,0 +1,75 @@
...
+                                        width={width}
+                                    />
+                                ))}
+                            </tr>
+                        ))}
+                    </tbody>
+                </table>
+            </div>
+            <style jsx>{`
+                .wrapper {
```
tomzemp, line 41: there's no wrapper class in this component, right? could we remove?

components/calendar/src/calendar/calendar-table.js
```diff
@@ -0,0 +1,75 @@
...
+                .calendarTableWrapper {
+                    padding-inline: ${spacers.dp4};
+                }
+            `}</style>
+        </>
+    )
+}
+
+CalendarTable.propTypes = {
+    calendarWeekDays: PropTypes.string,
```
tomzemp, line 70: wrong prop type?

components/calendar/src/calendar/calendar-table.js
```diff
@@ -0,0 +1,75 @@
...
+    cellSize: PropTypes.string,
+    styleOptions: StyleOptionsProps,
```
tomzemp, line 72: do you need to do `styleOptions: PropTypes.shape(StyleOptionsProps),`? (this would apply on the proptypes in a couple of components...but also might not be relevant if you implement Hendrik's comment on styleOptions)

tomzemp (approved):
looks really good 🙌

I left some minor comments on things I noticed.

In terms of the translations for aria labels, I'm not sure if anything is needed to configure this with transifex somehow? I don't see the calendar as resource under ui app component currently. (I think this is a minor issue, just something that I thought about when seeing the translations there)

PR #1163, opened 2022-10-18
feat: online/offline status messaging

DNR800:
Jira: [LIBS-230](https://dhis2.atlassian.net/browse/LIBS-230)

---

### Description

This branch builds on the usage of the useOnlineStatusMessage hook https://github.com/dhis2/app-runtime/pull/1258

This hook is used to add messages next to the online/offline status of the header bar in PWA apps like...

Online status message

---

### Known issues

-   This will appear broken until https://github.com/dhis2/app-runtime/pull/1258 can be merged and the version of `app-runtime` in this project is bumped.

---

### Checklist

-   [x] Tests were added

---

### Screenshots

Online status message

And for mobile where there is a lot of text wrapping should already be handled by the OnlineStatus component...

(WIP - this might need a little work to prevent compression of some the online/offline icon)

Message wraps when very long


DNR800:
Just something to note here too, running the cypress tests I ran into a problem with storybook running on port 5001 and Cypress expecting everything to run on port 5000 - despite my best efforts I couldn't kill the process on my mac that was running on 5000 already so my initial work around was to switch cypress to run on port 5001.

More googling found this issue 

https://medium.com/pythonistas/port-5000-already-in-use-macos-monterey-issue-d86b02edd36c

And searching through dhis2 slack I can see that people have had this problem before

https://dhis2.slack.com/archives/CJF3MDK8Q/p1638355389090300

Turning off the airplay receiver in my settings fixed this - but would it be worth thinking about changing storybook and cypress to run on 5001 to avoid anyone else experiencing the pain of this?

components/header-bar/src/online-status.js
```diff
@@ -18,15 +22,24 @@ const useOnlineStatusInfo = ({ online, lastOnline }) => {
             relativeTime: moment(lastOnline).fromNow(),
         })
     }
-    // todo: in the future, support 'CUSTOM' option
+
+    if (onlineStatusMessage) {
+        return onlineStatusMessage
```
DNR800, line 27: I've a question about this functional I think I have already partial covered this in a conversation with @KaiVandivier but would be good to get clarity.

Rummaging though the tickets I found a design that looked like this..

message with lastonline

Mentioned in https://docs.google.com/document/d/1BRU5yoiZDcgVDLJeFXih8uoI2DnjONOApbgwi1qU1FQ/edit#heading=h.7sxyqvjcxlix

So I wonder if the `onlineStatusMessage` needs to display alongside time `lastOnline`

2 main questions from this

1. Do we need to retain the the config check of `headerbar?.onlineStatusInfo === 'LAST_ONLINE'` for backward compatibility?
2. If we are retaining that config check, given the design I mention above should `onlineStatusMessage` act more like a prefix? So rather than either `lastOnline` or `onlineStatusMessage` or would be (pseudo code)`<>{onlineStatusMessage} {lastOnline}</>` so both could be potentially displayed? The other option here is that if you set `onlineStatusMessage` with a custom message and want the last online time added that you would have to specify that yourself in the custom message.

DNR800:
Again cypress test related..

I ran into a problem running one of the HeaderBar cypress tests locally

line 29 of `components/header-bar/src/features/the_headerbar_can_display_online_status.feature`

```
Scenario: The HeaderBar displays an offline status when offline
        Given the HeaderBar loads without error when PWA is enabled
        And the browser goes offline
        Then the status badge shows offline
```

Even on master this fails locally (I'm guessing it probably runs fine on CI though)

Seems to be Cypress code related from the error

Cypress error - headerbar - go offline

Is there anything I can do to get this to work locally? As it was failing on master for me I've worked around it for now

.yarnrc.yml
```diff
@@ -0,0 +1 @@
+yarnPath: .yarn/releases/yarn-1.22.19.cjs
```
KaiVandivier, line 1: I don't think we should be checking this and the target yarn file into version control here

components/header-bar/src/online-status.js
```diff
@@ -18,15 +22,24 @@ const useOnlineStatusInfo = ({ online, lastOnline }) => {
...
 export function OnlineStatus({ dense }) {
     const { online, lastOnline } = useOnlineStatus()
-    const info = useOnlineStatusInfo({ online, lastOnline })
+    const { onlineStatusMessage } = useOnlineStatusMessage()
+
+    const info = useOnlineStatusInfo({
+        online,
+        lastOnline,
+        onlineStatusMessage,
+    })
```
KaiVandivier, line 42: Following from the above, basically this would become: 
```suggestion
    const info = onlineStatusMessage
```
... which basically makes the info variable unnecessary

package.json
```diff
@@ -75,5 +75,6 @@
         "rimraf": "^3.0.2",
         "styled-jsx": "^4.0.1",
         "wait-on": "^6.0.0"
-    }
+    },
+    "packageManager": "yarn@2.4.2"
```
KaiVandivier, line 79: What's this for? If I recall correctly, I think we specifically want `yarn 1` for this repo

yarn.lock
```diff
@@ -3511,30 +3511,64 @@
...
+
 "@dhis2/app-service-alerts@3.2.5":
   version "3.2.5"
   resolved "https://registry.yarnpkg.com/@dhis2/app-service-alerts/-/app-service-alerts-3.2.5.tgz#008510fbe9dbccc1114ebbba0b880ad812f004e3"
   integrity sha512-oZuijTiko7+QF8xHWAqoJXqOxcd4FNpabmGNsWX1WD4kdMq22h7pHIXC9DLg7dYKbnlckihrzq4G9BvOHs/3Dg==
 
+"@dhis2/app-service-alerts@3.4.4":
+  version "3.4.4"
+  resolved "https://registry.yarnpkg.com/@dhis2/app-service-alerts/-/app-service-alerts-3.4.4.tgz#ccd8848f416e976a01ea11bbbd6e1ae873136777"
+  integrity sha512-o2DHlZIHxJQ4XwJ9uIr0UpMTi3ZK0W6u+FkSnPGwrvF+MCFsxA0+pjN7PfBo4acCQHPBwTZzcyq0WYiEAJM7OA==
```
KaiVandivier, line 3532: I believe you'll need to deduplicate this package here using `npx yarn-deduplicate yarn.lock`

components/header-bar/src/online-status.js
DNR800: Awesome, thanks @KaiVandivier thats make sense and simplifies whats happening here too :-)

yarn.lock
DNR800: Thanks :-)

components/header-bar/src/online-status.js
DNR800: Yep totally :-)

package.json
DNR800: I'm not even sure how that got in there.. I'm on a version of yarn 1.22 myself. I'll double check whats happening with the package.json files

components/header-bar/src/online-status.js
DNR800: Directly using `onlineStatusMessage` now

DNR800:
The failing cypress test seems to be a cypress issue

https://github.com/cypress-io/cypress/issues/17723

Thanks @KaiVandivier for the heads up on that

I could get the HeaderBar tests to run by resolving the version of cypress to an older version, add this to the root package.json

```
"resolutions": {
     "cypress": "6.9.1"
 },
```
this confirms whats on the issue thread above. Thats quite a few major versions back and might result in other problems I would be cautious about using it.

I tried resolving to version to a 10.X.X version but it looks like there have been a few changes since 9.7.0, it might be worth checking if this is still a problem after adopting a 10.x.x version of cypress.

For now I've added a comment to the test that fails as I don't have a fix for that right now.

DNR800:
I've currently got two tests that are also failing on this build for Poppers - oddly when I checkout master locally and run them they fail on master as well. Starting to look at this 

Screenshot 2022-10-20 at 18 07 20

Screenshot 2022-10-20 at 18 07 37

HendrikThePendric (approved):

Birkbjo (approved):
LGTM

PR #1175, opened 2022-10-31
chore: add adrs

DNR800:
### Description
This follows on from the retro that the team had - The point was raise that we would like a way to process and formalise technical decisions that supports an open and democratic process.

I have put together and ADR about adopting ADR's - you can read more on that in the ADR featured in this PR - 0001-adr-adoption.md

I've selected this codebase not for any specific reason, so if its felt that another repo should be used for introducing ADR's then please do say.

This PR should discussed by the team before being merged - this is PR informs the adoption of a process not just a change to this repo

### DIFF
--- a/decisions/0001-adr-adoption.md
+++ b/decisions/0001-adr-adoption.md
@@ -1,0 +1,104 @@
+# ADR adoption
+
+### Date: 31/10/2022
+
+### Status: proposed
+
+## Context
+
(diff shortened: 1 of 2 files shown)


Mohammer5:
I'm wondering if the [notes](https://github.com/dhis2/notes) repository might be a better candidate to put these files. Was there a specific reason you added them to the UI library?

PR #1177, opened 2022-11-02
fix(sharing-dialog): variable initialization error [DHIS2-14029]

KaiVandivier:
Implements [DHIS2-14029](https://dhis2.atlassian.net/browse/DHIS2-14029)

---

### Description

Fixes an `Uncaught ReferenceError: Cannot access 'mutation' before initialization` bug that crashes a consuming app. The `useDataEngine` hook doesn't accept any parameters so it's safe to remove there, and the `mutation` object is used below at `engine.mutate()`

---

### Checklist

-   [X] Tests were added

### DIFF
--- a/components/sharing-dialog/src/cascade-sharing/controls.js
+++ b/components/sharing-dialog/src/cascade-sharing/controls.js
@@ -38,7 +38,7 @@ export const Controls = ({ id, entityAmount }) => {
      * so we're using the engine directly as a workaround.
      */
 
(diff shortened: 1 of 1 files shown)


HendrikThePendric (approved):
Having a reference to the DHIS2 issue in the branch name probably confuses the GitHub-JIRA integration a bit. But no need to change the branch name just because of that. Changes make sense.

PR #1187, opened 2022-11-21
feat(selection-bar-item): [LIBS-371] display only and clear selection callback 

simonadomnisoru:
Implements [LIBS-371](https://dhis2.atlassian.net/browse/LIBS-371)

---

### Description

Allow clearing of single selections from `SelectionBarItem` component.
Add a `displayOnly` SelectionBarItem component
[Figma prototype](https://www.figma.com/proto/9zvqbQKtrWIElNQ8ZYxyha/capture-context-selector?page-id=0%3A1&node-id=3%3A3435&viewport=1740%2C62%2C0.28&scaling=min-zoom&starting-point-node-id=3%3A1134) 

---

### Checklist

-   [x] Tests were added
---

### Screenshots

Screenshot 2022-11-18 at 12 23 47

### DIFF
--- a/components/selector-bar/src/selector-bar-item/selector-bar-item.js
+++ b/components/selector-bar/src/selector-bar-item/selector-bar-item.js
@@ -47,8 +47,13 @@ export const SelectorBarItem = ({
     return (
         <button
             ref={buttonRef}
-            className={cx('selector-bar-item', className)}
+            className={cx(
+                'selector-bar-item',
+                className,
+                !displayOnly ? 'pointer' : ''
+            )}
             disabled={disabled}
+            onClick={() => setOpen && setOpen(true)}
             data-test={dataTest}
         >
             <span className="label">{label}</span>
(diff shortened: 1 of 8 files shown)


cooper-joe:
Pushed a commit to adjust icon and styling. Updated screenshot:
image

cooper-joe (approved):
Design LGTM 👍

components/selector-bar/src/selector-bar-item/selector-bar-item.js
```diff
@@ -29,32 +29,62 @@ export const SelectorBarItem = ({
...
     const Icon = open ? IconChevronUp24 : IconChevronDown24
 
     return (
         <button
             ref={buttonRef}
-            className={cx('selector-bar-item', className)}
+            className={cx(
+                'selector-bar-item',
+                className,
+                !displayOnly ? 'pointer' : ''
```
Mohammer5, line 49: Nitpick: The class name `pointer` includes the style that's supposed to be applied. IMO the class name should be descriptive & declarative, so another name would be more appropriate, maybe `openable` (not sure, but I suppose that's the intention here, right?)? 

Then the style applied when the selector bar item is openable can be changed without having to either change the or have a wrong  class name.

This is super nit picky though, so I'll leave that decision up to you

Mohammer5 (approved):
LGTM code-wise! :tada: 
Just one minor thing that I'll leave up to you

PR #1188, opened 2022-11-22
fix(menu): adjust visual styles

cooper-joe:
Implements [UX-122](https://dhis2.atlassian.net/browse/UX-122).

---

### Description

This PR makes visual style adjustments to the `MenuItem` and `MenuSectionHeader` components. These are minor changes based on seeing the components in use.

---

### Known issues

-   [x] Using a negative margin to pull the `Chevron` further to the right on items with children. Feels like a hack?

---

### Checklist

-   [x] Tests were added

### DIFF
--- a/components/menu/src/menu-item/menu-item.styles.js
+++ b/components/menu/src/menu-item/menu-item.styles.js
@@ -8,18 +8,12 @@ export default css`
         padding: 0;
         cursor: pointer;
         list-style: none;
-        transition: background-color 150ms ease-in-out;
         background-color: ${colors.white};
         color: ${colors.grey900};
         fill: ${colors.grey900};
-        font-size: 15px;
-        line-height: 18px;
-        user-select: none;
-    }
-
-    li.dense {
         font-size: 14px;
         line-height: 16px;
+        user-select: none;
     }
 
     li:hover {
@@ -28,7 +22,7 @@ export default css`
 
     li:active,
     li.active {
-        background-color: ${colors.grey400};
+        background-color: ${colors.grey300};
     }
 
     li.destructive {
@@ -59,8 +53,8 @@ export default css`
         display: inline-flex;
         flex-grow: 1;
         align-items: center;
-        padding: 0 ${spacers.dp24};
-        min-height: 48px;
+        padding: 0 ${spacers.dp16};
+        min-height: 40px;
         text-decoration: none;
         color: inherit;
     }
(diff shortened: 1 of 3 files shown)


components/menu/src/menu-item/menu-item.styles.js
```diff
@@ -88,23 +82,26 @@ export default css`
...
         height: 24px;
     }
 
     .chevron {
+        display: flex;
+        align-items: center;
         flex-grow: 0;
-        margin-left: ${spacers.dp48};
+        margin-left: ${spacers.dp24};
+        margin-right: -${spacers.dp12};
```
cooper-joe, line 104: Not sure if this is wise? I couldn't see a way of pulling the `Chevron` icon further to the right without restructuring the component.

ismay: Negative margins can lead to visual bugs. What's your intention here visually? Do you want it to be further right than this?

image

This is what it looks like for me currently, without the negative right margin.

ismay: It seems there's a default padding of 8px on the `<a>`: https://github.com/dhis2/ui/blob/5cd52efd58424a6bbeb5efe8838fd65814814f9b/components/menu/src/menu-item/menu-item.styles.js#L70-L72 If you want to go right to the right edge of the background, removing that padding will get you there.

The chevron svg has a width of its own, so visually there's still some space left:

image

If you want to move it any further you'll have to tweak the svg styles (setting the viewbox).

ismay: The existing styles are quite messy in terms of the code btw. We really should clean them up. This goes for a lot of the css of the components in ui. Might be worth setting aside some time to do a thorough refactor of those.

cooper-joe: Ah, interesting. My intention was to move the chevron to the right when used inside a `FlyoutMenu`. This is how the chevron icon is displayed inside `FlyoutMenu`:
image

So, it looks like the issue is perhaps something to do with the wrong styles being applied. 👀

cooper-joe: > The existing styles are quite messy in terms of the code btw. We really should clean them up. This goes for a lot of the css of the components in ui. Might be worth setting aside some time to do a thorough refactor of those.

Sounds good. I'm working my way through some minor visual tweaks, now that components have been in use for a while and I've gotten some in-use feedback. Could be good to clean up the styles as part of this process.

cooper-joe: The `with-chevron` class wasn't properly being applied when a menu-item had children. Fixed this [here](https://github.com/dhis2/ui/pull/1188/files#diff-95560ff80b247b89e0ce8a8eb8fb62109251d8f8ff5a3a9588f2f68780782893R53), and removed the hacky negative margin!

Mohammer5 (approved):
OK from my side

components/menu/src/menu-item/menu-item.styles.js
ismay: > Sounds good. I'm working my way through some minor visual tweaks, now that components have been in use for a while and I've gotten some in-use feedback. Could be good to clean up the styles as part of this process.

Yeah I agree. Maybe we could raise this in our platform meeting and set aside some time for it.

> The with-chevron class wasn't properly being applied when a menu-item had children. Fixed this [here](https://github.com/dhis2/ui/pull/1188/files#diff-95560ff80b247b89e0ce8a8eb8fb62109251d8f8ff5a3a9588f2f68780782893R53), and removed the hacky negative margin!

Ah nice, good detective work! Thanks for the fix!

ismay (approved):

PR #1224, opened 2023-01-26
fix: add whitespace between chips using flexbox gap

HendrikThePendric:


HendrikThePendric:
Already implemented so closing

PR #1241, opened 2023-02-21
feat: allow clearing a calendar input

kabaros:
Implements [DHIS2-14788](https://dhis2.atlassian.net/browse/DHIS2-14788)

relates to https://github.com/dhis2/data-quality-app/pull/1001

---

### Description

The current `CalendarInput` doesn't allow to "clear" selection as it defaults to today's date when no date is selected. This is an issue that was found while integrating with Data Quality, and following Tom's testing round, as one section of the app (Outlier Detection -> Advanced) requires having a Calendar component with a clear button.

The changes in this PR add a new property `defaultToToday` which is set to true by default to preserve backwards compatibility for the behaviour, but can be set to false so that the component does not default to today's date and allow a _null_ selection.

- a related issue was that today's date always had selected style (as it's always the selected one). This was fixed as well.

---

### Known issues

none

---

### Checklist

-   [x] Tests were added

---

### Screenshots

clear-calendar

[DHIS2-14788]: https://dhis2.atlassian.net/browse/DHIS2-14788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/calendar/src/calendar-input/calendar-input.js
+++ b/components/calendar/src/calendar-input/calendar-input.js
@@ -77,10 +77,23 @@ export const CalendarInput = ({
                     value={date}
                 />
                 {clearable && (
-                    <div className="calendar-clear-button">
+                    <div
+                        className={cx('calendar-clear-button', {
+                            // ToDo: this is a workaround to show the clear button in the correct place when an icon is shown.
+                            // Long-term, we should abstract and share the logic multi-select uses for the input-wrapper
+                            // https://dhis2.atlassian.net/browse/DHIS2-14848
+                            'with-icon':
+                                rest.valid ||
+                                rest.error ||
+                                rest.warning ||
+                                rest.loading,
+                            'with-dense-wrapper': rest.dense,
+                        })}
+                    >
                         <Button
                             dataTest="calendar-clear-button"
                             secondary
+                            small
                             onClick={() => calendarProps.onDateSelect(null)}
                             type="button"
                         >
@@ -101,7 +114,7 @@ export const CalendarInput = ({
                         modifiers={[offsetModifier]}
                     >
                         <Card>
-                            <Calendar {...calendarProps} date={value} />
+                            <Calendar {...calendarProps} date={date} />
                         </Card>
                     </Popper>
                 </Layer>
@@ -114,8 +127,15 @@ export const CalendarInput = ({
                     }
                     .calendar-clear-button {
                         position: absolute;
-                        inset-inline-end: 2px;
-                        inset-block-end: 2px;
+                        inset-inline-end: 6px;
+                        inset-block-start: 27px;
+                    }
+
+                    .calendar-clear-button.with-icon {
+                        inset-inline-end: 36px;
+                    }
+                    .calendar-clear-button.with-dense-wrapper {
+                        inset-block-start: 23px;
                     }
                 `}
             </style>


HendrikThePendric (approved):

tomzemp:
I'm a little confused now @kabaros (but this might be to vocabulary choices on the variables 😛)

With the current implementation, there's a [useEffect](https://github.com/dhis2/ui/blob/813622ab960a7530f772ff98bf04ddd613bdffc5/components/calendar/src/calendar-input/index.js#L49-L53) that's enforcing that the the value of the CalendarInput matches the initialDate.

So in the storybook example with the clear button, that works fine because the [onDateSelect](https://github.com/dhis2/ui/blob/813622ab960a7530f772ff98bf04ddd613bdffc5/components/calendar/src/stories/calendarInput.stories.js#L98) callback you pass to the CalendarInput component sets the date which then gets passed back to the CalendarInput (so then the `value` in the CalendarInput matches the `initialDate` and the useEffect hook doesn't have to change anything)

In the other storybook examples though, there is not the pass up and back down of the selectedDate to CalendarInput, so it appears that when you select a date, it does not get selected (because the useEffect sets `value` back to the initialDate), e.g. (I select 29 Oct but the CalendarInput is still on the initial passed `date` of 13 Oct:

 calendar_initial_date

I guess my expectation would be that the CalendarInput would be self-contained and that you could update the date within it and it would "remember" that regardless of whether you pass that back up with `onDateSelect`?

Maybe what is in place is the desired behaviour though? In that case, maybe it'd just be sufficient to change the variable name in the CalendarInput component to be something other than `initialDate` (maybe `dateFromParent`?), so that it's clearer that these need to be kept in sync?

HendrikThePendric (approved):

HendrikThePendric:
> When there is a consuming component, it will normally track the value of the calendar in its own state. So when the calendar is passed a date, that date should be reflected in the parent state as well as the calendar. Then any subsequent changes to the calendar should propagate back up to update the state.

I don't think the above is completely accurate. Our input components should be fully controlled. So the consuming component should _set_ the date rather than _track_ it. IMO this is a very important distinction.

And, I guess this is only partly related, when the date/value passed to input is `null`/`undefined`, then I think the internal state of the input can be undefined too. And when opening the actual calendar component, this can have an internal state that falls back to "today".

This actually also makes me think that we may not need a `defaultsToToday` prop. The input doesn't need a default value, and the calendar can just default to today whenever the input doesn't have a value.

Perhaps all of the above does introduce the need to expose a helper function that allows the consuming component to compute a correct date string based the current calendar system, which can be used by the input....

collections/ui/API.md
```diff
@@ -204,6 +204,34 @@ import { SplitButton } from '@dhis2/ui'
...
+
+
+#### Props
+
+|Name|Type|Default|Required|Description|
+|---|---|---|---|---|
+|calendar|any||*|the calendar to use such gregory, ethiopic, nepali - full supported list here: https://github.com/dhis2/multi-calendar-dates/blob/main/src/constants/calendars.ts|
+|cellSize|string|`'32px'`||the size of a single cell in the table forming the calendar|
+|date|string|||the current selected date in the format YYYY-MM-DD - if nothing provided, it will default to today|
+|defaultToToday|boolean|||whether to default to today's date or not|
```
tomzemp, line 226: isn't this no longer relevant? (also the text on `date` immediately above saying that `it will default to today`)?

components/calendar/API.md
```diff
@@ -0,0 +1,27 @@
...
+
+
+#### Props
+
+|Name|Type|Default|Required|Description|
+|---|---|---|---|---|
+|calendar|any||*|the calendar to use such gregory, ethiopic, nepali - full supported list here: https://github.com/dhis2/multi-calendar-dates/blob/main/src/constants/calendars.ts|
+|cellSize|string|`'32px'`||the size of a single cell in the table forming the calendar|
+|date|string|||the current selected date in the format YYYY-MM-DD - if nothing provided, it will default to today|
+|defaultToToday|boolean|||whether to default to today's date or not|
```
tomzemp, line 20: isn't this no longer relevant? (also the text on `date` immediately above saying that `it will default to today`)?

docs/docs/components/calendar-input.md
```diff
@@ -0,0 +1,126 @@
+---
+title: Calendar Input
+---
+
+import { Demo } from '@site/src/components/DemoComponent.jsx'
```
tomzemp, line 5: I think these imports should go in the example? Also, it looks like the example is missing code block formatting?

components/calendar/src/calendar-input/calendar-input.js
```diff
@@ -76,14 +70,26 @@ export const CalendarInput = ({
...
                     label="Pick a date"
                     {...rest}
                     type="text"
                     onFocus={onFocus}
                     value={value}
                 />
+                {clearable && (
+                    <div className="calendar-clear-button">
+                        <Button
+                            dataTest="storybook-clear-button"
```
tomzemp, line 84: I'd recommend a name other than `storybook-clear-button` (as it's in the main component, rather than in a storybook story)

docs/docs/components/calendar-input.md
```diff
@@ -0,0 +1,126 @@
...
+        .wrapper > div {
+                /* min-width: 240px */
+        }
+    `}
+    </style>
+</Demo>
+
+## API Reference
+
+The component takes the same props as [the calendar](./calendar) component, as well as the props for [InputField](./inputfield).
```
tomzemp, line 120: I guess a couple of these might not be relevant (because we're restricting to  type=text, and some of the props like `step` only apply when type=number)? I guess it's just a thought that we might clarify the documentation at some point?

tomzemp (approved):
looks good to me! 🎉

I left some minor comments (mostly documentation related). I don't think you need to address them before merging (/some of them, maybe you don't need to address them at all 😀).

I'm a bit confused by Hendrik's last comment (`Our input components should be fully controlled`) as it seems like the CalendarInput is uncontrolled (e.g. the consuming component can pass a `date`, but then the CalendarInput can set a new date and it's up to consuming component to decide what to do)? I think the current code is good, but just had that question about what we meant in terms of this component needing to be controlled? (This is probably just my confusion on interpreting the comment).

components/calendar/src/calendar-input/calendar-input.js
```diff
@@ -76,14 +68,36 @@ export const CalendarInput = ({
...
+                    value={date}
                 />
+                {clearable && (
+                    <div
+                        className={cx('calendar-clear-button', {
+                            // ToDo: this is a workaround to show the clear button in the correct place when an icon is shown.
+                            // Long-term, we should abstract and share the logic multi-select uses for the input-wrapper
+                            // https://dhis2.atlassian.net/browse/DHIS2-14848
+                            'with-icon':
+                                rest.valid || rest.error || rest.warning,
```
HendrikThePendric, line 86: I don't see any practical use cases, but just for completeness sake I think it'd be good to add `rest.loading`, since a loading input will also be suffixed with an icon, see here: https://github.com/dhis2/ui/blob/d0c497911733c8185126dd2c43d6a2069040c342/components/input/src/input/input.js#L182-L187

HendrikThePendric (approved):
Awesome

docs/docs/components/calendar-input.md
tomzemp: Makes sense now (after you explained it to me more 😅)

PR #1243, opened 2023-03-01
fix: include .md extension for relative links

amcgee:
Relative links without the .md extension break when there is a trailing slash in the browser URL:

`https://ui.dhis2.nu/components/calendar-input` + `./calendar` => `https://ui.dhis2.nu/components/calendar` `https://ui.dhis2.nu/components/calendar-input/` + `./calendar` => `https://ui.dhis2.nu/components/calendar-input/calendar`

Docusaurus should handle this for us properly by resolving absolute link targets when the source file references another .md in the source (bonus: this should also work when browsing the docs source on github)

See https://github.com/facebook/docusaurus/issues/2298

Implements [JIRA_ISSUE_ID](https://dhis2.atlassian.net/browse/JIRA_ISSUE_ID)

---

### Description

_text_

---

### Known issues

-   [ ] _issue_

---

### Checklist

-   [ ] Tests were added

---

### Screenshots

_supporting text_

### DIFF
--- a/docs/docs/components/calendar-input.md
+++ b/docs/docs/components/calendar-input.md
@@ -117,7 +117,7 @@ Ethiopic calendar (with narrow day names, short day names and localised to Engli
 
 ## API Reference
 
-The component takes the same props as [the calendar](./calendar) component, as well as the props for [InputField](./inputfield) that are relevant to an input of type `text`.
+The component takes the same props as [the calendar](./calendar.md) component, as well as the props for [InputField](./inputfield.md) that are relevant to an input of type `text`.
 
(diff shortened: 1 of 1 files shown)


Topener:
Ah this is interesting, and honestly a bit weird it breaks like that. But good find. We should indeed check if anything else is broken but there's a LOT to dig through!

Topener:
Looks like one of the commits should be updated from `fix:` to `docs:`

Topener (approved):
I think it's okay, and it's hard to find any other links that might need adjusting. But maybe on of the commits needs to be changed to a `docs:` commit instead of the current one?

PR #1244, opened 2023-03-02
fix(calendar): design adjustments

cooper-joe:
### Description

This PR makes some minor design changes to the `Calendar` component:
- Use default cursor when hovering day name, months, and years labels.
- Adjust styling for previous and next navigation buttons
- Adjust styling of `Today` to prevent conflict with day hover style.
- Remove a couple of invalid `spacer` variables that weren't being applied and are not needed.

---

### Checklist

-   [x] Tests were added

### DIFF
--- a/components/calendar/src/calendar/calendar-table-cell.js
+++ b/components/calendar/src/calendar/calendar-table-cell.js
@@ -49,6 +49,9 @@ export const CalendarTableCell = ({ day, cellSize, selectedDate }) => {
                     background-color: ${dayHoverBackgroundColor};
                     text-decoration: underline;
                 }
+                button:active {
+                    background-color: ${colors.grey300};
+                }
                 button.isSelected,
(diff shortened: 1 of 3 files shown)


cooper-joe:
Not sure why the single test for `supports_calendar_clear_button` should fail (is expecting `02` and getting `2`), as I haven't made any changes to that. @kabaros, would you be able to offer guidance?

cooper-joe:
> I am wondering if the cursor when hovering over the days should be a pointer. This seems like what other popular calendar libraries do ([react-calendar](https://projects.wojtekmaj.pl/react-calendar/), [day picker](https://react-day-picker.js.org/) etc..) and it kind of makes sense to me.

Yes, thanks for the reminder! Agreed, have made the change.

tomzemp (approved):
lgtm 🚀

PR #1245, opened 2023-03-02
chore: update docusaurus version

Topener:
Updates the docusaurus version from an old beta to the most recent version


amcgee:
@Topener - if this is still valid let's update it and make sure the PR checks pass?

Topener:
@amcgee lets abandon this one, as it didn't go in yet and we're now working on moving it into the developer portal _and_ version 3 is already released.

PR #1246, opened 2023-03-03
fix: use connection status in headerbar [LIBS-315]

KaiVandivier:
Implements [LIBS-315](https://dhis2.atlassian.net/browse/LIBS-315)

Will also be used for [DHIS2-12937](https://dhis2.atlassian.net/browse/DHIS2-12937)

---

### Description

Switches from `useOnlineStatus` to `useDhis2ConnectionStatus`  in the headerbar status badge to detect connection to the DHIS2 server instead of the network status

---

### Checklist

-   [ ] Tests were added


HendrikThePendric (approved):
LGTM!

For some reason GitHub wouldn't let me place any inline comments today.... Luckily I only wanted to add one to `components/sharing-dialog/src/cascade-sharing/controls.js`, line 32-34:
> `refetch` should be a stable reference, so there should be no problem adding it to the dependency array and get rid of the ESLint warning.

components/header-bar/src/header-bar.js
```diff
@@ -113,9 +106,7 @@ export const HeaderBar = ({
                     )}
                 </div>
 
-                {(pwaEnabled || showOnlineStatus) && !loading && !error && (
-                    <OnlineStatus dense />
-                )}
+                {pwaEnabled && !loading && !error && <OnlineStatus dense />}
```
tomzemp, line 109: I don't understand what the point of this is 🤔? The logical condition seems to be the same as above and the `dense` property means that the css class used then includes `display:none`?

tomzemp (approved):
lgtm

components/header-bar/src/header-bar.js
KaiVandivier: It's part of making the `<OnlineStatus>` component reusable for the design in two different screen sizes:

Screenshot 2023-03-17 at 1 49 03 PM
Screenshot 2023-03-17 at 1 49 12 PM

The first `<OnlineStatus>` is the "badge" and shows on larger screens and is hidden on larger screens; the second `<OnlineStatus>` (with the `dense` prop) is a "bar" that's hidden on larger screens and shown on smaller screens: [effect of `dense` prop on container](https://github.com/dhis2/ui/blob/LIBS-315-connection-status-in-headerbar/components/header-bar/src/online-status.js#L20), and [CSS with media query](https://github.com/dhis2/ui/blob/7887cfa019e9b2cab71b955fea987c4f81980b0f/components/header-bar/src/online-status.styles.js#L14-L36)

tomzemp: Cool. Thanks. Missed the @media rule when reviewing 😅

PR #1251, opened 2023-04-19
feat: add dataTest prop to icons

HendrikThePendric:
After running `yarn build`, the generated components now look like this:

```jsx
import * as React from 'react'

function SvgAdd16({ color, dataTest }) {
    return (
        <svg
            height={16}
            viewBox="0 0 16 16"
            width={16}
            xmlns="http://www.w3.org/2000/svg"
            color={color}
            data-test={dataTest}
        >
            <path
                d="M8 2a1 1 0 011 1v4h4a1 1 0 010 2H9v4a1 1 0 01-2 0V9H3a1 1 0 110-2h4V3a1 1 0 011-1z"
                fill="currentColor"
                fillRule="evenodd"
            />
        </svg>
    )
}

SvgAdd16.propTypes = {
    color: PropTypes.string,
    dataTest: PropTypes.string,
}
export default SvgAdd16
```

### DIFF
--- a/icons/templates/icon-template.js
+++ b/icons/templates/icon-template.js
@@ -12,15 +12,16 @@ function template(
     const tpl = template.smart({ plugins })
 
     return tpl.ast`
-        ${imports}
         import PropTypes from 'prop-types'
+        ${imports}
 
-        function ${componentName}({ color }) {
+        function ${componentName}({ color, dataTest }) {
             return ${jsx};
         }
 
         ${componentName}.propTypes = {
             color: PropTypes.string,
+            dataTest: PropTypes.string,
         }
 
         ${exports}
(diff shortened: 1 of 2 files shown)


icons/templates/icon-template.js
```diff
@@ -12,15 +12,16 @@ function template(
     const tpl = template.smart({ plugins })
 
     return tpl.ast`
-        ${imports}
         import PropTypes from 'prop-types'
+        ${imports}
```
HendrikThePendric, line 16: This is an unrelated change to appease the linter. `React` was imported before `PropTypes` which caused a import order ESLint error. Since these are generated files, probably nobody cares, but at least I will sleep better at night knowing that our generated files adhere to our ESLint rules.

Mohammer5 (approved):

PR #1253, opened 2023-04-20
fix: translate validation text inside validation functions

tomzemp:
Implements [LIBS-483](https://dhis2.atlassian.net/jira/software/c/projects/LIBS/issues/LIBS-483)

---

### Description

Currently the translation strings in form validators are defined outside of the exported validator function. This means that the translation is a constant, generated when the module gets imported. This works generally as users have to change their selected language and then navigate to another app, but this does not work in the context of the login app where the language selection is on the page, and we do not want to have to force refresh in order for translations to flow through.

---

### Known issues

-   [ ] With regards to the implementation in the login app: the validation strings returned from the validators will now update language once the validation is rerun (e.g. this occurs on field toggling with react final form validation) (so the translation is still not as instantaneous as some of the other on page translations)
- 

---

### Checklist

-   [ ] Tests were added

_The changes here do not change any functionality, so just updating the existing tests to pass seems sufficient._

---

### Screenshots
translations_validations


Mohammer5:
> do you think this could have a visible performance implication, having to call i18n.t several times rather than once?

I don't think that this should be an issue, these are primitive functions with little computation, just grabbing values by key from an object.

If you're very worried, I suppose the code could be changed so that the translation function is only called when the translation is needed:

Before:
```js
const boolean = (value) => {
    const invalidBooleanMessage = i18n.t('Please provide a boolean value')

    return isEmpty(value) || typeof value === 'boolean'
        ? undefined
        : invalidBooleanMessage
}
```

After:
```js
const boolean = (value) => {
    if (isEmpty(value) || typeof value === 'boolean') {
        return undefined
    }

    return i18n.t('Please provide a boolean value')
}
```

Mohammer5:
> With regards to the implementation in the login app: the validation strings returned from the validators will now update language once the validation is rerun (e.g. this occurs on field toggling with react final form validation) (so the translation is still not as instantaneous as some of the other on page translations)

I don't think that this should block this PR from being merged as this seems to be an edge case. A simple fix would be to increment a `key` prop when changing the language to force a rerender (not entirely sure that'd work, but it should be very straight forward to implement)

Mohammer5:
> it does a bit more than [key lookup](https://github.com/i18next/i18next/blob/v10.3.0/i18next.js#L597) but none of it looks expensive to be fair, also none of it seems to have changed in the 5 years since we introduced that version of the library.

Ah yeah, fair enough. I didn't expect that much to be going on.. 

> I guess my point is that if re-initialising i18next solves the login app use case, then maybe we can just do that (when the user changes language in the login screen) rather than changing the UI components here. I am not sure if that's feasible though with the wrapper we have around i18next - in that case, this solution looks good.

Yeah, worth a try I guess

Mohammer5:
@tomzemp I assume @kabaros meant calling `i18n.init()` again with the right `lng`

tomzemp:
FYI. I updated logic to move the translation until after the error was identified (as suggested by @Mohammer5). 

I will put this PR on hold, though, until we decide the overall approach for login app customization, as that may affect how we handle language selection. If we use web components for customization, the app may need to be refreshed when changing language, which would make this PR unnecessary.

PR #1301, opened 2023-05-19
fix(headerbar): improve clickable area interactions

cooper-joe:
Implements [UX-142](https://dhis2.atlassian.net/browse/UX-142)

---

### Description

Improves interaction with clickable elements of the headerbar:
- increase click area of interactive elements
- give interactive elements a `:hover` style
- use consistent `:focus` style

---

### Checklist

-   [x] Tests were added

---

### Screenshots

**Before:**

https://github.com/dhis2/ui/assets/33054985/dd10dde4-538f-49f0-9dee-bfe95838022a

**After:**

https://github.com/dhis2/ui/assets/33054985/3270e794-16a5-4507-ba43-efb6720168be

[UX-142]: https://dhis2.atlassian.net/browse/UX-142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/header-bar/src/apps.js
+++ b/components/header-bar/src/apps.js
@@ -1,7 +1,7 @@
 import { Card } from '@dhis2-ui/card'
 import { InputField } from '@dhis2-ui/input'
 import { useConfig } from '@dhis2/app-runtime'
-import { colors, theme } from '@dhis2/ui-constants'
+import { colors, spacers, theme } from '@dhis2/ui-constants'
 import { IconApps24, IconSettings24 } from '@dhis2/ui-icons'
 import PropTypes from 'prop-types'
 import React, { useState, useEffect, useCallback, useRef } from 'react'
@@ -186,9 +186,7 @@ const AppMenu = ({ apps, filter, onFilterChange }) => (
             div {
                 z-index: 10000;
                 position: absolute;
-                top: 28px;
-                right: -6px;
-                border-top: 4px solid transparent;
+                right: -4px;
             }
         `}</style>
     </div>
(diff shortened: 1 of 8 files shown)


KaiVandivier (approved):
Looks good! These are nice improvements :)

I'm curious if there are other options for the focus styles -- the 3px white inset border works and is probably good for contrast, but I find it a little clunky-looking

I tried a few other white outline options and didn't find anything that I like, but what about a _darker_ blue background for `:focus` and `:active` styles? I looked to Gmail to see what they do, and they actually do the same thing

I tried out what that would look like on the profile button:

https://github.com/dhis2/ui/assets/49666798/99f20dbd-2347-4aca-a162-0c8480e3ba6b

Approved because it's good as is, but let me know what you think 🙂

cooper-joe:
> I'm curious if there are other options for the focus styles -- the 3px white inset border works and is probably good for contrast, but I find it a little clunky-looking

Thanks for the review @KaiVandivier. The focus styles are intended for keyboard navigation and accessibility, so keeping them easy to see is a priority. I've reduced them down to `2px` outline to be more consistent with the rest of our components, as well as using `:focus:not(:focus-visible)` to avoid showing the focus style on a mouse-click, so this should help prevent any clunkiness.

Mohammer5 (approved):

KaiVandivier:
Nice @cooper-joe, I think that looks great 😀

PR #1320, opened 2023-05-25
feat: singleselect, noticebox rtl [LIBS-502]

tomzemp:
Implements [LIBS-502](https://dhis2.atlassian.net/jira/software/c/projects/LIBS/issues/LIBS-502)

---

### Description

Updates NoticeBox and SingleSelect components to use logical properties and hence preserve styling when used within an RTL environment.

I've updated from physical to logical properties. I wasn't quite sure how much to change; I ended up changing things like `padding: 4px 8px;` to `padding-block: 4px; padding-inline: 8px;`. Since for RTL-support we are interested in preserving the styling when text flow switches from left to right, it isn't really necessary to update to logical properties (for example for padding) if the left/right value is the same (i.e. we could just leave as `padding: 4px 8px;` because in terms of RTL support, the physical left-right padding here will be equivalent to logical inline padding) ; however, I thought it would be clearer to generally change to logical properties.

That said, I did not switch absolutely everything to logical properties (for example, I did not update `width` to `inline-size`). I thought it might be more confusing to use logical properties here. This is another case, where the physical `width/height` properties are sufficient for RTL support but  would not be sufficient if we wanted our components to be more responsive for `writing-mode: vertical-rl;` layouts*.

(*Mongolian is our only top-to-bottom UI language (depending on choice of script), but we do not have any translations into Mongolian, and the language is generally written in Cyrillic not traditional Mongolian script, so it seems like we do not want to spend time thinking about top-to-bottom layout).

---

### Checklist

-   [ ] Tests were added

- API docs: I'm not sure we want to update anything at the moment in our design documents to reference RTL support? The current situation is that some components will look correct in RTL (e.g. input) and others (NoticeBox) will not. Personally, I think maybe we would want to mention RTL support somewhere generally in the documents once we've updated most components?
- Tests: I was thinking about this, but am unsure how we might want to do testing for this. Our tests generally do not test for layout/design specifics, so they should work regardless of whether the layout is LTR/RTL (including for components where layout is slightly off in RTL). If we want RTL testing, maybe it would want to use snapshot testing (though I'm not sure how useful that would end up being 🤔) ?
---

### Screenshots

NoticeBox
image

SingleSelect
image

[LIBS-502]: https://dhis2.atlassian.net/browse/LIBS-502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KaiVandivier (approved):
Nice, looks good 🙂

PR #1329, opened 2023-06-08
translation: add translation for selector-bar

Birkbjo:
Adds `selector-bar` as a resource.

---

### Description

Would it be possible to do this so all components are included, instead of having to manually add each component to the config?

---

### Checklist

-   [ ] Tests were added


eirikhaugstulen (approved):
Looks good 👍

kabaros (approved):

PR #1366, opened 2023-06-28
fix: update input-type component styles

cooper-joe:
### Description

This PR updates styles for input-type components. Existing styles were inconsistent or buggy. 

Overview of changes:
- `Input` was 1px taller in Firefox at both `regular` and `dense` sizes. I've resorted to `max-height` to fix this, as nothing else worked.
    - I can't imagine using `max-height` being problematic because this is always a single line input, but just want to check this is OK?
- `SingleSelect` and `MultiSelect` `dense` sizes changed from `34px` to the correct `32px`.
- Adjusted some `box-shadow`s.
- `Textarea` padding, lineheight updated.
- Added some storybook demos for easier testing.

---

### Checklist

-   [x] Tests were added

### DIFF
--- a/components/select/src/multi-select/multi-select.stories.js
+++ b/components/select/src/multi-select/multi-select.stories.js
@@ -135,6 +135,14 @@ WithInitialFocus.args = { initialFocus: true }
 WithInitialFocus.parameters = { docs: { disable: true } }
 WithInitialFocus.storyName = 'With initialFocus'
 
+export const Dense = WithOptionsTemplate.bind({})
+Dense.args = { dense: true, placeholder: 'Dense sized multi select' }
+Dense.storyName = 'Dense'
+
+export const DenseWithSelections = WithOptionsTemplate.bind({})
+DenseWithSelections.args = { dense: true, selected: ['1', '2'] }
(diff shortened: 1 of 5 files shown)


components/select/src/multi-select/multi-select.stories.js
```diff
@@ -135,6 +135,14 @@ WithInitialFocus.args = { initialFocus: true }
 WithInitialFocus.parameters = { docs: { disable: true } }
 WithInitialFocus.storyName = 'With initialFocus'
 
+export const Dense = WithOptionsTemplate.bind({})
+Dense.args = { dense: true, placeholder: 'Dense sized mutli select' }
```
tomzemp, line 139: typo on `multi`

tomzemp (approved):
Looks good to me.

The only issue I could think of with max-height on the input would be if someone was using `type: image` on the input, but I'm not sure if that even works with our Input component, and it wouldn't make sense to use it in this way in any case.

components/select/src/multi-select/multi-select.stories.js
cooper-joe: thanks, fixed!

Topener (approved):
LGTM

PR #1386, opened 2023-07-13
fix(transfer): do not assume selected can be mapped

ismay:
This fixes a catch-22 I'm encountering with the transfer in combination with final-form. The transfer expects `selected` to be an array (i.e. not undefined). I think it would be good to make it more robust and not assume that selected is passed.

To fix the above in an app it's not too much trouble to supply an `initialValue` for a field with final-form, which is what I did initially. The problem is that an initialValue on a final-form field overrides the form's `initialValues` property.

Passing initialValues to a form is how you'd commonly seed a form with values if it's an edit form. So supplying an initialValue to the transfer based field fixes the transfer crashing if selected is not present, but it breaks the initialValues workflow for seeding an entire form with server-side values.

Now I could continue fixing this in the app with more convoluted solutions, but I thought I'd fix it at the source.

### DIFF
--- a/components/transfer/src/transfer.js
+++ b/components/transfer/src/transfer.js
@@ -138,14 +138,21 @@ export const Transfer = ({
      * Extract the selected options. Can't use `options.filter`
      * because we need to keep the order of `selected`
      */
-    const pickedOptions = actualFilterPickedCallback(
-        selected
-            .map((value) => options.find((option) => value === option.value))
-            // filter -> in case a selected value has been provided
-            // that does not exist as option
-            .filter(identity),
-        actualFilterPicked
-    )
+    let pickedOptions = []
+
+    // Only map if selected is an array
+    if (Array.isArray(selected)) {
+        pickedOptions = actualFilterPickedCallback(
+            selected
+                .map((value) =>
+                    options.find((option) => value === option.value)
+                )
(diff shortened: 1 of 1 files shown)


KaiVandivier (approved):
This seems fine -- is there anywhere else in the component it needs to handle a non-array value?

Just a suggestion -- I think the code would be a bit more self-descriptive like this:
```js
let pickedOptions = []
// Avoid breaking for non-array values, e.g. for Final Form initial values
if (Array.isArray(selected)) {
    pickedOptions = selected.map(/* ... */)
    // ...
}
```
It's a bit more restrictive for data type, but I doubt we're really supporting anything other than arrays here

(You could also use a conditional assignment too, like `pickedOptions = Array.isArray(selected) ? selected.map(/* ... */) : []`)

ismay:
> This seems fine +1 slightly_smiling_face is there anywhere else in the component it needs to handle a non-array value?

Not that I know of. I mostly went through the component looking for the source of the error, so I'm not sure if this might be an issue elsewhere. It's a fairly complex component so I can imagine there are. Though `selected` is one of the more obvious candidates.

I agree that your suggestion is clearer. I went back and forth with approaches on this, and settled on duck typing because the map is what we're using, but in hindsight just making sure that it's an array seems the clearest. I'll update it.

PR #1400, opened 2023-09-07
fix: select-field not showing in correct place

kabaros:
Fixes https://dhis2.atlassian.net/browse/LIBS-532

This fix involves disabling portal for the Select Menu Wrapper. 

###  Other explored solutions
I looked mainly at ensuring the ref is updated before when the parent is show, by using a callback ref (similar to [this](https://react.dev/reference/react-dom/components/common#ref-callback) and [this](https://legacy.reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node)). But there were complications, the first is that this technique works only with functional componet, I tried the class-based version of it but that didn't work. I tried converting the component into a functional one, and quickly got into a rabbit hole.

## Screenshot

https://github.com/dhis2/ui/assets/1014725/e9d75595-015d-4a20-83c6-6352294da4d1

### DIFF
--- a/components/select/src/select/menu-wrapper.js
+++ b/components/select/src/select/menu-wrapper.js
@@ -13,7 +13,7 @@ const MenuWrapper = ({
     selectRef,
 }) => {
     return (
-        <Layer onBackdropClick={onClick} transparent>
+        <Layer onBackdropClick={onClick} transparent disablePortal>
             <Popper
                 reference={selectRef}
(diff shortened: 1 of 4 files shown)


KaiVandivier (approved):
Tested in the reproduction app Tom made in the Jira issue 👍 

I can't think of any side effects of the non-portal layer 🤔

PR #1423, opened 2023-11-15
fix: types exports

Birkbjo:
This also updates alpha from master. 

This seems to be needed when to support various bundlers with `moduleResolution: 'bundler'` (in `tsConfig`).


Mohammer5 (approved):
LGTM

PR #1425, opened 2023-11-16
feat(header-bar): reduce height and minor design changes

cooper-joe:
Implements [LIBS-545](https://dhis2.atlassian.net/browse/LIBS-545) / [UX-144](https://dhis2.atlassian.net/browse/UX-144)

---

### Description

This PR makes visual changes to the `HeaderBar` component:
- reduce height to 40px
    - inner components are adjusted to compensate for the height reduction
- minor adjustments to base color and interaction colors (in order to improve readability of the slightly smaller text size)

---

### Checklist

-   [x] Tests were added (Not sure why the 2 tests are failing?)

---

### Screenshots

Before:
image

After:
image

[LIBS-545]: https://dhis2.atlassian.net/browse/LIBS-545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[UX-144]: https://dhis2.atlassian.net/browse/UX-144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amcgee (approved):
Tests should be fixed now - change looks good to me!  This *could* be considered a breaking change depending on how people are using it (absolute positioning offsets) but I don't think we need to be strict about that.  The alternative would be to do this change only in the global shell.. but not sure that's necessary

cooper-joe:
> The alternative would be to do this change only in the global shell

@amcgee and I discussed this further. This change will be deferred until the `HeaderBar` is moved to the global shell.

KaiVandivier:
Added to global shell: https://github.com/dhis2/global-shell-app/pull/8

PR #1427, opened 2023-11-21
feat: menu item suffix

cooper-joe:
Implements [LIBS-536](https://dhis2.atlassian.net/browse/LIBS-536)

---

### Description

This PR adds the optional `suffix` prop to `<MenuItem>`. A `suffix` is displayed at the end of the component. A `suffix` can be used alongside `chevron: true`.

---

### Checklist

-   [x] Tests were added

---

### Screenshots

image

image

[LIBS-536]: https://dhis2.atlassian.net/browse/LIBS-536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/menu/src/menu-item/features/accepts_suffix.feature
+++ b/components/menu/src/menu-item/features/accepts_suffix.feature
@@ -1,0 +1,5 @@
+Feature: The MenuItem accepts a suffix prop
+
+    Scenario: MenuItem renders supplied suffix
+        Given a MenuItem supplied with a suffix is rendered
+        Then the suffix will be visible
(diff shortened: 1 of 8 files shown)


Mohammer5 (changes requested):
From a functional perspective this looks good! Could you add a story with both the chevron and the suffix? That way it's clear that both can be used together and we can see what it looks like, too

Mohammer5 (approved):

Birkbjo (approved):
Nice! LGTM

PR #1435, opened 2023-11-28
fix(chip): add tabIndex prop

Birkbjo:
---

### Description

Adds `tabIndex` to chip component. 

I followed the pattern for other tabIndex, but I wonder why this is a `string` and not `number`? Also in my opinion these tabindex should default to `0` - but they don't? So I decided to keep it aligned with other components. 

---

### Known issues

-   [ ] _issue_

---

### Checklist

-   [ ] Tests were added

---

### DIFF
--- a/components/chip/src/chip.js
+++ b/components/chip/src/chip.js
@@ -22,6 +22,7 @@ const Chip = ({
     marginLeft,
     marginRight,
     marginTop,
+    tabIndex,
 }) => (
     <span
         onClick={(e) => {
@@ -36,6 +37,7 @@ const Chip = ({
             dragging,
         })}
         data-test={dataTest}
+        tabIndex={tabIndex}
     >
         <Icon icon={icon} dataTest={`${dataTest}-icon`} />
         <Content overflow={overflow}>{children}</Content>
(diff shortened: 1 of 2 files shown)


KaiVandivier (approved):

cooper-joe (approved):
Nice 👍 

This allows keyboard focus for the main chip body, but can we propagate that down to the `<Remove>` child component for `removable` type chips? Potentially out of scope here, so approved without this.

Birkbjo:
@cooper-joe thanks for your quick import. After testing this in storybook, it seems like this is not enough to actually make `tabIndex` useful. Because it's a span an-`enter`-press will not fire `onClick`. So we would have to handle the keypresses manually, or refactor this to be a button.

The only problem of making this a button is if you wanted the chip to be a link - link wrapping a button is not valid HTML.

cooper-joe:
I would prefer this to be a button rather than hack together our own keypress handling. Alas, wrapping the chip in the link is [exactly how it's being used](https://github.com/dhis2/dashboard-app/blob/1c99cc2a9e447f7fe9fd81d4e689cd3b97c41677/src/components/DashboardsBar/Chip.js#L38-L61), so looks like we may not have a choice unless we're prepared to release the breaking change. I'm open to the the breaking change, if it means a step towards better accessibility.

PR #1437, opened 2023-12-06
fix(select): adjust clear button

cooper-joe:
Implements [UX-154](https://dhis2.atlassian.net/browse/UX-154)

---

### Description

This PR changes the `SingleSelect` and `MultiSelect` clear button to use an icon button rather than a `Button` with text label.

`clearText` prop is used to populate the `Tooltip` that appears on hover.

---

### Checklist

-   [x] Tests were added

---

### Screenshots

Before: 
image
image

After:
image
image

[UX-154]: https://dhis2.atlassian.net/browse/UX-154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/select/src/single-select/input.js
+++ b/components/select/src/single-select/input.js
@@ -24,7 +24,7 @@ const Input = ({
     inputMaxHeight,
 }) => {
     const hasSelection = selected && typeof selected === 'string'
-    const onClear = (_, e) => {
+    const onClear = (e) => {
         const data = { selected: '' }
 
         e.stopPropagation()
@@ -72,7 +72,6 @@ const Input = ({
 
                 .root-right {
                     margin-inline-start: auto;
-                    margin-inline-end: 10px;
                 }
             `}</style>
 
--- a/components/select/package.json
+++ b/components/select/package.json
@@ -33,7 +33,6 @@
         "styled-jsx": "^4"
     },
     "dependencies": {
-        "@dhis2/prop-types": "^3.1.2",
         "@dhis2-ui/box": "9.4.2",
         "@dhis2-ui/button": "9.4.2",
         "@dhis2-ui/card": "9.4.2",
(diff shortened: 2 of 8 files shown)


components/select/src/single-select/input.js
```diff
@@ -24,7 +24,7 @@ const Input = ({
     inputMaxHeight,
 }) => {
     const hasSelection = selected && typeof selected === 'string'
-    const onClear = (_, e) => {
+    const onClear = (e) => {
```
Mohammer5, line 27: This doesn't appear to be correct, see here: https://github.com/dhis2/ui/blob/4643200c1adfdd40223e0b4d04b144b7d9a25a4a/components/button/src/button/button.js#L39

* Callback is passed to `InputClearButton`
* `InputClearButton` passes the callback to `Button`
* `Button` calls the onClick callback with the event as second argument

Mohammer5 (approved):
~~This is actually a change request, but I won't be working next week, so approving so I don't block this PR~~

EDIT: :+1:

components/select/src/single-select/input.js
tomzemp: @Mohammer5: won't this be okay though in the sense that now @cooper-joe is using the native <button> element rather than our <Button> component in <InputClearButton> ? We don't seem to allow users to pass their own `onClear` down, so this seems like it would be fine to me?

tomzemp (approved):

components/select/src/single-select/input.js
Mohammer5: I can't follow. What I can see, is:

* The select's input component defines the `onClear` function, which is called when the user clicks the clear button.
* The `onClear` function will call the `onChange` callback provided by the user
* With the currently proposed implementation, the second parameter wouldn't be an event

cooper-joe: I was under the impression that this was OK, as @tomzemp says the button is now a `<buttton>` rather than a `<Button>`.

Mohammer5: Ah yes, my bad! When I used GH to go to the definition, it sent me to the master branch... Sorry!

PR #1448, opened 2024-02-07
feat: rtl support for components [LIBS-525]

tomzemp:
Implements [LIBS-525](https://dhis2.atlassian.net/browse/LIBS-525)

---

### Description

This PR adds RTL support to our existing components. This involves:
- updates to use logical properties (e.g. `padding-inline-start` instead of `padding-inline-end`)
- some additional changes (e.g. icon orientation) handled with css pseudoclasses (`:dir(rtl)`) (e.g. to transform arrows in transfer component) to make sense with dir=rtl
- some direct remappings from `left` to `right` (and vice versa) where this is provided (e.g. in Tables)
- some remapping of cases where we have allowed for non-logical property css (e.g. remapping `margin` in divider if dir=rtl)

**Note on approach:** I've made these changes manually rather than using a code mode, in part because the code mods I saw seemed to mostly be focused on css files where we have styling as jsx for most components. Also, making manual changes allowed more control over the actual changes (see points below). Finally, there were a number of issues that would not be addressed by updating logical properties.

**Note on logical properties:** For the most part I have only updated to logical properties to convert left/right to inline. This simplifies things a bit as some existing shorthand notation does not need updating (e.g. `padding: 4px 8px` has equal values on left/right, so doesn't _need_ to be converted to logical property). I opted for mostly minimally intrusive changes, so I left valid values that did not cause issues.

I didn't think it was worth it to convert top/bottom to block logical properties because the changes needed to support tb text is much more comprehensive (and involves updating css with changes like `width` to `inline-size` and also setting `writing-mode` (which we're not doing)). There's only one major script (traditional Mongolian) that is written exclusively TB, and DHIS2 is only used by NGOs in Mongolia; also the Mongolian language can be written in Cyrillic (which seems to be the more common script). So, in general I've just kept top/bottom references and only switched to block in a couple cases where I was already breaking out something like `margin: 1px 5px 10px 2px`.

---

### Known issues

-   [ ] Switch component does not display checkmark in RTL rotation

LTR:
image

RTL:
image

This is because I've applied a transformation on the entire switch icon, which would also rotate the checkmark (undesired), so I have defaulted to hiding it until we make more comprehensive update to Switch

-   [ ] Calendar and CalendarInput were not updated because of errors preventing their display, but I believe these were already RTL compliant?
---

### Checklist

I've generally not updated API docs because I haven't changed the API. I think we want people to generally assume that the component will display correctly in RTL? I have updated the chip component documentation because I have added `marginInlineStart` and `marginInlineEnd` to complement `marginLeft` and `marginRight` props we already had
-   [ ] Tests were added
We discussed this before in context of previous updates and decided that adding tests for RTL was not necessary (at least for logical properties updates). If we think there should be tests for non-logical-property changes, I could add some?
I've added a demo (generally called RTL) for components that will look different in RTL (i.e. most of them). I've added a story even in a few cases where there are no actual updates to the component itself (e.g. Linear Loader) just to illustrate how things will look.

Note because of the portal, I sometimes need to use useEffect within the storybook demos to set the document direction (with app-platform the portal direction should also be set, so this is strictly a step for making the storybook demos behave appropriately)

---

### Screenshots

There are a lot of changes, so probably best to run the storybook locally and review the RTL stories.

[LIBS-525]: https://dhis2.atlassian.net/browse/LIBS-525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/divider/src/divider.js
```diff
@@ -2,23 +2,41 @@ import { colors, spacers } from '@dhis2/ui-constants'
...
-            }
-        `}</style>
-        <style jsx>{`
-            div {
-                margin: ${dense ? `${spacers.dp4} 0` : margin};
-            }
-        `}</style>
-    </div>
-)
+const flipMargin = (margin) => {
```
tomzemp, line 5: I'm not totally sure if we want to try to remap margin provided as a prop, but this seemed probably more helpful to attempt?

components/chip/API.md
```diff
@@ -22,8 +22,10 @@ import { Chip } from '@dhis2-ui/chip'
 |dragging|boolean||||
 |icon|element||||
 |marginBottom|number|`4`||`margin-bottom` value, applied in `px`|
-|marginLeft|number|`4`||`margin-left` value, applied in `px`|
-|marginRight|number|`4`||`margin-right` value, applied in `px`|
+|marginInlineStart|number|`4`||`margin-inline-start` value, applied in `px`|
```
tomzemp, line 25: I thought we would want to leave the existing `marginLeft` and `marginRight` properties and not treat them as equivalent to inline-start inline-end, but that going forward it would be better to also provide for people to enter the margin in a logical-property-compliant manner (hence addition of `marginInlineStart` and `marginInlineEnd` here)

components/popper/src/popper.js
```diff
@@ -32,7 +42,10 @@ const Popper = ({
     const { styles, attributes } = usePopper(referenceElement, popperElement, {
         strategy,
         onFirstUpdate,
-        placement,
+        placement:
+            document.documentElement.dir === 'rtl'
```
tomzemp, line 46: I don't know what the ideal is here? The [popper](https://popper.js.org/) library that we're using says that it handles RTL, but it seems to provide placement only with physical properties. I assume we want to switch these, so I have done so based on the document.documentElement direction that is being set by app-platform, but I'm not sure if there's a reason why we might want to strictly interpret the placement as a physical property?

components/table/src/data-table/table-elements/table-data-cell/table-data-cell.js
```diff
@@ -5,6 +5,12 @@ import PropTypes from 'prop-types'
 import React, { forwardRef } from 'react'
 import styles from './table-data-cell.styles.js'
 
+const rtlCorrespondingAlignments = {
```
tomzemp, line 8: here it also seemed like we would want to assume that the physical properties we have on this component for alignment should actually behave as logical properties. Though I guess we could also add new options like `inline-start`, `inline-end` for alignment?

components/table/src/data-table/table-elements/table-data-cell/table-data-cell.js
```diff
@@ -30,50 +36,56 @@ export const TableDataCell = forwardRef(
...
-                    background-color: ${backgroundColor || colors.grey200};
-                }
-                :global(tr.selected:hover) > td:not(.staticStyle) {
-                    background-color: ${backgroundColor || '#cdeae8'};
-                }
-            `}</style>
-        </td>
-    )
+    ) => {
+        const rtlAlign = rtlCorrespondingAlignments[align]
```
kabaros, line 40: shouldn't be the case, but let's fallback to what they pass if no match `const rtlAlign = rtlCorrespondingAlignments[align] ?? align`

kabaros (approved):
great stuff @tomzemp  💯 - some minor comments, but let's merge it asap (at least to alpha) and see how it works in an app

PR #1449, opened 2024-02-09
feat: add aria attributes to loaders

d-rita:
Implements [560](https://dhis2.atlassian.net/browse/LIBS-560)

---

### Description

This feature improves the accessibility of the loaders by adding aria attributes to them, i.e.
- the `aria-valuenow` and `aria-label` attributes to the linear loader
- the `aria-label` attribute to the circular loader

---

### Checklist
-   [x] Tests were added

### DIFF
--- a/components/loader/src/circular-loader/__tests__/circular-loader.test.js
+++ b/components/loader/src/circular-loader/__tests__/circular-loader.test.js
@@ -1,0 +1,26 @@
+import { mount } from 'enzyme'
+import React from 'react'
+import { CircularLoader } from '../circular-loader.js'
+
+describe('Circular Loader', () => {
+    it('renders the circular loader with aria label', () => {
+        const wrapper = mount(
+            <CircularLoader
+                dataTest={'circular-loader-test'}
+                ariaLabel="Circular Loader"
+            />
+        )
+        const actual = wrapper.find({ 'data-test': 'circular-loader-test' })
+        expect(actual.prop('role')).toBe('progressbar')
+        expect(actual.prop('aria-label')).toBe('Circular Loader')
+    })
+
+    it('renders the circular loader without aria label', () => {
+        const wrapper = mount(
+            <CircularLoader dataTest={'circular-loader-test'} />
+        )
+        const actual = wrapper.find({ 'data-test': 'circular-loader-test' })
+        expect(actual.prop('aria-label')).toBe(undefined)
+        expect(actual.prop('role')).toBe('progressbar')
+    })
+})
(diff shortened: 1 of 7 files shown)


components/loader/src/circular-loader/circular-loader.js
```diff
@@ -67,6 +69,7 @@ CircularLoader.defaultProps = {
 }
 
 CircularLoader.propTypes = {
+    ariaLabel: PropTypes.string,
```
kabaros, line 72: @cooper-joe should the `ariaLabel` be required here?

kabaros (approved):
LGTM - just the one question for Joe to answer

components/loader/src/circular-loader/circular-loader.js
cooper-joe: No, I don't think so. As I understood the requirements, an accessible label can be provided alongside the loader in a container. So in that case, `ariaLabel` isn't technically required. And for the purposes of adoption, I'd prefer to throw lint or console warnings for these violations rather than prop errors.

components/loader/src/circular-loader/__tests__/circular-loader.test.js
```diff
@@ -0,0 +1,26 @@
+import { mount } from 'enzyme'
+import React from 'react'
+import { CircularLoader } from '../circular-loader.js'
+
+describe('Circular Loader', () => {
+    it('renders the circular loader with aria label', () => {
+        const wrapper = mount(
+            <CircularLoader
+                dataTest={'circular-loader-test'}
+                ariaLabel="Circular Loader"
```
kabaros, line 10: the aria-label should actually not be camelCased .. just `aria-label`. Here is a snippet from the docs:

```
In React, all DOM properties and attributes (including event handlers) should be camelCased. For example, the HTML attribute tabindex corresponds to the attribute tabIndex in React. The exception is aria-* and data-* attributes, which should be lowercased. For example, you can keep aria-label as aria-label.
```
https://legacy.reactjs.org/docs/dom-elements.html

kabaros (changes requested):

components/loader/src/circular-loader/__tests__/circular-loader.test.js
d-rita: Noted. Will update

d-rita: @kabaros the props argument for the label is the one that is camelCased. The implementation itself it as `aria-label`. For example: https://github.com/dhis2/ui/blob/ac2b17b5cdbf8941f520a15e7c97a08b554ba6ce/components/loader/src/circular-loader/circular-loader.js#L24 
Should I still make the change?

kabaros (approved):

kabaros (approved):

PR #1451, opened 2024-02-12
feat: update accessibility in button component

Chisomchima:
feat: update accessibility in button component 

---

### Description

Implemented [JIRA_ISSUE_ID](https://dhis2.atlassian.net/browse/LIBS-565) by updating the Button component to handle `title` and `aria-label` attributes passed as part of `otherProps`. This change ensures that accessibility attributes can be easily included when using the Button component.

---

### Known issues

-   [x] None

---

### Checklist

-   [x] Tests were added

---

### Screenshots
- None

---

### Additional Notes 
- Updated the `Button` component to destructure `title` and `ariaLabel` from `otherProps` and apply them to the `<button>` element.
- Ensured that `title` and `aria-label` attributes are correctly passed to the rendered `<button>` element.
- Modified the component to handle additional attributes passed as `otherProps` while maintaining accessibility.

_supporting text_


cooper-joe (approved):
Looks good to me 👍 

As an aside, as mentioned on the ticket, I think it would be nice to introduce some soft warning or linting if there is no button label **and** there is no aria-label. However, we may want to address this as a library wide linting strategy, so deferring for now.

PR #1455, opened 2024-02-16
chore: update warning in button component

Chisomchima:
Implements [LIBS-565](https://dhis2.atlassian.net/browse/LIBS-565)

---
### Description

Implemented [LIBS-565](https://dhis2.atlassian.net/browse/LIBS-565) by updating developer warning in button component. This change ensures that accessibility attributes can be easily included when using the Button component. 

---

### Known issues

-   [ ] None

---

### Checklist

-   [ ] Tests were added

---

### Screenshots
- None

---

### Additional Notes 
- Updated the `Button` component to destructure `title` and `ariaLabel` from `otherProps` and apply them to the `<button>` element.
- Ensured that `title` and `aria-label` attributes are correctly passed to the rendered `<button>` element.
- Modified the component to handle additional attributes passed as `otherProps` while maintaining accessibility.

_supporting text_

[LIBS-565]: https://dhis2.atlassian.net/browse/LIBS-565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIBS-565]: https://dhis2.atlassian.net/browse/LIBS-565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/button/src/button/button.js
```diff
@@ -37,6 +37,14 @@ export const Button = ({
         }
     }, [initialFocus, ref.current])
 
+    const { 'aria-label': ariaLabel, title } = otherProps
+
+    if (!children && !title && !ariaLabel) {
+        console.debug(
+            'Button component has no children but is missing title or ariaLabel attribute.'
```
Birkbjo, line 44: Log statement here does not match the if-condition above. Here you need both `title` and `ariaLabel` to be present.

Do we want to `ariaLabel` to fallback to `title`?

components/button/src/button/__tests__/Button.test.js
```diff
@@ -4,6 +4,52 @@ import React from 'react'
...
+    })
+
+    describe('warning for missing aria-label or title', () => {
+        it('No warning  if children exist but aria-label or title is missing', () => {
+            render(<Button>Children content</Button>)
+
+            expect(consoleSpy).not.toHaveBeenCalled()
+        })
+
+        it('does not warn if aria-label or title is present', () => {
```
Birkbjo, line 24: This "or" should be "and" according to the test and implementation, right?

components/button/src/button/button.js
```diff
@@ -37,6 +37,14 @@ export const Button = ({
         }
     }, [initialFocus, ref.current])
 
+    const { 'aria-label': ariaLabel, title } = otherProps
```
kabaros, line 40: not part of this PR - but I think `ariaLabel` in the previous tests should be updated to `aria-label` like in [this test](https://github.com/dhis2/ui/pull/1451/files#diff-2fdc96d318bd3f7fe855ac745e982336111185c7fd55c50a78151bd533d10c12R30)

kabaros: yeah we should make sure these are `aria-label` (not camelcased). This applies to your previous PR as well @d-rita  - let's update them to make them consistent `aria-*` 

```
In React, all DOM properties and attributes (including event handlers) should be camelCased. For example, the HTML attribute tabindex corresponds to the attribute tabIndex in React. The exception is aria-* and data-* attributes, which should be lowercased. For example, you can keep aria-label as aria-label.

```
https://legacy.reactjs.org/docs/dom-elements.html

components/button/src/button/__tests__/Button.test.js
Chisomchima: yeah, you're right, "and" seems like a better choice in this case.

kabaros (approved):
approved - will merge to alpha

PR #1457, opened 2024-02-20
chore: update modal accessibility improvements

Chisomchima:
Implements [Jira Ticket](https://dhis2.atlassian.net/browse/LIBS_567)

---

### Description

Implements  [Jira Ticket](https://dhis2.atlassian.net/browse/LIBS_567) by updating the Modal CloseButton component to update `title` and `aria-label` attributes, as well as the close onClick of the escape button

---

### Known issues

-   [ ] None

---

### Checklist

-   [ ] Tests were added

---

### Screenshots
- None

---

_supporting text_


components/modal/src/modal/modal.js
```diff
@@ -28,6 +28,15 @@ export const Modal = ({
...
 }) => {
     const layerStyles = resolveLayerStyles(hide)
+
+    const handleKeyDown = (event) => {
+        if (event.key === 'Escape' && onClose) {
+            onClose()
+        }
+    }
+
+    document.addEventListener('keydown', handleKeyDown)
```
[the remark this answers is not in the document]
Chisomchima: Thanks for the feedback. I have updated this commit with the changes above.

Birkbjo (changes requested):

components/modal/src/modal/modal.js
```diff
@@ -28,6 +28,21 @@ export const Modal = ({
...
     const layerStyles = resolveLayerStyles(hide)
+
+    const handleKeyDown = (event) => {
+        if (event.key === 'Escape' && onClose) {
+            onClose()
+        }
+    }
+
+    useEffect(() => {
+        document.addEventListener('keydown', handleKeyDown)
```
Birkbjo, line 39: You would probably want to return early when `hide === true`, so that the event listener is not active while its not on the screen.

Birkbjo: What I mean is something like this:

```js
    useEffect(() => {
        if (hide) {
            return
        }

        const handleKeyDown = (event) => {
            // think it's good practice to check for `defaultPrevented` and respect that
            if (!event.defaultPrevented && event.event.key === 'Escape' && onClose) {
                onClose()
            }
        }
        document.addEventListener('keydown', handleKeyDown)

        return () => {
            document.removeEventListener('keydown', handleKeyDown)
        }
    }, [hide, onClose])
```

components/modal/src/modal/modal.test.js
```diff
@@ -1,8 +1,30 @@
-import { render, screen } from '@testing-library/react'
+import { render, screen, fireEvent } from '@testing-library/react'
 import React from 'react'
+import { CloseButton } from './close-button.js'
 import { Modal } from './modal.js'
 
 describe('Modal', () => {
+    describe('Accessibility', () => {
+        it('closes when ESC key is pressed', () => {
```
Birkbjo, line 8: As mentioned in the `Split Button`-PR I think it would make sense to add some test-cases with button component that also handle "ESC keys". Eg. Selects. And test that the modal is not closed if esc is pressed when that is open.

components/modal/src/modal/modal.js
```diff
@@ -28,6 +28,27 @@ export const Modal = ({
...
 }) => {
     const layerStyles = resolveLayerStyles(hide)
+
+    useEffect(() => {
+        if (hide) {
+            return
+        }
+
+        const handleKeyDown = (event) => {
+            event.preventDefault()
```
Birkbjo, line 38: I'm not sure if `preventDefault` is necessary, I believe it's just for some HTML-components. Maybe @ismay has some input in regards to this?

Birkbjo (approved):
Looks good to me! Just one question I'm not totally sure of in regards to `preventDefault()`. 

Would be great to have some tests for the modal that tests for correct escape-handling with other components like Select and SplitButton inside the modal. But we can add that later, and after the `SplitButton` PR is merged.

components/modal/src/modal/modal.js
Chisomchima: without the prevent default, keys like enter and space bar work as well, what I want is for escape key only to be able to close the modal.

components/modal/src/modal/close-button.js
```diff
@@ -2,15 +2,18 @@ import { colors, theme } from '@dhis2/ui-constants'
...
 import React from 'react'
+import i18n from '../locales/index.js'
 
 const createClickHandler = (onClick) => (event) => {
     onClick({}, event)
 }
 
 export const CloseButton = ({ onClick }) => (
     <button
+        title="Close modal dialog"
```
kabaros, line 13: @Chisomchima the `title` is still not translated

PR #1467, opened 2024-03-21
chore: update user avatar accessibility

Chisomchima:
Implements [LIBS-573](https://dhis2.atlassian.net/browse/LIBS-573)

---

### Description
update aria-label attribute on UserAvatar and TextAvatar to improve accessibility 

[LIBS-573]: https://dhis2.atlassian.net/browse/LIBS-573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/user-avatar/src/user-avatar.js
+++ b/components/user-avatar/src/user-avatar.js
@@ -27,6 +27,7 @@ const UserAvatar = ({
     large,
     medium,
     small,
+    ariaLabel,
 }) => (
     <div
         className={cx(className, {
@@ -37,6 +38,8 @@ const UserAvatar = ({
             extralarge,
         })}
         data-test={dataTest}
+        aria-label={ariaLabel || name}
+        aria-hidden={!name}
     >
         {avatarId ? (
             <ImageAvatar avatarId={avatarId} dataTest={`${dataTest}-image`} />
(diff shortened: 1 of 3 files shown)


components/user-avatar/src/user-avatar.js
```diff
@@ -37,6 +37,8 @@ const UserAvatar = ({
             extralarge,
         })}
         data-test={dataTest}
+        aria-label={name}
```
kabaros, line 40: I think the requirements [here](https://dhis2.atlassian.net/browse/LIBS-573) is to expose a property `aria-label` in the props and use it (maybe you can fallback to `name` if that doesn't exist but should expose the property nevertheless.

For `text-avatar`, it can remain as it is

kabaros (changes requested):

d-rita (commented):
LGTM! However, I think the aria-label prop has to be defined in the UserAvatarProps type as an optional string as described in the ticket.

components/user-avatar/src/user-avatar.js
Chisomchima: Added the improvements

kabaros (approved):

PR #1468, opened 2024-03-26
feat(TabBar | Tab): accessibility improvements for TabBar & Tab

alaa-yahia:
Implements [LIBS-557](https://dhis2.atlassian.net/browse/LIBS-557)

---

### Description

- TabBar has role="tablist" 

- Tab has role="tab" 

- `Selectedprop` maps to `aria-selected`

- `Disabled` prop maps to `aria-disabled`

- TabBar acts as a composite component, a user can tab to TabBar 

- A user can use left/right arrow keys to move between Tab components. 

- Pressing the left/right arrow keys will cycle back to the first or last tab, ensuring continuous navigation.

- Moving to a Tab activates it.

- Tab components with overflowing labels displays a Tooltip with full label on hover and focus.

---

### Checklist

-   [ ] Tests were added


kabaros (changes requested):
the code looks good - could you add tests to cover the new scenarios. You can use add e2e if you can, but it might be easier to add unit tests for these new scenarios.

Also add a storybook story for the overflow scenario.

alaa-yahia:
> the code looks good - could you add tests to cover the new scenarios. You can use add e2e if you can, but it might be easier to add unit tests for these new scenarios.
> 
> Also add a storybook story for the overflow scenario.

The storybook story is already included [here](https://ui.dhis2.nu/demo/?path=/story/tab-bar--tab-states), Do you suggest to add a separate one for the overflow ?
For the tests, there wasn't a unit test in the folder probably because the consumer is responsible for managing the state not the `TabBar` component. Should I include an example component and use it in the unit test?

kabaros:
> Do you suggest to add a separate one for the overflow ?

Yes, I mean to add a story for the overflow scenario

> Should I include an example component and use it in the unit test?
Yes, you'll need to do that .. I don't know if it needs to manage the state (or just pass down tabs to it), but basically a component to be able the new logic you added for handling keys, so you render a tab bar, and trigger key events to make sure the right tabs have the focus.

alaa-yahia:
> > Do you suggest to add a separate one for the overflow ?
> 
> Yes, I mean to add a story for the overflow scenario
> 
> > Should I include an example component and use it in the unit test?
> > Yes, you'll need to do that .. I don't know if it needs to manage the state (or just pass down tabs to it), but basically a component to be able the new logic you added for handling keys, so you render a tab bar, and trigger key events to make sure the right tabs have the focus.

What I mean the text overflow story is included here: 
image

kabaros:
> What I mean the text overflow story is included here:
ah ok .. yeah then that's fine. No need to add anything.

PR #1471, opened 2024-04-08
feat(Chip): accessibility improvements

alaa-yahia:
Implements [LIBS-552](https://dhis2.atlassian.net/browse/LIBS-552)

---

### Description

Implement:
 [LIBS-553](https://dhis2.atlassian.net/browse/LIBS-553)
 [LIBS-554](https://dhis2.atlassian.net/browse/LIBS-554)
 [LIBS-555](https://dhis2.atlassian.net/browse/LIBS-555)

---

### Checklist

-   [ ] Tests were added


components/chip/src/chip/chip.js
```diff
@@ -0,0 +1,166 @@
...
+        if (!onRemove) {
+            return
+        }
+        if (event.key === 'Backspace' || event.key === 'Delete') {
+            onRemove({}, event)
+        }
+    }
+
+    return (
+        <button
```
Birkbjo, line 40: Chips are mainly used for links. Button is definitely better than the previous span, but I think we should look into how we would support links, instead of having to do the navigation imperatively in `onClick` handler. It's not valid HTML to wrap a button in a link, so this component would have to support it. 

It might be worth supporting a `component`-prop, and `href` as props. See how material UI does it: https://mui.com/material-ui/react-chip/#clickable-link . A lot of times you would want to use a Chip with eg. `react-router`s `NavLink`. 

@kabaros @ismay

components/chip/src/chip/chip.js
```diff
@@ -0,0 +1,166 @@
...
+    marginRight,
+    marginTop,
+    marginInlineStart,
+    marginInlineEnd,
+}) => {
+    const handleKeyDown = (event) => {
+        if (!onRemove) {
+            return
+        }
+        if (event.key === 'Backspace' || event.key === 'Delete') {
```
kabaros, line 34: I'd say also `event.stopPropagation` here

components/chip/src/chip/chip.js
kabaros: it sounds correct .. but not in the scope of this PR, right? we can add a separate ticket for it

kabaros (changes requested):
very good PR, not an easy one to implement - two things missing for me:

- I am able to navigate through the chips, but I don't get a visual feedback that a chip is selected. They should have a focused style similar to buttons. (see screenshot below)

- Please also add a StoryBook to show the new component ChipGroup, and update the documentation page (for Chip, or separate one) to mention this new component and its use.

chip-accessibility

components/chip/src/chip/chip.js
Birkbjo: You're right, but we may want to release that together. I brought it up because this PR changes the underlying element from span to a button. And wrapping a button in anchor tag is not valid, so it's potentially breaking .

PR #1473, opened 2024-04-15
fix(button): fix prop types for button

Birkbjo:
### Description

I don't think we should extend `HTML`-elements directly. In some cases like `tabIndex` we're actually using a conflicting type, which causes type-errors in our types. 
`tabIndex` should really be a `number` (that's what the html type is), but that would be breaking due to our prop-types etc.

So instead of ButtonProps extending the `HTMLButtonElement`, I'm using type intersection with the `ComponentPropsWithoutRef<'button'>` without the keys for `ButtonProps`, to avoid conflicts. 
I also split this in two types types because eg. `export interface ButtonProps extends Omit<React.ComponentPropsWithoutRef<'button'>, keyof ButtonProps>`) wouldn't work because of circular references. 

See https://github.com/typescript-cheatsheets/react/blob/main/docs/advanced/patterns_by_usecase.md#wrappingmirroring, and expand "Why not ComponentProps or IntrinsicElements or [Element]HTMLAttributes or HTMLProps or HTMLAttributes?"  for why we're using `ComponentPropsWithoutRef`, as opposed to the other prop-types.

### DIFF
--- a/components/button/types/index.d.ts
+++ b/components/button/types/index.d.ts
@@ -16,7 +16,6 @@ type ButtonOpenEventHandler<
     Event extends React.SyntheticEvent = React.MouseEvent<HTMLButtonElement>
 > = (arg0: ButtonEventPayload & { open: boolean }, event: Event) => void
 
-
 export interface BaseButtonProps {
     /**
      * Component to render inside the button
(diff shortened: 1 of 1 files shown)


eirikhaugstulen (approved):
LGTM!

components/button/types/index.d.ts
```diff
@@ -102,6 +102,9 @@ export interface ButtonProps extends HTMLButtonElement {
     onKeyDown?: ButtonEventHandler<React.KeyboardEvent<HTMLButtonElement>>
 }
 
+export type ButtonProps = BaseButtonProps &
+    Omit<React.ComponentPropsWithoutRef<'button'>, keyof BaseButtonProps>
```
KaiVandivier, line 106:
```suggestion
// For prop types we don't define, use native HTML button props
// See https://github.com/dhis2/ui/pull/1473
export type ButtonProps = BaseButtonProps &
    Omit<React.ComponentPropsWithoutRef<'button'>, keyof BaseButtonProps>
```

KaiVandivier (approved):
Looks good -- suggested a comment since it's a complex type, and for the poor souls without git lens 😁

PR #1474, opened 2024-04-15
feat(FileListItem): accessibility improvements for FileListItem

alaa-yahia:
Implements [LIBS-562](https://dhis2.atlassian.net/browse/LIBS-562)

---

### Description

Remove action in list of attached files is focusable and triggerable with `enter`& `delete` keys.

---

### Checklist

-   [ ] Tests were added

### DIFF
--- a/components/file-input/src/file-list/file-list-item.js
+++ b/components/file-input/src/file-list/file-list-item.js
@@ -14,91 +14,108 @@ const FileListItem = ({
     onCancel,
     cancelText,
     dataTest,
-}) => (
-    <p className={cx('selected-file', className)} data-test={dataTest}>
-        <span className="icon">
-            {loading ? (
-                <CircularLoader extrasmall />
-            ) : (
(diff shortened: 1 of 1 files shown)


components/file-input/src/file-list/file-list-item.js
```diff
@@ -14,91 +14,108 @@ const FileListItem = ({
...
+    const handleKeyDown = (event) => {
+        if (!onRemove) {
+            return
+        }
+        if (
+            event.key === 'Enter' ||
+            event.key === 'Backspace' ||
+            event.key === 'Delete'
+        ) {
+            onRemove({}, event)
```
Birkbjo, line 27: Do we want to call `event.stopPropagation` here?

kabaros (approved):
tested and working well - just address @Birkbjo comment please, it's generally safer to do `event.stopPropagation` for these special handlers, but let us know if that's not the case here.

file-input-focus

PR #1500, opened 2024-05-08
fix: li key to be unique

flaminic:
### Description

Small fix to list key in segmented control


components/segmented-control/src/segmented-control.js
```diff
@@ -35,7 +35,7 @@ export const SegmentedControl = ({
     return (
         <ul className="segmented-control" aria-label={ariaLabel}>
             {options.map(({ label, value, disabled }) => (
-                <li key="label">
+                <li key={`option-${value}`}>
```
Birkbjo, line 38: I guess the original intent was to use the label as the key, but forgot to use the variable.

Value makes sense though!

Birkbjo (approved):
LGTM

tomzemp (approved):
good catch 🙏

PR #1511, opened 2024-06-03
feat: accessibility updates from alpha

kabaros:
This is a cherry-pick for features already in `alpha` branch. Fixes:

- Button accessibility: https://dhis2.atlassian.net/browse/LIBS-565 
- Loaders accessiblity: https://dhis2.atlassian.net/browse/LIBS-560
- Tooltip accessibility: https://dhis2.atlassian.net/browse/LIBS-558
- Split button accessibility: https://dhis2.atlassian.net/browse/LIBS-566

### DIFF
--- a/components/tooltip/src/tooltip.js
+++ b/components/tooltip/src/tooltip.js
@@ -115,6 +115,7 @@ const Tooltip = ({
                     onFocus={onFocus}
                     onBlur={onBlur}
                     ref={popperReference}
+                    tabIndex={0}
                     aria-describedby={open ? 'tooltipContenDhis2Ui' : ''}
                     data-test={`${dataTest}-reference`}
                 >
--- a/components/tooltip/src/tooltip.js
+++ b/components/tooltip/src/tooltip.js
@@ -9,8 +9,6 @@ const TOOLTIP_OFFSET = 4
 
 const popperStyle = resolve`
     z-index: ${layers.applicationTop};
-    pointer-events: none;
-
     // Hide popper when reference component is obscured (https://popper.js.org/docs/v2/modifiers/hide/)
     div[data-popper-reference-hidden="true"] {
         visibility: hidden;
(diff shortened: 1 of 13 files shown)


components/tooltip/src/tooltip.js
```diff
@@ -66,33 +64,63 @@ const Tooltip = ({
...
                     ref: popperReference,
                 })
             ) : (
                 <span
                     onMouseOver={onMouseOver}
                     onMouseOut={onMouseOut}
+                    onFocus={onFocus}
+                    onBlur={onBlur}
                     ref={popperReference}
+                    aria-describedby={open ? 'tooltipContenDhis2Ui' : ''}
```
alaa-yahia, line 118: Could you add `tabIndex={0}` to this span?
I forget to add it in my pr

components/tooltip/src/tooltip.js
```diff
@@ -66,33 +64,63 @@ const Tooltip = ({
...
             {typeof children === 'function' ? (
                 children({
                     onMouseOver: onMouseOver,
                     onMouseOut: onMouseOut,
+                    onFocus: { onFocus },
+                    onBlur: { onBlur },
                     ref: popperReference,
                 })
             ) : (
                 <span
```
alaa-yahia, line 112: Could you add tabIndex={0} to this span?
I forgot to add it in my pr

d-rita (commented):
Couple of thoughts (might be out of scope of the tickets): 

1. Tested the `split button` and i think we should be able to Tab it - just like the button component and activate it when i press `Enter`. Since there are 2 buttons, I can tab to any of the 2 to activate what i want. The `Esc` functionality after viewing the dropdown content works as expected. 

2. Tested the `Tooltip`. I can move focus to the one with a button trigger by tabbing to it, but the tooltip does not show as expected when the trigger receives focus. The 'Esc' functionality does work for the button trigger example.

alaa-yahia:
> Couple of thoughts (might be out of scope of the tickets):
> 
> 1. Tested the `split button` and i think we should be able to Tab it - just like the button component and activate it when i press `Enter`. Since there are 2 buttons, I can tab to any of the 2 to activate what i want. The `Esc` functionality after viewing the dropdown content works as expected.
> 2. Tested the `Tooltip`. I can move focus to the one with a button trigger by tabbing to it, but the tooltip does not show as expected when the trigger receives focus. The 'Esc' functionality does work for the button trigger example.

@d-rita The issue with the Tooltip arises from the missing tabIndex. Since it is a <span> element, it will not receive focus by default. Please verify if it now works as expected after adding the tabIndex.
Thank you

d-rita:
@alaa-yahia Tooltip works as expected 🎉

d-rita (approved):

Chisomchima:
@d-rita your concerns have been added to a new (ticket)[https://dhis2.atlassian.net/browse/LIBS-611] and implementations can be found on this (PR)[https://github.com/dhis2/ui/pull/1512]

PR #1533, opened 2024-06-23
(menu): add keyboard accessibility to menus

d-rita:
Implements [LIBS-623](https://dhis2.atlassian.net/browse/LIBS-623)

---

### Description
This feature handles keyboard accessibility of the menu and its children components in the following ways:

- [ ] A user can use the Up and Down arrow keys to navigate a list of menu items by moving focus between them. 
- [ ] A user can use the Space or Enter key to trigger actionable menu items, e.g toggle, open link, act like a button, and open a submenu. 

---

### Checklist

-   [ ] Tests were added


components/menu/src/menu/use-menu.js
```diff
@@ -0,0 +1,91 @@
...
+            return
+        }
+
+        const menu = menuRef.current
+
+        // Focus the first menu item when the menu receives focus
+        const handleFocus = (event) => {
+            if (event.target === menuRef.current) {
+                const firstItemIndex = focusableItemsIndices?.[0]
+                menuRef.current.children[firstItemIndex].focus()
```
flaminic, line 73: this bit errors in the storybook for AutoHideFirstSectionHeaderDivider. it is because firstItemIndex in this case is undefined. Maybe we want to wrap this in a conditional?
```suggestion
                menuRef.current.children[firstItemIndex].focus()
```

components/menu/types/index.d.ts
```diff
@@ -69,6 +77,7 @@ export interface MenuItemProps {
      * When true, nested menu items are shown in a Popper
      */
     showSubMenu?: boolean
+    tabIndex?: number
```
flaminic, line 80: i think suffix (from a previous pr is missing here as well, can we add it as part of this pr?

flaminic:
Thanks Diana. Looks good to me. I only have some minor comments above

components/menu/src/menu/use-menu.js
d-rita: Okay, so the AutoHideFirstSectionHeaderDivider only contains 2 section headers as its children and in my implementation, those do not receive visual focus. So i have to adjust for cases where the menu does not contain any focusable items. Good catch, thank you

components/menu/types/index.d.ts
d-rita: Will do

flaminic (approved):

d-rita:
There definitely was a bit more to this task. 😅 Thank you for the review @flaminic and @kabaros. 🎉

> as discussed privately, it'd be good to:
> 
> 1. memoise `childrenToRender` to make sure we're not recalculating it unnecessarily
> 2. add some comments for the reasons for the extra checks (handling warning when passing regular html etc...)
> 
> but this can be done in the subsequent PR

I have implemented this feedback in this [PR](https://github.com/dhis2/ui/pull/1539). I have also added the menu children role checks and warning if none is provided as was discussed.

PR #1534, opened 2024-06-24
fix(selectorbar): handle components other than SelectorBarItem correctly

alaa-yahia:
### Description

This pull request addresses an issue with the SelectorBar component when it is passed children components other than SelectorBarItem. The existing logic for managing focus relies on components forwarding refs, which causes failures when other components that don't forward refs are passed as children.

### Changes:
Updated the focus management logic to avoid relying on forwarded refs.
Ensured that the SelectorBar component behaves correctly regardless of the types of child components passed.

---

### Checklist

-   [ ] Tests were added

### DIFF
--- a/components/selector-bar/src/selector-bar-item/selector-bar-item.js
+++ b/components/selector-bar/src/selector-bar-item/selector-bar-item.js
@@ -24,158 +24,147 @@ const offsetModifier = {
  * "multiple" selections, this would be more or less impossible to predict
  * inside this component
  */
-export const SelectorBarItem = React.forwardRef(
-    (
-        {
-            children,
-            className,
-            dataTest,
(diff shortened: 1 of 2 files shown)


d-rita (commented):
Works as expected with different children receiving focus as I loop through them with the `Right` and `Left` arrow keys. 🎉

However, there are some things I noticed that can be worked on in separate tickets. For example, 
- when I press `Space` on the selector bar item, I cannot close it after with the same key press. 
- when I tab to the `ClearSelections` button, its parent receives focus but not itself until I press the `Tab` key again. It might be better for it to receive focus when someone has moved to it so they can just press `space` or `enter` to activate its functionality

d-rita (approved):

PR #1551, opened 2024-07-04
fix: do not hijack input when inside a menu

kabaros:
When an input is inside a menu item, the logic to handle a `space` (for accessibility) was hijacking any spaces entered in the input. Now we're restricting the handling of spaces for toggling a menu item to the menu item itself. (Bug Reported by Tracker)

I added a test to recreate the bug, and validate the fix.


d-rita (approved):
LGTM

PR #1553, opened 2024-07-15
fix(ts and prop types): correct some types and propTypes

Mohammer5:
While moving the instance manager FE from CRA to Vite, I encountered some type issues in this repository, see commit messages:

* fix(icon types): export IconSubtract16 & IconSubtract24
* fix(menu item types): make suffix prop optional
* fix(modal types): accept ReactNodes as children
* fix(tag types): accept ReactNodes as children

### DIFF
--- a/components/modal/types/index.d.ts
+++ b/components/modal/types/index.d.ts
@@ -36,7 +36,7 @@ export interface ModalContentProps {
 export const ModalContent: React.FC<ModalContentProps>
 
 export interface ModalTitleProps {
-    children?: string
+    children?: React.ReactNode
     dataTest?: string
 }
 
--- a/components/menu/types/index.d.ts
+++ b/components/menu/types/index.d.ts
@@ -80,7 +80,7 @@ export interface MenuItemProps {
     /**
      * A supporting element shown at the end of the menu item
      */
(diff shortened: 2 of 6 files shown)


components/modal/types/index.d.ts
```diff
@@ -4,7 +4,7 @@ import { LayerBackdropClickHandler } from '@dhis2-ui/layer'
 export type ModalOnCloseEventHandler = LayerBackdropClickHandler
 
 export interface ModalProps {
-    children?: React.ReactNode
+    children?: React.ReactNode || React.ReactNode[]
```
Birkbjo, line 7: I don't think this is needed? `ReactNode` already allows arrays, through `ReactFragment`-type?

Birkbjo (changes requested):
I'm pretty sure we don't need `ReactNode[]`. Other than that, looks good.

tomzemp (approved):

components/modal/types/index.d.ts
Mohammer5: Ah yeah, good catch! Plus I put it on the wrong type, should've been ModalTitleProps.

Birkbjo (approved):

PR #1568, opened 2024-08-05
[WIP] docs: migrating demos to embed instead

Topener:
[WIP]


kabaros (approved):
💯 approved .. just the two comments we discussed on Slack (for future improvements):

- I think the Reload button doesn't add a lot of value, but as you said you intend to add more interactive demos and then it might be useful for it to reset state (I would all it "Reset" though, or just have a reload icon or something more subtle)

- I think the two buttons (reload + show full demo) can be designed a bit more inline with the rest of the page .. they feel a bit jarring for me (@cooper-joe might have better input here though)

Finally, I think we can remove the demo links next. They're redundant now as we have links in all the embedded stories

docs/src/components/DemoComponent.module.css
```diff
@@ -20,4 +20,37 @@
...
+
+.buttons {
+    display: flex;
+    gap: 8px;
+}
+
+.button {
+    background-color: var(--ifm-color-primary);
+    color: white;
+    border: 1px solid var(--ifm-color-emphasis-200);
```
cooper-joe, line 36:
```suggestion
    background-color: transparent;
    color: var(--ifm-color-emphasis-900);
    border: 1px solid var(--ifm-color-emphasis-700);
```

docs/src/components/DemoComponent.module.css
```diff
@@ -20,4 +20,37 @@
...
+    border-radius: 4px;
+    cursor: pointer;
+    font-size: 0.8rem;
+    padding: 4px 8px;
```
cooper-joe, line 40:
```suggestion
    padding: 2px 4px;
```

docs/src/components/DemoComponent.module.css
```diff
@@ -20,4 +20,37 @@
...
+    text-decoration: none;
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    transition: background-color 0.3s ease;
+}
+
+.button:hover {
+    background-color: var(--ifm-color-primary-dark);
```
cooper-joe, line 49:
```suggestion
    background-color: var(--ifm-color-emphasis-900);
    border-color: var(--ifm-color-emphasis-900);
    color: #fff;
```

docs/src/components/DemoComponent.jsx
```diff
@@ -1,15 +1,80 @@
...
+                Demo
+                <div className={styles.buttons}>
+                    <button className={styles.button} onClick={handleReload}>
+                        Reload
+                    </button>
+                    <button
+                        className={styles.button}
+                        onClick={handleShowFullDemo}
+                    >
+                        Show Full Demo
```
cooper-joe, line 54:
```suggestion
                        Go to full demo
```

cooper-joe (changes requested):
A few suggestions for the button styling. 

Agreed with @kabaros about renaming `Reload` to `Reset demo`. I also think it might be worth omitting this button for now until we have the interactive functionality in place.

Topener:
@cooper-joe @kabaros there are already some demo's that do allow altering (such as the alert bar) which makes the reloading of the demo important. I could make it optional and only show it when it allows for it, but I guess that's an improvement we can handle later.

PR #1574, opened 2024-08-07
fix: verify date input only on blur on calendar input, close calenda…

flaminic:
Implements [LIBS-652](https://dhis2.atlassian.net/browse/LIBS-652)

---

### Description

This PR temporally solves a tabbing problem, where tabbing away from the calendar input does not go into the calendar widget popup. We will look into proper accessibility of this component in a following PR. 

This Pr also changes the component so that date validation is only done on blur, rather than on the input value changes

---

### Checklist

-   [x] Tests were added

---

### Screenshots

_supporting text_


kabaros (approved):
💯 🥇 .. works as expected, fixes the onBlur issue, as well as performance and displaying error while typing 🙏🏿

PR #1594, opened 2024-09-11
docs: fix storybook input error

Topener:
Fixes an error introduced with https://github.com/dhis2/ui/pull/1590/ causing an error in storybook prod. Implements [libs-674](https://dhis2.atlassian.net/browse/LIBS-674)

Specifically tagging @KaiVandivier to check this as the circular dependency issue sounds like it needed solving too


Mohammer5:
This seems wrong to me. The component is within the `@dhis2-ui/input` library, so importing a component from itself via it's module name shouldn't be happening.

If there is a circular dependency, meaning that `src/input` imports from `src/input-field`, that what should be fixed instead as that shouldn't be happening

Topener:
Closing this in favour of the better fix

PR #1598, opened 2024-09-14
fix: update default props to JS default params

kabaros:
Implements [LIBS-682](https://dhis2.atlassian.net/browse/LIBS-682)

As we move to React 18 in platform as part ofhttps://github.com/dhis2/app-platform/pull/876, all our components and apps need to stop using defaultProps as these were deprecated  and [started generating a warning](https://github.com/facebook/react/pull/25699) since React 18.3.

Most of this transformation was done using a [codemod](https://gist.github.com/kabaros/45ed16660ac89dc48d22e03836b1e981) - the codemod is a bit brittle and tied to this project for now, but can be made more generic and incorporated in `d2 migrate` as we test it in more projects.

[LIBS-682]: https://dhis2.atlassian.net/browse/LIBS-682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KaiVandivier:
Worked with @amcgee and found that the tests that were hanging were ones using `Popper` components, and then discovered that the default `modifiers = []` param was creating a new array on each rerender, which triggered a `useMemo` recomputation since it's a dependency, which caused a render loop

Setting the prop default to a stable object reference fixed the loop issue, which was testable in storybook stories, so we updated other default props to similar stable objects -- running tests now to see if they work 🤞

amcgee (approved):
With the update to use stable default prop references this looks good to me!

kabaros:
great find! thanks @amcgee @KaiVandivier

PR #1602, opened 2024-09-23
docs: fix storybook inputField error [LIBS-674]

amcgee:
Alternative to #1594 which doesn't rely on external package exports


package.json
```diff
@@ -28,7 +28,6 @@
         "setup": "./scripts/setup.js",
         "build": "./scripts/build-world.sh",
         "build:lib": "./scripts/build.js",
-        "build:docs": "./scripts/build-docs.js",
```
amcgee, line 31: This change is unrelated, but this script is broken and unused so removing it here

Topener (approved):

KaiVandivier (approved):
Nice! 🙂

PR #1609, opened 2024-09-26
fix: clear button and arrow alignment issues

alaa-yahia:
Implements [LIBS-654](https://dhis2.atlassian.net/browse/LIBS-654)

- Resolved the alignment issue of the clear button when warning or error icons are displayed.
- Resolved the issue where the previous and next arrow buttons would shift position based on the length of the month name.


kabaros (approved):
works as expected👌🏿

PR #1619, opened 2024-10-14
feat: make input field clearable and add prefix icon

d-rita:
Implements [LIBS-699](https://dhis2.atlassian.net/browse/LIBS-699)

[LIBS-699]: https://dhis2.atlassian.net/browse/LIBS-699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/input/src/input-field/input-field.js
+++ b/components/input/src/input-field/input-field.js
@@ -40,7 +40,6 @@ class InputField extends React.Component {
             inputWidth,
             autoComplete,
             clearable,
-            clearText,
             prefixIcon,
             dataTest = 'dhis2-uiwidgets-inputfield',
         } = this.props
@@ -83,7 +82,6 @@ class InputField extends React.Component {
                         readOnly={readOnly}
                         autoComplete={autoComplete}
                         clearable={clearable}
-                        clearText={clearText}
                         prefixIcon={prefixIcon}
                         width={inputWidth}
                     />
@@ -97,8 +95,6 @@ const InputFieldProps = {
     /** The [native `autocomplete` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete) */
     autoComplete: PropTypes.string,
     className: PropTypes.string,
-    /** function to clear the input value */
-    clearText: PropTypes.func,
     /** Makes the input field clearable */
     clearable: PropTypes.bool,
     dataTest: PropTypes.string,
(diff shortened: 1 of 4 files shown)


components/input/src/input/input.js
```diff
@@ -199,6 +220,35 @@ export class Input extends Component {
                     input {
                         width: 100%;
                     }
+
+                    .input-prefix-icon input {
+                        padding-left: 30px;
```
Birkbjo, line 225:
```suggestion
                        padding-inset-start: 30px;
```

components/input/src/input/input.js
```diff
@@ -199,6 +220,35 @@ export class Input extends Component {
...
+                    }
+
+                    .input-clearable input {
+                        padding-right: 30px;
```
Birkbjo, line 229:
```suggestion
                        padding-inline-end: 30px;
```

components/input/src/input/input.js
```diff
@@ -187,6 +200,14 @@ export class Input extends Component {
                         'read-only': readOnly,
                     })}
                 />
+                {clearable && value.length ? (
+                    <button onClick={clearText}>
```
Birkbjo, line 204: I think we should add a `className`, and `type=button`. If this doesn't have a type, it will default to `submit`, and post if it's within a form.

components/input/src/input/input.js
```diff
@@ -187,6 +200,14 @@ export class Input extends Component {
                         'read-only': readOnly,
                     })}
                 />
+                {clearable && value.length ? (
```
Birkbjo, line 203: `value` can be undefined, causing this to crash. 

```suggestion
                {clearable && value?.length ? (
```

components/input/src/input-field/input-field.js
```diff
@@ -90,6 +97,10 @@ const InputFieldProps = {
     /** The [native `autocomplete` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete) */
     autoComplete: PropTypes.string,
     className: PropTypes.string,
+    /** function to clear the input value */
+    clearText: PropTypes.func,
```
amcgee, line 101: @d-rita do we need a separate callback function for this?  I think it might be possible to just call `setValue('')`?  Maybe there are cases when you want to reset it to the default value... but that could be handled by the setValue callback?

d-rita: In the Input component, I could use the `onChange` handler that we pass in to update value to an empty string. Then within the parent component where this `InputField` is being used, the `onChange` handler (which uses the setValue) can update the value accordingly. If there is a defaultValue, that can also be set within there.

amcgee (approved):

kabaros (approved):

PR #1620, opened 2024-10-16
feat(select accessibility): add <SingleSelectA11y/> component

Mohammer5:
Implements [JIRA_ISSUE_ID](https://dhis2.atlassian.net/browse/LIBS-559) (partially)

---

### Description

Adds an accessible version of the single select component.
This component is significantly different from the existing single select component in some aspects,
so I decided to create a new component rather than modifying the old one.

Style-wise there should be no difference between the existing component and the new one.
There are some differences in which props exist and which props are expected:

The new component
* expects a `idPrefix` prop.  
  The HTML element that displays the selected value (combobox) needs to point to the element that has the options (listbox), which is done by using id attributes. Instead of creating one randomly, I decided to make this prop requried.
* expects an options array rather than children.  
  This makes it easier to access the options without having to [go through children](https://github.com/dhis2/ui/blob/master/components/select/src/select/option-helpers.js#L17:L24). In order to support custom options, the options can have a `component` prop, which will [override the default option style](https://github.com/dhis2/ui/blob/3a73e558bd0f13a1169fbc353f086202ce922b60/components/select/src/single-select-a11y/option.js#L67) while the accessible option wrapper will stay intact
* accepts a `valueLabel` prop.  
  This should help providing the selected value (and its label) while still loading the options. The component will try to get the label from the options array, but if we know that there's a chance that the options array won't include the selected option (e.g. due to filtering, loading options, displaying a partial options list), the label can be supplied. This eliminates an issue we've had in the past with other components where the selected option would have to be added to the options array, regardless of whether it should be displayed or not 
* requires the `clearText` prop to be present when the select is `clearable`.  
  This way we can have an `aria-label` on the clear button.
* has a filter, which is now controlled by the consumer (`filterValue` & `onFilterChange`).  
  Consumers can now use the filter to load more options or control the options list in any other way
* has a different set of `data-test` properties

---

### Breaking changes (disregarding prop changes)

* The component can't be given an arbitrary set of children. It doesn't accept any children whatsoever

---

### Thinks I want to discuss

* [ ] The new component has `A11y` in its name. I don't think we should replace the old component but deprecate it to give developers time to transition to the accessible variants. **But how should we call the new one?**
* [ ] The new component accepts an `autoFocus` prop, which was called `initialFocus` before. I didn't really think about the name until I realized it deviates. I suppose I should just use the old name?
* [ ] When using custom options, every option in the array has to have a `component` property. Would it make sense to introduce a `optionComponent` prop on the select which would apply to all options unless an individual option has a `component` property?

---

### Open TODOs

* [ ] Keyboard interactions
* [ ] Loader at the bottom of the list is missing

---

### Checklist

-   [x] Tests were added


Mohammer5:
@Topener 

> What exactly are the breaking changes compared to the previous input besides new props and features? Are certain functionalities no longer working? 

The existing component accepts children, which means you can pass anything to the component and it will be rendered in the menu, which you can see [here](https://ui.dhis2.nu/demo/?path=/story/single-select--with-invalid-options).
This means that apps that rely on this feature can't do the same with the new component.

Other potential breaking changes relate to custom options: The options in the new component uses a `button` element and customization can only be done within the button. If apps were putting block elements or several buttons into a single option, this wouldn't be possible anymore.

---

> I would say a breaking change release would be better than introducing a new component, and if both components should be kept in the package maybe the old one should be renamed so people have the option to upgrade with just a name change?

I can elaborate my reasoning for introducing an additional component with a new name: I'm mostly worried about the impact on devs when they want to upgrade UI for other reasons. Say we introduce a breaking change but the teams don't have time at that moment to transition, then release another feature that is unrelated to the select component and one of the apps needs that change. If the app has many selects, it'd be a lot of work. Releasing two components, deprecating the old one and keeping its name for the time-being would allow devs a transition time.

Of course we could force apps to use the new component and be more accessible asap. This mostly depends on how we want to move forward organization-wide. My stance is to be careful which breaking changes in essential components and give developers enough time to transition from one component to another without stopping them from upgrading the library when they need other changes.

---

> I am also missing some documentation about this in the docs/docs directory.

Yes, this is still draft and the docs checkbox isn't checked. The goal is to discuss the general direction now that I have something I can show.

Topener:
Alright I get the reasoning! I'm not the one to decide that but that would be my recommendation. Releasing a new major version also allows for breaking changes and I think that is the best way forward as it would be a new best practice. It also allows people to stay on an older version if they don't have time to upgrade. We'll just have to document the breaking change properly, not only in the releasenotes but also in the `/docs/docs` directory with the component

Perhaps @amcgee can share his opinion on the matter.

components/select/src/single-select-a11y/single-select-a11y.js
```diff
@@ -0,0 +1,392 @@
...
+    menuLoadingText: PropTypes.string,
+
+    /** Allows to modify the max height of the menu **/
+    menuMaxHeight: PropTypes.string,
+
+    /** String that will be displayed when the select is being filtered but the options array is empty **/
+    noMatchText: requiredIf((props) => props.filterable, PropTypes.string),
+
+    /** For a11y: How aggressively the user should be updated about changes in options **/
+    optionUpdateStrategy: PropTypes.oneOf(['off', 'polite', 'assertive']),
```
Birkbjo, line 348: Is this only used for `aria-live`? If so could we at least mention that's what it controls? Or even a reference to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live

docs/docs/recipes/single-select-a11y-server-side-filtering.js
```diff
@@ -0,0 +1,260 @@
...
+function useLoadFilteredDataElementsQuery(customOptions) {
+    const engine = useDataEngine()
+    const timeout = useRef(null)
+    const abortController = useRef()
+    const [loading, setLoading] = useState(false)
+    const [error, setError] = useState(null)
+
+    const refetch = useCallback(
+        ({ searchTerm, page }) => {
+            if (timeout.current) {
```
Birkbjo, line 43: I feel like this is unnecessarily complex? Do I understand correctly that this is to prevent refetches happening when one is already in flight, potentially overwriting data? Why don't we just check `isLoading` before calling `refetch`?

Mohammer5: > Why don't we just check isLoading before calling refetch

In that case you'd have to wait for the result after ever character you enter into the search field.
When scrolling down the list, the same `refetch` function is used, but that's just to keep things DRY.

> I would also think this could potentially "skip" a page if a next-page is inflight and it's called again within 200ms, cancelling the previous page?

Loading the next page is actually prevented when `loading` is true (see [here](https://github.com/dhis2/ui/blob/98514fd7fc99ae2e586a7f1a3cec493e9c5bdd56/docs/docs/recipes/single-select-a11y-server-side-filtering.js#L200)).
It will cancel the previous request when loading the next page and then changing the filter while the request is still outgoing.

Mohammer5: > It's very thorough tho - I will give you that!

The idea was to show the simplest filtering as well as the most complex, allowing the user to either add a little (to the simple use case) or remove from the thorough examples.

Adding more examples is always a good idea IMO when it makes sense!
Should I add one like this: Always load the same amount of items when not searching (showing the latest values only) and displaying all search results, unpaginated)

components/select/src/single-select-a11y/single-select-a11y.js
Mohammer5: Fair point! I'm having difficulties with finding a good name though.. Just `ariaLive` or `ariaLiveForOptions`? What do you think

components/select/src/single-select-a11y/single-select-a11y.js
```diff
@@ -0,0 +1,392 @@
...
+    loading = false,
+    menuLoadingText: _menuLoadingText = '',
+    menuMaxHeight = '288px',
+    noMatchText: _noMatchText = '',
+    optionUpdateStrategy = 'polite',
+    placeholder = '',
+    prefix = '',
+    tabIndex = '0',
+    valid = false,
+    value = '',
```
Birkbjo, line 85: Would it make sense for the value to be an object with the same structure as the options? 
I feel like this would make it more consistent when passing the props.

Mohammer5: Yeah, that's actually a good idea.. Would get rid of the `valueLabel` prop, too!

PR #1623, opened 2024-10-25
fix(calendar): stop extra call to validation when choosing from picker

kabaros:
Fixes a bug when using the day picker (rather than manual entry) where the blur handler causes an extra call to onDateSelect with undefined, causing a flicker in the consumer. 

This also addresses an issue with "clear" functionality to allow consumers to clear an input regardless of validation rules if it's clearable - this should also be [updated here](https://vscode.dev/github/dhis2/multi-calendar-dates/blob/alpha/src/utils/validate-date-string.ts#L67) to not throw an error on empty dates, but for now, it's handled in the component and will follow up with a PR for multi-calendar-dates after discussing the default behaviour.

The PR also adds some unit tests for the various validation states.


alaa-yahia (approved):
LGTM!

PR #1628, opened 2024-11-05
fix: peer dependency issue with npm publish

kabaros:
`npm publish` seems to be failing [on CI on master](https://github.com/dhis2/ui/actions/runs/11578843795/job/32235877694) - The error is pointing to conflicting versions of `@types/react` - this PR attempts to fix it (without adding a module resolution ). It seems like the incorrect version of @types/react (v17) is coming from selector-bar, which was using an outdated testing-library version. This change removes that dependency since components don't need their own test-library versions - a side effect of that was also that `@testing-library/jest` was being hoisted to the top-level and was working there, even though it's not installed.

The old docusaurus version we use, also points to an outdated @types/react version but I think adding a dev dependency at the top level should sort it.

PS: I don't know why this issue only appears on `npm publish`.


KaiVandivier (approved):
Seems good to me 👍

PR #1637, opened 2024-11-25
fix: remove redundant wrapper object from onfocus and onblur callbacks

HendrikThePendric:
Just a small fix


amcgee (approved):
LGTM 🎉  thanks @HendrikThePendric

PR #1638, opened 2024-11-25
feat(menu-item): add ariaLabel prop to use if label prop is not a string

KaiVandivier:
Implements [LIBS-715](https://dhis2.atlassian.net/browse/LIBS-715)

---

### Description

If `label` is not a string, the `aria-label` attribute ends up as `[object Object]`. An `ariaLabel` prop on the `MenuItem` is added to provide a better value for `aria-label` in that case

---

### Checklist

-   [X] Tests were added
-   [ ] ~Storybook demos were added~

[LIBS-715]: https://dhis2.atlassian.net/browse/LIBS-715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/menu/src/menu-item/__tests__/menu-item.test.js
+++ b/components/menu/src/menu-item/__tests__/menu-item.test.js
@@ -73,4 +73,46 @@ describe('Menu Component', () => {
             false
         )
     })
+
+    describe('aria-label attribute', () => {
+        it('produces a log if the label prop is not a string', () => {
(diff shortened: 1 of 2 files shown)


amcgee (approved):
LGTM

d-rita (approved):

PR #1641, opened 2024-11-26
fix: add format to calendar input props

flaminic:
### Description

simple type fix for calendar input


kabaros (approved):

flaminic:
> @flaminic is it correct that this should only apply to the CalendarInput component and not to Calendar?

YES! thank you Austin.

amcgee (approved):
LGTM 👍

PR #1642, opened 2024-11-27
chore: remove unused react-dev-utils dependency to fix yarn workspace…

amcgee:
Fix for the error

```
error Invariant Violation: expected workspace package to exist for "react-dev-utils"
```

When trying to use `yarn workspace` commands


KaiVandivier (approved):
Seems good to me 👍

Mohammer5 (approved):

PR #1644, opened 2024-11-28
fix(types): fix missing width type for input

Birkbjo:
---

### Description

This PR: LIBS-699/clearable-input-and-prefix-icon added `width` as a prop, but forgot to add it to the types.


Topener (approved):
LGTM

PR #1647, opened 2024-12-04
fix: allow input field to have zero value

flaminic:
### Description

This fixes changing number fields with value 0 to having value ""

### Checklist

-   [x] Tests were added

---

### Screenshots

_supporting text_

### DIFF
--- a/components/input/src/input-field/input-field.js
+++ b/components/input/src/input-field/input-field.js
@@ -66,7 +66,13 @@ class InputField extends React.Component {
                         onChange={onChange}
                         name={name}
                         type={type}
-                        value={value || ''}
+                        value={
+                            value === null ||
+                            value === undefined ||
+                            Number.isNaN(value)
+                                ? ''
+                                : value
+                        }
                         placeholder={placeholder}
                         disabled={disabled}
                         max={max}
--- a/components/input/src/input-field/input-field.js
+++ b/components/input/src/input-field/input-field.js
@@ -66,13 +66,7 @@ class InputField extends React.Component {
                         onChange={onChange}
                         name={name}
                         type={type}
-                        value={
-                            value === null ||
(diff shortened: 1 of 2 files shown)


components/input/src/input-field/input-field.js
```diff
@@ -66,7 +66,13 @@ class InputField extends React.Component {
...
                         name={name}
                         type={type}
-                        value={value || ''}
+                        value={
+                            value === null ||
+                            value === undefined ||
+                            Number.isNaN(value)
+                                ? ''
+                                : value
+                        }
```
amcgee, line 75: Could this just use a [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)?  Are we also trying to handle `NaN` [falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) or just trying to make sure `0` can be passed?

```suggestion
                        value={value ?? ''}
```

amcgee (changes requested):

components/input/src/input-field/input-field.js
flaminic: Yes, i was not sure here. I doubt NaN is ever going to be passed as a value, but I thought maybe safer to keep the previous behaviour for everything that is not 0, -0 or other forms of 0? But happy to change it as well. Do you think `??` would be better?

amcgee:
I'm fine with either here @flaminic !

The one important thing I think is that if we're intending to have defined behavior for NaN values (the solution as written now) then we should make sure we have a unit test for that case. 

I think it would be fine to

KaiVandivier (approved):

amcgee (approved):
🦖 🎉

PR #1652, opened 2024-12-17
fix: correct component imports, open apps in same window

amcgee:
Some imports were relative (`../../../../`) when they should have imported from other component packages (`@dhis2-ui/input`, for example)

I also updated the behavior of app links to open in the same window instead of a new one, maintaining the behavior of the previous app menu.  We might want to support cmd/ctl+enter to open in a new tab.


d-rita (approved):
LGTM!

PR #1654, opened 2025-01-07
chore: add workflow for dependency-track

martinakraus:
Implements [SEC-60](https://dhis2.atlassian.net/browse/SEC-60)

---

### Key features

1. Integration of Static Analysis Security Scanning Tool: Dependency Track: https://dtrack.security.dhis2.org/projects
2. Running every night so it won't bother Developers 

---

### Description

Dependency Track will scan the created SBOM and analyze for CVEs and open vulnerabilities.
Those reports will be evaluated by the security team and will be brought back to the dev teams if something crucial pops up

---

[SEC-60]: https://dhis2.atlassian.net/browse/SEC-60?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/generate-and-upload-bom.yml
+++ b/.github/workflows/generate-and-upload-bom.yml
@@ -1,0 +1,50 @@
+name: 'This workflow creates bill of material and uploads it to Dependency-Track each night'
+
+on:
+    pull_request:
+        types: ['opened', 'edited', 'reopened', 'synchronize']
+        
+# on:
+#    schedule:
+#        - cron: '0 0 * * *'
+
+concurrency:
+    group: ${{ github.workflow}}-${{ github.ref }}
+    cancel-in-progress: true
+
+defaults:
+    run:
+        shell: bash
+
+jobs:
+    create-bom:
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v4
+            - uses: actions/setup-node@v4
+              with:
+                  node-version: 20.x
+
+            - uses: c-hive/gha-yarn-cache@v1
+            - run: yarn install --frozen-lockfile
+
+            - name: Install CycloneDX CLI
+              run: |
+                  curl -s https://api.github.com/repos/CycloneDX/cyclonedx-cli/releases/latest | grep "browser_download_url.*linux.x64" | cut -d '"' -f 4 | wget -i -
+                  sudo mv cyclonedx-linux-x64 /usr/local/bin/
+                  sudo chmod +x /usr/local/bin/cyclonedx-linux-x64
+            - name: Generate BOMs
+              run: |
+                  npm install -g @cyclonedx/cdxgen
+                  cdxgen -o sbom.json
+            - name: Upload SBOM to DependencyTrack
+              env:
+                  DEPENDENCY_TRACK_API: 'https://dt.security.dhis2.org/api/v1/bom'
+              run: |
+                  curl -X POST "$DEPENDENCY_TRACK_API" \
+                      --fail-with-body \
+                      -H "Content-Type: multipart/form-data" \
+                      -H "X-Api-Key: ${{ secrets.DEPENDENCYTRACK_APIKEY }}" \
+                      -F "project=53c6ea2f-413f-45b9-a360-e366f917277d" \
+                      -F "bom=@sbom.json"
--- a/.github/workflows/generate-and-upload-bom.yml
+++ b/.github/workflows/generate-and-upload-bom.yml
@@ -1,12 +1,8 @@
 name: 'This workflow creates bill of material and uploads it to Dependency-Track each night'
 
 on:
-    pull_request:
-        types: ['opened', 'edited', 'reopened', 'synchronize']
-        
-# on:
-#    schedule:
-#        - cron: '0 0 * * *'
+    schedule:
+        - cron: '0 0 * * *'
 
 concurrency:
     group: ${{ github.workflow}}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-and-upload-bom.yml
```diff
@@ -0,0 +1,46 @@
...
+                  node-version: 20.x
+
+            - uses: c-hive/gha-yarn-cache@v1
+            - run: yarn install --frozen-lockfile
+
+            - name: Install CycloneDX CLI
+              run: |
+                  curl -s https://api.github.com/repos/CycloneDX/cyclonedx-cli/releases/latest | grep "browser_download_url.*linux.x64" | cut -d '"' -f 4 | wget -i -
+                  sudo mv cyclonedx-linux-x64 /usr/local/bin/
+                  sudo chmod +x /usr/local/bin/cyclonedx-linux-x64
```
amcgee, line 32: @martinakraus is the downloading of this binary necessary since we install `cdxgen` using `npm` in the next step..?

It seems to me that we should prefer `npm install`, as they do in this CycloneDX github action (though using a slightly different tool, the goal of creating a `node_modules` BOM looks the same...) https://github.com/CycloneDX/gh-node-module-generatebom/blob/master/index.js#L28

.github/workflows/generate-and-upload-bom.yml
```diff
@@ -0,0 +1,46 @@
...
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@v4
+            - uses: actions/setup-node@v4
+              with:
+                  node-version: 20.x
+
+            - uses: c-hive/gha-yarn-cache@v1
+            - run: yarn install --frozen-lockfile
```
amcgee, line 26: This is probably fine, and likely comes from some of our other actions in this repository.  We don't use this in our more modern standard repos, though, so might not be necessary to have this dependency.

amcgee (changes requested):
Generally looks good!  Added a few comments.

It would also probably be a good idea to move this to a shared workflow (see https://github.com/dhis2/workflows-platform) and then refer to it from any repositories where we want to track the BOM

.github/workflows/generate-and-upload-bom.yml
```diff
@@ -0,0 +1,46 @@
...
+                  cdxgen -o sbom.json
+            - name: Upload SBOM to DependencyTrack
+              env:
+                  DEPENDENCY_TRACK_API: 'https://dt.security.dhis2.org/api/v1/bom'
+              run: |
+                  curl -X POST "$DEPENDENCY_TRACK_API" \
+                      --fail-with-body \
+                      -H "Content-Type: multipart/form-data" \
+                      -H "X-Api-Key: ${{ secrets.DEPENDENCYTRACK_APIKEY }}" \
+                      -F "project=53c6ea2f-413f-45b9-a360-e366f917277d" \
```
amcgee, line 45: If we make this project ID a repository variable we can extract this whole workflow and call it from individual repositories - see for example https://github.com/dhis2/route-manager-app/blob/main/.github/workflows/comment-and-close.yml

.github/workflows/generate-and-upload-bom.yml
amcgee: We have two repositories for shared workflows, https://github.com/dhis2/workflows (not necessarily reusable workflows) and https://github.com/dhis2/workflows-platform where we currently keep reusable workflows that apply across most of our apps

martinakraus: @amcgee I would love to do both. But there are two issues:
https://www.npmjs.com/package/%40cyclonedx/yarn-plugin-cyclonedx is used for creating SBOMS with CycloneDX github action. This required Yarn > 3.
There are some frontend repos (for example `dashboard-app` which are still on Yarn 1.2.2
So in order to move it to a common Workflow Repo and use this Workflow everywhere, I believe a common Yarn version is required. 
Or I can use the CycloneDX Github action at least justin those frontend Repos which are at least on Yarn version 3

amcgee: We only support `yarn v1` for the moment, so there won't be any repos that have yarn v3 unfortunately.

amcgee: It still seems that the binrary download here isn't necessary though right?  Or is that a prerequisite for `npm install -g @cyclonedx/cdxgen` in the next step?  Otherwise I'm not sure where this downloaded binary executable is being used?

.github/workflows/generate-and-upload-bom.yml
```diff
@@ -0,0 +1,46 @@
...
+
+            - name: Install CycloneDX CLI
+              run: |
+                  curl -s https://api.github.com/repos/CycloneDX/cyclonedx-cli/releases/latest | grep "browser_download_url.*linux.x64" | cut -d '"' -f 4 | wget -i -
+                  sudo mv cyclonedx-linux-x64 /usr/local/bin/
+                  sudo chmod +x /usr/local/bin/cyclonedx-linux-x64
+            - name: Generate BOMs
+              run: |
+                  npm install -g @cyclonedx/cdxgen
+                  cdxgen -o sbom.json
```
amcgee, line 36: (minor) Technically I think this could be just `npx @cyclonedx/cdxgen -o sbom.json`, so you don't need to install anything globally

.github/workflows/generate-and-upload-bom.yml
martinakraus: Thanks so much @amcgee you are absolutely right - this was totally unused.
The CycloneDX CLI can be used to merge SBOMs together in one - I have used it for dhis2-core. There I have created a SBOM for each project and in the enc I merged them in one SBOM.
But not necessary at all here - this was just copy + paste confusion. Sorry for that!

.github/workflows/generate-and-upload-bom.yml
martinakraus: So I removed the c-hive/gha-yarn-cache@v1 Workflow as well :)

amcgee (approved):
LGTM 🎉

PR #1655, opened 2025-01-27
fix(select): adjust active option style

cooper-joe:
Implements [UX-178](https://dhis2.atlassian.net/browse/UX-178)

---

### Description

This PR adjusts the `active` style for `SingleSelect` options to improve readability and better align with native browser `select` styles. Minor cosmetic changes to the `MultiSelect` `active` option are included for consistency.

---

### Checklist

-   [x] Tests were added

---

### Screenshots

Before (`SingleSelect`): 

image

After (`SingleSelect`):

image

Before (`MultiSelect`):

image

After (`MultiSelect`):

image

`SingleSelect` option handling multiline content:

image

[UX-178]: https://dhis2.atlassian.net/browse/UX-178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/select/src/multi-select-option/multi-select-option.js
+++ b/components/select/src/multi-select-option/multi-select-option.js
@@ -20,7 +20,7 @@ const MultiSelectOption = ({
     value,
 }) => (
     <div
-        className={cx(className, { disabled })}
+        className={cx(className, { active, disabled })}
         data-test={dataTest}
         data-value={value}
         data-label={label}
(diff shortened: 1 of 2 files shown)


kabaros (approved):
I am a bit worried that this might break some tests in consuming apps, depending on how they test select options .. but this has to do more with how brittle these tests are and they should be updated anyhow - for example, I've seen instances where we'd use react-testing-library getByTestId followed by checking the content, which would fail with this change, while  using getByText directly would work and should be the correct way of writing such tests

PR #1658, opened 2025-01-31
fix: app redirects on command palette launch

d-rita:
Implements [DHIS2-18943](https://dhis2.atlassian.net/browse/DHIS2-18943)

---

### Description

This PR separates the keydown logic specific to the command palette component from that attached to the document.

**Previous Behaviour**:
-  Previously, when a user pressed the `Enter` key on the apps menu icon, an app would open immediately. 
- This was due to the command palette launching and the document `Enter` key logic triggering the selection of the first highlighted item, hence redirecting to the app. 

**Fixed Behaviour**: 
- Only the key down logic for the keyboard shortcuts (`Ctrl/Meta` + `/`) used to open the command palette is attached to the document since it can be triggered anywhere within the app. 
- All other keydown functionality specific for use within the command palette is attached to its parent component. 
- Now, pressing the `Enter` key on the apps menu icon opens the command palette without a problem

---

### Checklist

-   [ ] Tests were added

---

### Screenshots
**Before**

https://github.com/user-attachments/assets/f5c96123-acca-4da6-9300-0917d55151a1

**After**

https://github.com/user-attachments/assets/b429a555-4951-4898-8f73-37398bb1dcdf

[DHIS2-18943]: https://dhis2.atlassian.net/browse/DHIS2-18943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/header-bar/src/command-palette/command-palette.js
+++ b/components/header-bar/src/command-palette/command-palette.js
@@ -88,15 +88,14 @@ const CommandPalette = ({ apps, commands, shortcuts }) => {
                 <IconApps24 color={colors.white} />
             </button>
             {openModal ? (
-                <ModalContainer setShow={setOpenModal} show={openModal}>
-                    <div
-                        data-test="headerbar-menu"
-                        className="headerbar-menu"
-                        ref={modalRef}
-                        tabIndex={0}
-                        onFocus={handleFocus}
-                        onKeyDown={handleKeyDown}
-                    >
+                <ModalContainer
+                    setShow={setOpenModal}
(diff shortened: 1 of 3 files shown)


d-rita:
- Issue [DHIS2-18943](https://dhis2.atlassian.net/browse/DHIS2-18943) fixed in #1661
- SonarQube issues addressed in #1661

PR #1659, opened 2025-02-03
feat: add commands support and make logout searchable

d-rita:
Implements [LIBS-735](https://dhis2.atlassian.net/browse/LIBS-735)

---

### Description

- This PR is an attempt to start the work on adding commands to the command palette, which means users can search for them too. 

The current implementation:
- adds a hook for the `logout` command functionality. 
- retrieves all commands created through a `useCommands` hook, which are then passed to the `headerbar` component making them available and searchable. (In this case, only the `logout` command has been added)

(**_TBD_**: whether commands are fetched from the backend or defined in the UI here)

This PR also handles:

- updating the action handlers in the `useActions` hook to have an action attribute with its associated action handler, for example, for `Browse Apps`, its action is a function to switch views to the `all apps list view`. 
- passing the actions array to the `useNavigation` hook to simplify the `Enter` key logic. 
- refactoring the `Enter` key down logic by utilising the app's `defaultAction(url)` attribute to open it in a new window, and the commands/actions `action` attribute to trigger its attached handler. 
- differentiating between an icon and image

**Conclusion**:  Users can search for the logout command across the command palette. 

---

### Checklist

-   [ ] Tests were added

---

### Screenshots

**Logout Search Results**
logout results

**All Commands View**
browse cmds

### DIFF
--- a/components/header-bar/src/command-palette/command-palette.js
+++ b/components/header-bar/src/command-palette/command-palette.js
@@ -44,7 +44,7 @@ const CommandPalette = ({ apps, commands, shortcuts }) => {
         setOpenModal,
         itemsArray: currentViewItemsArray,
         showGrid: apps?.length > 0,
-        actionsLength: actionsArray?.length,
+        actionsArray,
     })
 
     useEffect(() => {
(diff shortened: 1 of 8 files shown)


d-rita:
This was changed to a searchable logout action instead of command in #1661. Commands support will be implemented at a later stage.

d-rita:
Closing this PR since the issue is worked on in #1661

PR #1663, opened 2025-02-10
feat: add year month selector to calendar component

alaa-yahia:
Implements [LIBS-726](https://dhis2.atlassian.net/browse/LIBS-726)

image

---

### Checklist

-   [ ] Tests were added

### DIFF
--- a/components/calendar/src/calendar-input/calendar-input.js
+++ b/components/calendar/src/calendar-input/calendar-input.js
@@ -38,6 +38,7 @@ export const CalendarInput = ({
     strictValidation,
     inputWidth,
     dataTest = 'dhis2-uiwidgets-calendar-inputfield',
+    pastOnly,
     ...rest
 } = {}) => {
     const ref = useRef()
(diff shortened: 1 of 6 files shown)


components/calendar/src/calendar-input/calendar-input.js
```diff
@@ -38,6 +38,7 @@ export const CalendarInput = ({
     strictValidation,
     inputWidth,
     dataTest = 'dhis2-uiwidgets-calendar-inputfield',
+    pastOnly,
```
kabaros, line 41: as a general approach - shouldn't this just be controlled by `minDate` / `maxDate` from the client? The client should set the max date (we have a helper that gets the current date in the specified calendar) .. it will then avoid having hardcoded number of years that seems tracker-specific right now

alaa-yahia: The minDate and maxDate validation parameters are not being passed through to the calendar component. Also, the useDatePicker hook is not utilizing these validation parameters because the validateDateString function is being called directly within the calendar input component.

cooper-joe:
Design changes from review are included in [this PR](https://github.com/dhis2/ui/pull/1669)

kabaros (approved):

PR #1671, opened 2025-02-20
feat(command-palette): grid layout

d-rita:
Addresses [LIBS-752](https://dhis2.atlassian.net/browse/LIBS-752)

---

### Description

This PR adjusts the grid layout on the `HomeView` depending on the screen size. 

- For screens with a viewport `width < 480`, it shows a `2 * 4` grid (columns * rows). 
- For larger screens, a `4 * 2` grid is shown. 

It also adjusts the layout and navigation accordingly depending on how many apps are available in the grid. 

---

### Checklist

-   [x] Tests were added

---

### Screenshots

1. Full grid - small screens

https://github.com/user-attachments/assets/a905c538-c78b-449f-ace8-3b3222792b9c

2. Partial grid - small screens

https://github.com/user-attachments/assets/a48f545f-1900-46a2-bf6e-7ba6e5c3c6fe

3. Partial Grid - Large screens

https://github.com/user-attachments/assets/9776969d-0c00-448f-afdb-cacc121e1e9e

[LIBS-752]: https://dhis2.atlassian.net/browse/LIBS-752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/header-bar/src/command-palette/hooks/use-navigation.js
```diff
@@ -77,13 +81,20 @@ export const useNavigation = ({ itemsArray, actionsArray }) => {
             const actionsListLength = actionsArray.length
 
             if (showGrid) {
+                const apps = itemsArray?.filter((item) => item.type === APP)
+                const expectedGridSize = gridColumns * gridRows
+                const appsGridSize =
+                    apps.length >= expectedGridSize
+                        ? expectedGridSize
+                        : apps.length
```
amcgee, line 89: This logic is repeated, perhaps it should be refactored or passed as an argument directly?

components/header-bar/src/command-palette/hooks/use-grid-layout.js
```diff
@@ -0,0 +1,76 @@
...
+                } else {
+                    setGridColumns(columns)
+                    setGridRows(Math.trunc(availableApps / columns + 1))
+                }
+            }
+        } else {
+            setGridColumns(columns)
+            setGridRows(rows)
+        }
+    }, [apps, gridLayout])
```
amcgee: I think this can all be rewritten as a single `useMemo` instead of 2 additional state variables - the function takes `apps` and `gridLayout` (maybe with a different name?) and returns an object `{ rows: x, columns: y }`

Then call it with a `useMemo`:
```js
const rowsAndColumns = useMemo(() => getRowsAndColumns(apps, gridLayout), [apps, gridLayout])
```

amcgee: I feel like `gridLayout` is actually a good name for the full rows/columns/totalCount object and maybe what's currently called `gridLayout` should instead be something like `formFactor` or `isMobile` or something?  Not a huge issue

components/header-bar/src/command-palette/utils/grid-functions.js
amcgee, line 1: It might make sense to have a `Grid` class instead of a bunch of helper functions?

amcgee (changes requested):
Made a few code cleanup suggestions - using `useMemo` is better than `useEffect` here because it avoids an additional re-render

d-rita:
Grid navigation logic superseded by #1675

PR #1678, opened 2025-03-17
fix(input): prefix and clear button

d-rita:
### Description

This PR repositions the prefix and clear button depending on the direction flow of content. It also ensures that the clear button  is kept within the input border and does not overlap a status icon if rendered

---

### Checklist

-   [x] Tests were added

---

### Screenshots

**Before**
1. RTL - wrong position of prefix and clear icon
rtl - non compliance

2. Overlap of clear button with status icon
overlap

**After**
1. Full width - LTR
LTR - long input

2. Full width - RTL
RTL - long input

3. Fixed Width - LTR
LTR - short input 

4. Fixed Width - RTL
RTL - short input


JohanGHole (approved):
LGTM!

cooper-joe (approved):

PR #1685, opened 2025-05-12
feat(button): add controlled component functionality to SplitButton

eirikhaugstulen:
Summary:
- Added functionality for controlling the open-state through passing in `open` and `onToggle`
- Added storybook example with advanced functionality (`FlyoutMenu` & `Modal`)


KaiVandivier (approved):
Looks good!

I have a readability tip: since `typeof this.props.open === 'boolean'` is used a lot, you could maybe make a method to reuse that:
```js
isControlled = () => typeof this.props.open === 'boolean'
```
that you could then use:
```js
const open = this.isControlled() ? this.props.open : this.state.open
```

PR #1698, opened 2025-07-10
fix: small fix in transfer proptype

flaminic:
The component accepts every number for maxSelections, so changing the prototype accordingly to avoid false warnings


KaiVandivier (approved):
Yeah looks good! 😊

PR #1703, opened 2025-09-18
fix: adjust type definitions to align with prop-types

HendrikThePendric:
I encountered some type mismatches between the prop-types and the type definitions. Quite likely this was to do with the usage of the `requiredIf` prop-type helper.

I've added a comment to each type-definition change with a reference to the prop-type it belongs to.

### DIFF
--- a/components/pagination/types/index.d.ts
+++ b/components/pagination/types/index.d.ts
@@ -14,7 +14,7 @@ export interface PaginationProps {
     isLastPage?: boolean
     nextPageText?: TranslateableProp
     pageCount?: number
-    pageLength?: boolean
+    pageLength?: number
     pageSelectText?: TranslateableProp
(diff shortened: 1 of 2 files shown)


components/pagination/types/index.d.ts
```diff
@@ -14,7 +14,7 @@ export interface PaginationProps {
     isLastPage?: boolean
     nextPageText?: TranslateableProp
     pageCount?: number
-    pageLength?: boolean
+    pageLength?: number
```
HendrikThePendric, line 17: https://github.com/dhis2/ui/blob/3be39ae9c7cb874ea48be34d315a15d915f889d6/components/pagination/src/pagination.js#L141

amcgee (approved):
LGTM

PR #1707, opened 2025-10-30
feat(headerbar): fix locale handling

d-rita:
Implements [LIBS-798](https://dhis2.atlassian.net/browse/LIBS-798)

---

### Description
The previous implementation does not parse the locales with variants well enough to trigger the relevant change in language in the UI. When we pass a `keyUILocale` that is a language variant, say `pt_BR` directly to `i18n.changeLanguage`, it will not be picked up and the UI defaults to `English`.

In this implementation, we parse locales using the same functionality in the [app platform's adapter](https://github.com/dhis2/app-platform/blob/8e2f83c3b70ea2a66a566e1bae3d8326724d0cf8/adapter/src/utils/localeUtils.js#L15). It reformats the locale to follow the [BCP 47 syntax](https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag#bcp_47_syntax). For example, `pt_Br` becomes `pt-Br` which triggers the translation. 

---

### Checklist

-   [ ] Tests were added

---

### Screenshots
**For 'pt_Br'**
pt_Br

[LIBS-798]: https://dhis2.atlassian.net/browse/LIBS-798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/header-bar/src/header-bar.js
+++ b/components/header-bar/src/header-bar.js
@@ -55,11 +55,30 @@ export const HeaderBar = ({
         }))
     }, [data, baseUrl])
 
+    // sources
+    // 1. App Platform: /adapter/src/utils/localeUtils.js#L15
+    // 2. https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag
+    const parseLocale = (locale) => {
+        const [language, region, script] = locale.split('_')
+
+        let languageTag = language
+        if (script) {
+            languageTag += `-${script}`
+        }
+        if (region) {
+            languageTag += `-${region}`
+        }
+        return new Intl.Locale(languageTag)
+    }
+
     // See https://jira.dhis2.org/browse/LIBS-180
     if (!loading && !error) {
         // TODO: This will run every render which is probably wrong!
         // Also, setting the global locale shouldn't be done in the headerbar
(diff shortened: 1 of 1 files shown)


KaiVandivier (changes requested):
So this would be perfect... if DHIS2 locale handling made sense 😅 instead, it's currently going the "wrong way"

--> Our input is either `pt-BR` (BCP-47, v42+) or `pt_BR` (Java .toString, <v42), and our **output** needs to be `pt_BR` (the Java.toString format) for the transifex files we have

P.S., testing can be confusing: it seems that if you set `i18n.changeLanguage('pt-BR')`, something works in a way that it falls back to `pt` -- I can tell since I looked at the translations in `pt_BR.po` and `pt.po`, and some strings are not translated for `pt_BR`,  like "Search for apps". When using `pt-BR` as the language, though, "Search for apps" _is_ translated, so it must be getting that from the `pt.po` file (since we don't have the fallback logic here yet)

I think adding this function can sort everything out: https://github.com/dhis2/app-platform/blob/master/adapter/src/utils/localeUtils.js#L59

PR #1735, opened 2026-04-12
feat(multi-select): add collapseSelectionAfter prop

eirikhaugstulen:
## Summary
- Add a `collapseSelectionAfter` prop to `MultiSelect` that replaces individual chips with "X selected" text once the selection count exceeds the given threshold

https://github.com/user-attachments/assets/fb835e66-6303-4000-9112-c862711c1fd4

### DIFF
--- a/components/select/src/multi-select/input.js
+++ b/components/select/src/multi-select/input.js
@@ -3,6 +3,7 @@ import { colors } from '@dhis2/ui-constants'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React from 'react'
+import i18n from '../locales/index.js'
 import {
     InputClearButton,
     InputPlaceholder,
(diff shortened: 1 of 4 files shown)


cooper-joe:
Nice suggestion @eirikhaugstulen, I think this is a helpful addition to the component.

I do think it is important to frame this as a robustness or resilience feature, rather than an endorsement of using this component for very large selections. If a field often ends up with many selected items, a Transfer or an always-open SelectPanel might be a better fit. But this feature does help prevent an awkward UI when a multiselect occasionally has many selections. This can be explained in the component documentation, I'm happy to add that as a follow up to this.

KaiVandivier (approved):
Please add the new prop to `/components/select/types/index.d.ts` -- otherwise, looks good

PR #1744, opened 2026-05-22
feat(select): flexibility in dropdown menu width

arina-gutuboe:
Implements [LIBS-837](https://dhis2.atlassian.net/browse/LIBS-837)

---

### Description

`SingleSelect` and `MultiSelect` measure the input width and pass it to the dropdown as a fixed pixel width. That works for a wide input, but breaks when the input becomes too narrow: long labels get compressed in the menu, and the result looks unnecessarily tight.

This PR adds two optional props — `menuMinWidth` and `menuMaxWidth` — on `SingleSelect`, `MultiSelect`, `SingleSelectField`, and `MultiSelectField`. They are forwarded through the shared `Select` / `MenuWrapper` stack.

image

---

### Known issues

With only `menuMinWidth` (no `menuMaxWidth`), the menu can grow without an upper bound for very long option labels; `menuMinWidth` and `menuMaxWidth` should be used together when labels can vary significantly in length.

---

### Checklist

-   [x] Tests were added

---

### Screenshots

`menuMaxWidth` is slightly greater than the width of the input, but less than the width of the options:
Screenshot 2026-05-22 at 11 37 33

with `menuMinWidth`:
Screenshot 2026-05-22 at 11 40 20

[LIBS-837]: https://dhis2.atlassian.net/browse/LIBS-837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/select/src/multi-select/multi-select.js
+++ b/components/select/src/multi-select/multi-select.js
@@ -15,6 +15,8 @@ const MultiSelect = ({
     selected = staticArr,
     tabIndex,
     maxHeight,
+    menuMinWidth,
+    menuMaxWidth,
     inputMaxHeight,
     onChange,
     onFocus,
@@ -72,6 +74,8 @@ const MultiSelect = ({
                     menu={menu}
                     tabIndex={tabIndex}
                     maxHeight={maxHeight}
+                    menuMinWidth={menuMinWidth}
+                    menuMaxWidth={menuMaxWidth}
                     onChange={onChange}
                     onFocus={onFocus}
                     onKeyDown={onKeyDown}
@@ -131,6 +135,10 @@ MultiSelect.propTypes = {
     loading: PropTypes.bool,
     loadingText: PropTypes.string,
     maxHeight: PropTypes.string,
+    /** Sets a maximum width for the dropdown menu */
+    menuMaxWidth: PropTypes.string,
+    /** Sets a minimum width for the dropdown menu */
+    menuMinWidth: PropTypes.string,
     /** Required if `filterable` prop is `true` */
     noMatchText: requiredIf((props) => props.filterable, PropTypes.string),
     placeholder: PropTypes.string,
--- a/components/select/src/multi-select/multi-select.js
+++ b/components/select/src/multi-select/multi-select.js
@@ -135,9 +135,9 @@ MultiSelect.propTypes = {
     loading: PropTypes.bool,
     loadingText: PropTypes.string,
     maxHeight: PropTypes.string,
-    /** Sets a maximum width for the dropdown menu */
+    /** See [dropdown menu width](https://developers.dhis2.org/docs/ui/components/select#dropdown-menu-width) */
     menuMaxWidth: PropTypes.string,
-    /** Sets a minimum width for the dropdown menu */
+    /** See [dropdown menu width](https://developers.dhis2.org/docs/ui/components/select#dropdown-menu-width) */
     menuMinWidth: PropTypes.string,
     /** Required if `filterable` prop is `true` */
     noMatchText: requiredIf((props) => props.filterable, PropTypes.string),
--- a/components/select/src/select/menu-wrapper.js
+++ b/components/select/src/select/menu-wrapper.js
@@ -7,11 +7,23 @@ import React from 'react'
 const MenuWrapper = ({
     children,
     dataTest,
+    inputWidth,
     maxHeight = '280px',
-    menuWidth,
+    menuMaxWidth,
+    menuMinWidth,
     onClick,
     selectRef,
 }) => {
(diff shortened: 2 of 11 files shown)


components/select/src/multi-select/multi-select.js
```diff
@@ -131,6 +135,10 @@ MultiSelect.propTypes = {
     loading: PropTypes.bool,
     loadingText: PropTypes.string,
     maxHeight: PropTypes.string,
+    /** Sets a maximum width for the dropdown menu */
```
tomzemp: these need to be defined as 'px', right? I tried for example percentages, and they did not work. If the idea is that we just use 'px', it might be good to update the description here (as it flows through into documentation for component)

components/select/src/select/menu-wrapper.js
```diff
@@ -7,11 +7,23 @@ import React from 'react'
...
+    menuMinWidth,
     onClick,
     selectRef,
 }) => {
+    // menuMinWidth or menuMaxWidth enables flexible sizing (fit-content), with
+    // min-width = max(input, menuMinWidth). Without them, width matches the input.
+    const flexible = menuMinWidth || menuMaxWidth
+    const width = flexible ? 'fit-content' : inputWidth
+    const flexibleMinWidth = menuMinWidth
+        ? `max(${inputWidth}, ${menuMinWidth})`
```
tomzemp, line 22: just curious: are we using the css calc definition here to avoid issues with taking the max() ourselves and just passing the resulting number?

tomzemp (approved):
Looks good to me! 🎉 (I've read through the code and tested in the aggregate data entry app).

Are we going to extend this to SimpleSingleSelect component as well? (The docs say we're planning on deprecating SingleSelect (https://github.com/dhis2/ui/blob/master/docs/docs/components/select.md?plain=1#L41) though I don't know if we've ever moved forward with that, but we seem to be using both in our apps.

We might also want to add some notes in the select documentation for this (above the  notes generated from the props) to explain the behaviour.

components/select/src/select/menu-wrapper.js
arina-gutuboe: Yes, that’s correct, and to ensure consistent behavior in the affected components.

components/select/src/multi-select/multi-select.js
arina-gutuboe: That's a good point. Thanks!
I updated the select documentation with the new behavior explanation and added links in the props description. Left a note about avoiding percentages too.

arina-gutuboe:
> Are we going to extend this to SimpleSingleSelect component as well? (The docs say we're planning on deprecating SingleSelect (https://github.com/dhis2/ui/blob/master/docs/docs/components/select.md?plain=1#L41) though I don't know if we've ever moved forward with that, but we seem to be using both in our apps.

Created [LIBS-869](https://dhis2.atlassian.net/browse/LIBS-869) 



[LIBS-869]: https://dhis2.atlassian.net/browse/LIBS-869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

kabaros (commented):
it looks great @arina-gutuboe - just adding to Tom's point, would you be able to add the behaviour to SimpleSelect too? The "Simple" varieties make these components accessible and are the receommended way going forward - The SingleSelect is marked for deprecation: https://developers.dhis2.org/docs/ui/components/select#single-select .. we're just doing it gradually since it's used in a lot of places and there was no way to make the changes backwards compatible (I can give you more context if you'd like)

it would be good to add the same functionality to SimpleSingleSelect before merging - happy to help with that

kabaros:
> > Are we going to extend this to SimpleSingleSelect component as well? (The docs say we're planning on deprecating SingleSelect (https://github.com/dhis2/ui/blob/master/docs/docs/components/select.md?plain=1#L41) though I don't know if we've ever moved forward with that, but we seem to be using both in our apps.
> 
> Created [LIBS-869](https://dhis2.atlassian.net/browse/LIBS-869)

Sorry @arina-gutuboe , I didn't notice your comment about the linked ticket. I will merge this one, and you can catch up with the other ticket after.

[LIBS-869]: https://dhis2.atlassian.net/browse/LIBS-869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

kabaros (approved):

PR #1750, opened 2026-06-18
feat: add React 19 support

HendrikThePendric:
Implements N/A — prompted by a Dependabot PR

---

### Description

Adds React 19 to peer ranges across all packages, keeping React 16/18 support. No public API changes.

- Peer ranges widened to `"^16.13 || ^18 || ^19"` in 49 package.json files.
- Dropped the `react-popper` dependency and kept a local copy of its `usePopper` hook in `@dhis2-ui/popper`. `react-popper` is no longer maintained (archived Dec 2024) and never added React 19 support. The local copy uses `@popperjs/core` directly — same as react-popper did. Runtime behavior is unchanged.
- Ref types in popper/popover/tooltip/intersection-detector widened to accept React 19's `RefObject<T | null>`. Compatible with @types/react 16, 18, and 19.
- Testing-library devDependencies bumped to versions declaring React 19 peer support.

---

### Checklist

-   [x] API docs are generated **N/A**
-   [ ] Tests were added
    -   [ ] ⚠️ TO DO: Run tests against React 19 (and 18 and 16.8+ -- make a test matrix)
-   [x] Storybook demos were added **N/A**

---

### Screenshots

No visual changes.


HendrikThePendric:
This PR will get merged once we add React v19 support to App Platform as a whole. This may or may not coincide with releasing a next major version, in which case we should first merge #1751 into this one.

PR #1753, opened 2026-06-29
fix: sort root nodes unless opted out via suppressAlphabeticalSorting

HendrikThePendric:
Implements [LIBS-864](https://dhis2.atlassian.net/browse/LIBS-864)

---

### Description

Root nodes are now sorted alphabetically unless opted out via `suppressAlphabeticalSorting`. I've scoped this as a fix, because I'd say that this behaviour should have been implemented all along. Sorting children but not roots, while having a generic `suppressAlphabeticalSorting` prop in place is inconsistent behaviour.

#### Incidental
- I deduped the deps as well
- It seems an i18n string for the select was missing, it "popped up" when ran `yarn setup`. Not sure why it didn't get added before, but it seems like a valid addition to me.

---

### Checklist

-   [x] API docs are generated N/A
-   [x] Tests were added
-   [x] Storybook demos were added N/A

[LIBS-864]: https://dhis2.atlassian.net/browse/LIBS-864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tomzemp (approved):
Looks good to me @HendrikThePendric 🙏 . I've read through the code and tested in the users app (data entry app has custom org unit tree, so we'll need to copy these changes).

The only thing I noticed on review is that the existing helper function `sortNodeChildrenAlphabetically` is using `left.displayName.localeCompare(right.displayName)`. I guess this then uses the sorting conventions of the browser language, whereas we might want to use the user's ui language? This can cause some differences in the sort, e.g. English would alphabetise `[Åland, Belgium, Zimbabwe]` whereas Norwegian would alphabetise `[Belgium, Zimbabwe, Åland]` (or English `[na,ño,ny]`, vs Spanish `[na,ny,ño]`). I think this is a minor issue, but might be nice to fix in the context of this PR?

HendrikThePendric:
@tomzemp this is a good observation and it would indeed be better to use the UI locale from DHIS2 instead of the browser locale. But as you mention, this is existing logic and we should probably mark it as out-of-scope. Getting the user locale would involve importing the `useCurrentUserInfo` hook, which was added not so long ago. So I guess this could be problematic for older apps, and maybe is a breaking change?

kabaros (approved):

PR #1757, opened 2026-07-13
fix: tackle slowness of CalendarInput and Ethiopic calendar failures

kabaros:
Bumps @dhis2/multi-calendar-dates to 3.0.0-alpha.1 also:
- fixes severe typing lag / CI timeouts in CalendarInput (a broken-memoization bug in useDatePicker) - not a breaking change for consumers
- Un-skips the previously-blocked Ethiopic validation test and adds Jest coverage for gaps (Ethiopic/Nepali widget selection, several untested props, a perf regression suite). 
- Also fixes a real bug found along the way: CalendarInput passed minDate/maxDate to useDatePicker under the wrong key names, so in strict mode the widget kept an out-of-range date selected even though the input correctly rejected it. 
- Types the validation field on CalendarInput's onDateSelect payload, previously undocumented in types/index.d.ts.

The bump to multi-calendar-dates https://github.com/dhis2/multi-calendar-dates/pull/102 bumps it to latest 0.5 and also vendors it to fix the Ethiopic calendar. These changes are breaking in multi-calendar-dates, but none of the changes affect the API of the UI component - `useDatePicker` consumed signature is unchanged. ⚠️ it might still warrant a breaking change though given the big internal refactor

Fixes: 
- https://dhis2.atlassian.net/browse/LIBS-763 and https://dhis2.atlassian.net/browse/DHIS2-21183 (calendar slowness)
- https://dhis2.atlassian.net/browse/LIBS-830 (Ethiopian calendar through vendoring of Temporal in multi-calendar)

### DIFF
--- a/components/calendar/package.json
+++ b/components/calendar/package.json
@@ -38,7 +38,7 @@
         "@dhis2-ui/input": "10.16.4",
         "@dhis2-ui/layer": "10.16.4",
         "@dhis2-ui/popper": "10.16.4",
-        "@dhis2/multi-calendar-dates": "2.1.2",
+        "@dhis2/multi-calendar-dates": "3.0.0-alpha.1",
         "@dhis2/prop-types": "^3.1.2",
         "@dhis2/ui-constants": "10.16.4",
         "@dhis2/ui-icons": "10.16.4",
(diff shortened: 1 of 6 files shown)


KaiVandivier (approved):
Looks good! (waiting for stable release of multi-calendar-dates)

PR #1764, opened 2026-09-11
feat: support writing components in TypeScript, with a placeholder OuTree

HendrikThePendric:
### Context

In the not-too-distant future I will be working on a new organisation unit component. This new component would live alongside the existing `OrganisationUnitTree`, because I do not expect it to be a drop-in replacement, but something that apps can adopt/migrate to gradually. I'd like to call the new component `OuTree` and I would like to write it in TypeScript.

This current PR is a prelude to the work on the new `OuTree` component and makes it possible to write UI components in the repo in TypeScript. You will find a new component in this PR already but it is simply a stub to verify that the TypeScript setup works correctly for tests, stories, etc. This package is currently set to private so nothing will get published.

### Description

Partial TypeScript support was present already:
- `cli-app-scripts` already included `@babel/preset-typescript`
- The Jest transform already matched `.tsx`
- `d2-style` already globbed `ts,tsx`

What was missing or providing friction was:
- Type-checking (Babel strips types without checking them) (also runs in CI)
- Declaration output
- ESLint parser
- A few globs and module resolution settings needed to be updated because these only accounted for `.js`
- Relative imports in TypeScript are extensionless (`./ou-tree`), the inverse of the explicit `.js` extensions the JavaScript packages require, so the ESLint rule is flipped for `.ts`/`.tsx` only

I tried to make as much as possible global to the repo:
- The root `tsconfig.json` is the TypeScript project for the whole repo, so components need no tsconfig of their own. Editors resolve it for any `.ts`/`.tsx` file too
- `styled-jsx` ambient declaration is in the project level `typings/` dir
- `yarn typecheck` is a single `tsc --noEmit` over every package. Type-checking is a form of linting, so it works the way linting already does here: one command at the root, no per-package script
- `scripts/build-types.js` emits the declarations. Emit has to happen per package, since each publishes its own types and one `tsc` run writes to one `outDir` — but the script reads the compiler options from the root `tsconfig.json`, so they can't drift from the ones used to type-check
- The global ESLint config has been updated to support TypeScript 
- `generate-api-docs.js` now supports packages without a `src/index.js` and documents TypeScript components (this needed no dependency change since react-docgen 5.4.3 already reads TS prop interfaces).

Each component will still need a few changes to enable TypeScript, but no new files:
- The build script gains `&& node ../../scripts/build-types.js`
- `package.json` type fields — `types` and `exports.types` point at `./build/types/index.d.ts` instead of the hand-written `types/` dir, and `files` drops its `types` entry
- `d2.config.js` entry point becomes `src/index.ts`

#### Instructions for Claude

I also added some instructions for Claude in `components/ou-tree/CLAUDE.md`. I opted to keep this local to the component for 2 reasons:
1. Providing instruction for the entire repo is probably a substantial job and would need to be done by code owners and via some sort of democratic process.
2. The instructions cover conventions that apply inside a TypeScript package, so some/most of the instructions/conventions are different from what we'd see in other components.

#### Fixes for indirectly related issues

- `yarn lint` failed on `master`: generated build output wasn't excluded from Prettier, and a `.gitignore` rule for `API.md` contradicted the fact that all 52 are tracked and updated in feature commits. 
- `build-world.sh` had no `set -e`, so its exit status was only the last command's. Every step was non-blocking, including `yarn build:lib`, and the publish job gates on it — so a broken or missing `build/` could have reached `semantic-release`. Docs and Storybook stay non-blocking on purpose, since shipping component fixes matters more than the docs site building, but they now emit a warning visible in the checks UI. The library build does fail hard, which makes sense: we don't want to ship a broken or missing `build/` 😬 .

#### Not fixed
The docs site is broken on `master` too — Docusaurus rejects the `url` in its config, and a sidebar entry points at a deleted recipe. I've left these alone, because there is no clearly correct fix: the invalid sub-path in `url` is exactly what produced correct canonical URLs for the developer portal, and moving it to `baseUrl` would prefix every asset path and break the Netlify PR preview, which is the only thing that consumes this build.  (The new behaviour of `build-world.sh` is visible on this PR — the docs failure is reported as a warning but the pipeline stays green).

---

### Checklist

-   [x] API docs are generated — and the generator now handles TypeScript components, which it previously skipped
-   [x] Tests were added

---


kabaros (approved):
This look great - I have two comments:
- I don't understand why  components still need their own tsconfig? 
- Instead of d2-style, it'd be great if we move to the [shared configs](https://github.com/dhis2/style-configs) .. that's not necessarily on this PR at all, but I thought it might make some of the boilerplate easier to manage, so maybe worth a quick try.

kabaros (approved):

PR #1765, opened 2026-09-14
chore: migrate from cli-style to the shared style configs

HendrikThePendric:
Implements: N/A — follow-up to the [`style-configs`](https://github.com/dhis2/style-configs) rollout ([LIBS-614](https://dhis2.atlassian.net/browse/LIBS-614))

### Description

Replaces `@dhis2/cli-style` with `@dhis2/config-eslint/react` and `@dhis2/config-prettier` from
[dhis2/style-configs](https://github.com/dhis2/style-configs). That carries **eslint 7 → 9 and flat
config** and **prettier 2 → 3**, and takes eslint 7, husky 7, commitlint 12 and ls-lint out of the
dependency tree.

`cli-style` is being phased out by the extensibility team. Its commitlint, ls-lint and stylelint
configs are still in progress upstream, so commitlint is inlined locally here — as in
[event-visualizer-app#336](https://github.com/dhis2/event-visualizer-app/pull/336).

Two commits are formatting only and can be skipped: `style: reflow markdown list markers`
(45 files) and `style: reformat code` (18 files). The other six touch 26 files.

---

### Implementation choices

**Four rules the shared config leaves off are enabled here**, keeping what cli-style did:

- `import/extensions` — the JS packages publish ESM, where Node does no extension guessing, so
  relative imports need `.js`. TS sources are the inverse (`never`), since TypeScript resolves
  extensionless.
- `import/no-unresolved` — off upstream over a resolver issue this repo does not hit.
- `react/react-in-jsx-scope` + `react/jsx-uses-react` — off upstream via `jsx-runtime`.
  `@dhis2/cli-app-scripts` runs `@babel/preset-react` in classic mode, so JSX compiles to
  `React.createElement` and a missing import fails at runtime rather than at build time. The second
  rule stops `no-unused-vars` flagging the import the first one requires.

The shared config's ignores are app-shaped as well (`./*.js`, and `src/`-scoped globs that never
match `components/*/src`), so this repo's list replaces them and re-includes the root config files.

**Other points to take note of:**

- The 62 hand-written `.d.ts` files stay unlinted, as before this PR. Linting them costs ~350
  errors, 277 of them `react/no-unused-prop-types` reading type declarations as components.
- `react/no-object-type-as-default-prop` and `react-hooks/static-components` are off for stories
  only — both concern re-render churn in shipped components.
- eslint stays on 9: eslint-plugin-import and eslint-plugin-react do not support 10 yet
  ([#3230](https://github.com/import-js/eslint-plugin-import/pull/3230),
  [#3979](https://github.com/jsx-eslint/eslint-plugin-react/pull/3979)).
- `@commitlint/cli` is on v20 rather than v21, which needs node ≥ 22.12.
- Node is pinned to `24.x` rather than `lts/*`, so node 26 becoming LTS on 2026-10-28 does not move
  CI on its own.
- Prettier 3 sets markdown list-marker spacing at one space and exposes no option for it, hence the
  45-file commit.

---

### New mechanisms

- **`commitlint.config.mjs` at the root.** Both commit jobs used to require `@dhis2/cli-style` at
  runtime to locate their config. `.mjs` is the filename both actions look for by default, so that
  lookup step is gone rather than moved. `lint-pr-title` also replaces
  `pull-request-name-linter-action`, pinned to the removed node12 runtime, with commitlint reading
  the title from `env:` — it is untrusted input.
- **Plain git hooks.** They sourced a husky shim that never existed here: nothing installs it and
  `core.hooksPath` is unset, so neither has run on a fresh clone. They are dependency-free now.
- **`yarn lint` is `eslint . && prettier --check .`**, and `format` the equivalent, replacing the
  four `d2-style` scripts. `build-world.sh` formats build output directly.

---

### Behaviour changes

- Prettier now also covers `.css`/`.scss`, which `d2-style`'s `*.{md,json,yml,html}` pattern did
  not. One file was reformatted.
- `eslint.config.mjs` is linted; the shared config ignores `eslint.config*` by default.
- Node 20 reached end of life on 2026-04-30, so `node-version: 20.x` pinned every job to an
  unsupported runtime. Moving also satisfies `@eslint/core` and `eslint-visitor-keys`, which want
  `^20.19.0 || ^22.13.0 || >=24`.
- The `@typescript-eslint/parser` pin and the `typescript ~5.5.4` ceiling added in #1764 are gone.
  `typescript-eslint` v8 supersedes both, and disables `no-undef`, `no-redeclare` and
  `no-dupe-class-members` for `.ts`/`.tsx` — the set that was being worked around there.

---

### Known issues

None.

---

### Checklist

- [x] API docs are generated — unchanged; `build:api` runs clean
- [x] Tests were added — none needed; the existing 976 pass
- [x] Storybook demos were added — n/a

---

### Verification

All green: `yarn lint` (0 errors; the 10 remaining warnings are pre-existing
`react-hooks/exhaustive-deps`), `CI=true yarn lint`, which adds `import/no-cycle` and
`import/no-self-import`, `yarn typecheck` on TypeScript 5.9.3, `yarn test` (126 suites, 976
passed), and `build:lib` + `build:api` + the storybook build. CI passes too, including all seven
e2e shards.

Commitlint was checked differentially: across 150 commits of master, v20 and the v12 `cli-style`
bundles accept and reject the same set.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[LIBS-614]: https://dhis2.atlassian.net/browse/LIBS-614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kabaros (approved):

.hooks/pre-commit
```diff
@@ -1,4 +1,20 @@
 #!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
+set -e
 
-yarn d2-style check --staged
+staged_files_without_deletions() {
```
kabaros, line 4: I normally use lint-staged here .. maybe we could discuss what approach to take in general and standardise, but happy with this if it works

